Issue #1527, opened 2013-11-06
[BUG] Uniqueness validator incorrectly fails validation on update


WooDzu:
With the following model $id is the primary key and $name must be unique within $project_id

```php
use Phalcon\Mvc\Model\Validator\Uniqueness;

class Tags extends \Phalcon\Mvc\Model {
  public $id;
  public $project_id;
  public $name;

  public function initialize() {
    $this->setSource('table_tags');
  }

  public function validation()
  {
    $this->validate(new Uniqueness([
      'field' => ['project_id', 'name'],
      'message' => 'The tag already exists in selected project'
    ]));

    return ($this->validationHasFailed() != true);
  }
}
```
Lets say I have the following data:

```php
$data = array(
    'id' => 1,
    'project_id' => 1,
    'name' => 'Some tag',
);
```
It works fine when creating a new record with Tags::save(), but then Tags::update() is failing the validation. There might be something I'm missing here or the Uniqueness validation is not excluding current $id from the check.

Tested on:
Phalcon 1.3.0 and 1.2.4
PHP 5.4.18 and 5.5.5
Windows 7

Proffnet:
Due to 
```
PHALCON_CALL_METHOD(&operation_made, record, "getoperationmade");
if (PHALCON_IS_LONG(operation_made, 2)) {
```
in phalcon source code you must set model in update operation mode "2" then validation will check unique key.
Set:
```
$this->_operationMade=2;
```
during validation to resolve this problem.

salarmehr:
@ This is a horrible bug. Does is solved in version 2.0?

salarmehr:
This bug is open for months. Any news??

salarmehr:
I come up with the following validator:

````
class UniquenessValidator extends \Phalcon\Mvc\Model\Validator
{
    public function validate($record)
    {

        $message = $this->getOption('message') ?: 'DEFAULIT_MESSAGE';
        $field = $this->getOption('field');
        $count = count($record->query()->where("$field=:value: and id!=:id:")->bind([
            'value' => $record->$field,
            'id' => $record->id
        ])->execute());
        if ($count == 0) {
            return true;
        }
        $this->appendMessage($message,$field);
        return false;
    }
}

xfight0:
This is a terrible bug :cactus:

andresgutierrez:
Can you confirm this is still happening in Phalcon 2?

xfight0:
[...]

PR #1600, opened 2013-11-25
[BUG] Test case for issue #1527

pgasiorowski:
This PR adds a test case for issue #1527

Uniqueness validator incorrectly fails validation


salarmehr:
@ This is a horrible bug. Does is solved in version 2.0?

PR #1941, opened 2014-01-30
Make Phalcon\Mvc\Model::findFirst() work when phqlLiterals is false

ghost:
See #886


dschissler:
Its seems like this is a long standing open bug.  Is there any way to work around this?

PR #2889, opened 2014-10-08
Include and require volt function

zacek:
The volt function "include" always throws exception if template was not found. This pull requests changes this behaviour:

1) new function "require" is added that works as "include" used to work before
2) "include" now ignores missing templates

To achieve this, partial method had to be changed too: an 3rd optional parameter "ignore_missing" was added. This parameter had to be taken into account in simple view, but there it has no effect (partial always throws exceptions on missing files, because it is not supported there)

Usage:
{% include "partials/nonexistent.volt" %} (never fails)
{% require "partials/nonexistent.volt" %} (fails if the partial nonexistent.volt doesn't exist)
{{ partial("partials/nonexistent", ['a': 'abc'], true) }} (never fails because of the third optional parameter)

This can be usefull when including e.g. page context help which is optional. There is no simple solution how to test the existence of a partial before using it. This function should solve it.

This pull request was created as a followup of PR #2367 (rebased on 1.3.4)


sergeyklay:
@andresgutierrez If you close the PR - explain why please

andresgutierrez:
@sergeyklay 1.3.4 was released 3 months ago

PR #3044, opened 2014-11-16
Use PHPUnit ~4.2 from composer

ovr:


Green-Cat:
This should probably be closed:
- The temp variable was already replaced by an anonymous var in phalcon/acl/adapter/memory.zep
- The phpunit version was increased, and then decreased back to 3.7.* for backward compatibility.

PR #3092, opened 2014-12-01
[2.0.0] autoescape option for volt engine

vlad-khramov:


Green-Cat:
:+1: This would be nice to have. Many developers expect this functionality coming from twig.

sergeyklay:
Why?

PR #3105, opened 2014-12-05
[2.0.0] Fix getRelationRecords for compound relation

ziras:
It looks like there is a typo, isn't it?
If this is a typo, is this a valid fix?

Sorry I'm quite new to this project.
I don't understand it fully yet.
Please forgive me if I'm wrong.


SidRoberts:
@phalcon 

Yeah, this is definitely a typo, see https://github.com/phalcon/cphalcon/blob/1.3.5/ext/mvc/model/manager.c#L2303.

Issue #2786, opened 2014-09-05
Question on Model hooking


ogarbe:
Can someone from @phalcon or @niden  can look on this question please ?

http://forum.phalconphp.com/discussion/3108/mimic-doctrine-table-inheritance#C11061
##

PR #3107, opened 2014-12-07
[2.0.0] Add late state binding on model and collection

ogarbe:
ORM enhancements : 
- Replace new builder by models manager createBuilder call on model queries
- Late state binding on hydration (cloneResultMap call) to be able to override class creation (refs #2786) - also done here for 1.3.3 : link to #2789 
- Add ini "orm.late_state_binding" (default true) to enable / disable the functionality for models as it's slower than previous implementation (due to the fact that there is no monomorphic method cache)

With this we can override the instantiation of an object to mimic Doctrine inheritance (http://the-phpjs-ldc.rgou.net/symfony1/more-with-symfony/en/09-Doctrine-Form-Inheritance.markdown)

``` php
/**
     * Assigns values to a model inherited class from an array returning a new model.
     *
     * @param \Phalcon\Mvc\Model $base          object to hydrate
     * @param array              $data          data to use for hydration
     * @param array              $columnMap     column mapping
     * @param int                $dirtyState    object state
     * @param boolean            $keepSnapshots keep snapshot
     *
     * @return \Phalcon\Mvc\Model
     */
    public static function cloneResultMap(
        $base,
        $data,
        $columnMap,
        $dirtyState = null,
        $keepSnapshots = null
    ) {
        if (isset($data['columnUsedForTableInheritance'])) {
            $class = '\Ournamespace\Models\\' . $data['columnUsedForTableInheritance'];
            if (class_exists($class)) {
                $base = new $class();
            }
        }
        return parent::cloneResultMap(
            $base,
            $data,
            $columnMap,
            $dirtyState,
            $keepSnapshots
        );
    }
```

### DIFF
--- a/config.json
+++ b/config.json
@@ -76,6 +76,10 @@
         "orm.enable_literals": {
             "type": "bool",
             "default": true
+        },
+        "orm.late_state_binding": {
+            "type": "bool",
+            "default": true
         }
     },
     "destructors": {
--- a/phalcon/mvc/collection.zep
+++ b/phalcon/mvc/collection.zep
@@ -446,7 +446,7 @@ abstract class Collection implements CollectionInterface, InjectionAwareInterfac
 				/**
 				 * Assign the values to the base object
 				 */
-				return self::cloneResult(base, document);
+				return static::cloneResult(base, document);
 			}
 			return false;
 		}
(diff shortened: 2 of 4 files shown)


ogarbe:
you won' t take this change ?

sergeyklay:

andresgutierrez:

PR #3108, opened 2014-12-07
Fix #3073 - Disable view when redirecting

### LINKED ISSUE #3073: [Bug] Perhaps response::redirect should disable the current view?
Inside of beforeExecuteRoute router plugin:
```php
// I just added this line to prevent the view from firing even though it is canceled.
$dispatcher->getActiveController()->view->disable();

// After redirect it tries to access the view and in my case it is a dispatcher security redirect.
// There is a form on the old view view that causes it to crash
$this->response->redirect($path);
return false;
```
It appears that redirecting doesn't always kill the view from being accessed.  It appears as though it is rendered even though it is not displayed.  This flared up when a view was trying to access a form that was not being set because that controller action was not being called.  I hunted this down by finding that the controller action and ::initialize was not being called but that the view was still being created.

So perhaps response redirecting should kill the current view.  I always figured that returning false would do this but it doesn't seem to be the case.

ogarbe:
cf #3073

### DIFF
--- a/phalcon/http/response.zep
+++ b/phalcon/http/response.zep
@@ -431,7 +431,7 @@ class Response implements ResponseInterface, InjectionAwareInterface
 	 */
 	public function redirect(location = null, externalRedirect = false, int statusCode = 302) -> <ResponseInterface>
 	{
-		var header, url, dependencyInjector, matched, message;
+		var header, url, dependencyInjector, matched, message, view;
 
 		if !location {
 			let location = "";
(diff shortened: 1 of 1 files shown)


quantum13:
@phalcon @ogarbe hello, 
View\Simple is not instance of viewInterface, so this commit breaks code which uses view\simple (has no disable method).

Also, if $application->useImplicitView(false) used, disabling views is useless in redirect

PR #3109, opened 2014-12-08
[2.0.0] Changes in registering of modules

vlad-khramov:
Some changes:
- registerAutoloaders is called after registering modules for all modules:
  - It low-cost operation, but very usefull: in multimodule application each module can autoload own classes.
- if in module class exists method 'initialize' it calles after registering
  - It can be used for registering own routes by each module

This two changes are steps to independent reusable modules. For example https://github.com/quantum13/cphalcon/blob/694bf70888c7254218a6d0aca99c94814ee34247/unit-tests/modules/objects/Module.php
- forward to another module #121 (registerServices called in each module in dispatch loop)
  - useful for https://github.com/phalcon/cphalcon/issues/121#issuecomment-45719826
- one backward incompatibility: di must be setted to application before registering modules

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,11 +4,14 @@
  - Phalcon\Forms\Form::label does not share attributes with the ones that belong to the element
  - Added parameter 'params' to set default conditions when retrieving related records
  - Added filters "int!" and "float!" that cast values to the respective type
- - Filter "alphanum" now use regular expression instead of internal function for better behave in 64bits
+ - Filter "alphanum" now uses a regular expression instead of internal functions for better behaviour in 64bits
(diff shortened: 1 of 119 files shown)


skollro:
@quantum13 you have to update your code because of recent changes in Application class... Do you think there's the chance that your code will be merged? I'm waiting for this features, otherwise I have to implement a workaround for such a module loader in plain PHP ;)

7thcubic:
actually the workaround is just 4/5 lines
```php

$modroutes = glob(APPLICATION_PATH."/*/config/routes.php");
foreach($modroutes as $route)
{
    include_once $route;
}
```

PR #3138, opened 2014-12-16
[2.0.0] added method getModule() in class \Phalcon\Mvc\Application

KorsaR-ZN:
Added method `getModule()` in class `\Phalcon\Mvc\Application`
Changes in method `handle()`, gets module definition via new method `getModule()`


phalcon:
Thanks

PR #3141, opened 2014-12-17
[2.0.0] fixed compatibility with 1.3.x

KorsaR-ZN:
Correction in the event `beforeModuleStart` added `moduleName` to event


phalcon:
Thanks

PR #3165, opened 2014-12-28
Zephir config.json Write more info

ovr:


sergeyklay:
Hi @ovr @phalcon 
Can you please add support for these directives like in Phalcon 1.3.x:

ba0f43de-3b1b-11e4-8a58-62bcabab2534

ovr:
https://github.com/phalcon/cphalcon/blob/2.0.0/config.json#L35

sergeyklay:
@ovr 

```php
<?php
$directives = ini_get_all('phalcon');
print_r($directives); // empty
?>
```

Also I can explicitly specify in my models:
```php 
self::setup(['notNullValidations' => false]);
```
selection_001

PR #3185, opened 2015-01-04
registerDirs with merge fix

KarlAustin:
Fix for: "Uncaught exception 'RuntimeException' with message 'Trying to call method merge on a non-object'" when calling registerDirs with merge.

Test case updated to reflect a merge case.


phalcon:
Thanks

PR #3188, opened 2015-01-06
Allow to override the default statsKey

dugwood:
Please just confirm the fact that «isset» in https://github.com/phalcon/cphalcon/blob/2.0.0/phalcon/cache/backend/libmemcached.zep#L86 will treat «null» as «not set» (PHP-like, but I don't know in Zephir).

This is to fix the same issue as https://github.com/phalcon/cphalcon/pull/3183, but in the session handler part.


phalcon:
Thanks

dugwood:
Tested from a fresh new build this morning, that works great. Thanks a lot, that was a big performance killer :-(

PR #3214, opened 2015-01-12
[2.0.0] Implemented new unit tests for Redis and Added ability to specify many keys in delete() method of Redis backend

moisemorard:
Hello,

With this PR, it's now possible to add several keys at the same time in delete method of Redis backend.
This method implement native functionnality of [phpredis driver](https://github.com/phpredis/phpredis#del-delete)

``` php
        $frontCache = new Phalcon\Cache\Frontend\Data(array('lifetime' => 20));
        $cache = new Phalcon\Cache\Backend\Redis($frontCache, array(
            'host' => 'localhost',
            'port' => 6379
        ));
        $cache->save('key1', 'value1');
        $cache->save('key2', 'value2');
        $cache->save('key3', 'value3');
        $cache->delete(array('key1','key2','key3'));
```

Delete method return now Long Number of keys deleted.

Thanks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ before_script:
  - ./vendor/bin/zephir generate
  - (cd ext; export CFLAGS="-g3 -O1 -std=gnu90 -Wall -DZEPHIR_RELEASE=1"; phpize && ./configure --enable-phalcon && make --silent -j4 && sudo make --silent install && phpenv config-add ../unit-tests/ci/phalcon.ini)
  - ulimit -c unlimited || true
+ - phpenv config-add ./unit-tests/ci/redis.ini
 
 script:
  - ./runTests.sh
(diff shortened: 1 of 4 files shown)


phalcon:
The prototype of delete does not work that way in other adapters

moisemorard:
Thanks for your reply,

Ok I understand but it's also possible to delete one key like that:
```php
$cache->delete('key1');
```
So it's backward compatible.
I thought it was cool to implement native possibilities of redis in Phalcon.
And I think it's possible to implement this in other backends.
If I'm out of the guidelines contributions, don't hesitate to tell me.

Thank u

SachaMorard:
Hello

Could you consider this PR again and tell us if you accept it or not ?
In fact we are working on the issue #3218 we report (NFR) and we are waiting for your answer to finish it.
Don't hesitate to tell us what you expect.

Best regards
Sacha

sergeyklay:
Why?

andresgutierrez:

PR #3220, opened 2015-01-13
[2.0.0] Fix addForeignKey() method on Phalcon\Db\Dialect\MySQL

RemyJeancolas:
Thix fix allows to name the constraint explicitly (the name wasn't taken into consideration with the old syntax)

### DIFF
--- a/phalcon/db/dialect/mysql.zep
+++ b/phalcon/db/dialect/mysql.zep
@@ -374,12 +374,12 @@ class MySQL extends Dialect //implements Phalcon\Db\DialectInterface
 		}
 
 		if schemaName {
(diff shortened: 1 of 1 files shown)


sergeyklay:

RemyJeancolas:
Hello @phalcon, this update is very "light" and is easy to understand, why have you closed this PR?
It is *really* usefull to explicitly name constraints in MySQL (which actually doesn't wok), I don't understand why you don't want to merge it.
Could you explain me your point of view?

KorsaR-ZN:
@RemyJeancolas 
Branch 2.0.0 has been freezed.
Please it reopen in the branch 2.0.x

PR #3224, opened 2015-01-14
Use options["server"] and set _collection variable

PatrickG:
options["server"] was never used.
this->_collection was never set.


andresgutierrez:
This is already fixed, thanks

PR #3229, opened 2015-01-15
[2.0.0] Fix #3135

### LINKED ISSUE #3135: [NFR] allow Request::getBestLanguage() to return language without dialect
When I use `\Phalcon\Mvc\Request->getBestLanguage()` on my browser, it returns "en-US".  I think it would be helpful in some cases to have a way to just return "en", as my language files won't be broken down into "en-UK", "en-US", "en-CA", etc.

Something like `getBestLanguageFamily()`, or a variable passed to `getBestLanguage()` would be helpful.

KoriSeng:
Fix #3135: allow Request::getBestLanguage() to return language without dialect

### DIFF
--- a/phalcon/http/request.zep
+++ b/phalcon/http/request.zep
@@ -1024,9 +1024,16 @@ class Request implements RequestInterface, InjectionAwareInterface
 	 *
 	 * @return string
 	 */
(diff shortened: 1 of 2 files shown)


andresgutierrez:

PR #3231, opened 2015-01-15
[2.0.0] Allow unmodified model to be skipped from saving

KoriSeng:
Fix allow model to skip saving if it is unmodified, Require snapshot to be true.

Prevent some database error when attempting to save with unique fields.


phalcon:
This is already implemented with setDynamicUpdate(true)

Issue #3219, opened 2015-01-13
[Phalcon 2] Phalcon\Db\Adapter doesn't fill OnUpdate and OnDelete actions in method describeReferences()


RemyJeancolas:
Hello,

It seems that for all the objects returned by a call to describeReferences(), the methods getOnUpdate() and getOnDelete() return null, for example :

``` php
$references = $connection->describeReferences('my_robots_parts');
foreach ($references as $reference) {
    echo $reference->getOnUpdate(); // returns null
    echo $reference->getOnDelete(); // returns null
}
```

Thanks for your return.

olivier-monaco:
https://github.com/phalcon/cphalcon/blob/2.0.0/phalcon/db/adapter.zep#L1069

Definition array not populated with onUpdate/onDelete. In fact, there seems to be nothing about them in method describeReferences of dialects.

7thcubic:
i found the sql to fix that. attempting to write a patch for it.

//var sql = "select a.TABLE_NAME,a.COLUMN_NAME,a.CONSTRAINT_NAME,a.REFERENCED_TABLE_SCHEMA,a.REFERENCED_TABLE_NAME,a.REFERENCED_COLUMN_NAME,b.update_rule,b.delete_rule from information_schema.KEY_COLUMN_USAGE as a join information_schema.REFERENTIAL_CONSTRAINTS as b on a.CONSTRAINT_NAME = b.CONSTRAINT_NAME and a.CONSTRAINT_SCHEMA = b.constraint_schema where a.REFERENCED_TABLE_NAME is not null and ";
		//if schema {
			//let sql .= "a.CONSTRAINT_SCHEMA = '" . schema . "' AND a.TABLE_NAME = '" . table . "'";
		//} else {
			let sql .= "a.TABLE_NAME = '" . table . "'";
		//}
		//return sql;

Ideally that will give mysql adapter the 6th and 7th field providing the actions. just need to fix it > >

7thcubic:
fixed. well it only works in mysql though and i kinda had to alter adapter to do it. its not pretty at all for those who arent using mysql. =/

since i dont use other db's i cant fix it

RemyJeancolas:
@olivier-Monaco: in fact, I saw that it was not implemented, the question was about the possibility of precisely implement this functionality.
Like @7thcubic I use mysql, and I implemented a fix similar to its own that solves the problem (for mysql only).
[...]

PR #3232, opened 2015-01-15
[2.0.0] Mysql reference actions. #3219 fix

KoriSeng:
Retrieves the onUpdate and onDelete action, added checks on reference to prevent breaking of other dialects.


andresgutierrez:
Could you please fix the braces style?

Issue #3227, opened 2015-01-14
Email Validation


sergeyklay:
### Intro

After a little research it became apparent that the current validators using `FILTER_VALIDATE_EMAIL` is not quite perfect (`Phalcon\Mvc\Model\Validator\Email` and `Phalcon\Validation\Validator\Email`).

Below I will give a small example test.

### Emails

These and many other, more exotic, but real email addresses are valid (see [RFC822](http://www.ietf.org/rfc/rfc0822.txt?number=822), [RFC2822](http://tools.ietf.org/html/rfc2822), [RFC2821](http://tools.ietf.org/html/rfc2821), [RFC3696](http://tools.ietf.org/html/rfc3696))
```
"Joe\\Blow"@example.com
john@doe.com
john.doe@site.com
john_doe@site.com
john_doe@domain.site.com
john@doe-site.com
"Abc@def"@example.com
j.o.h.n@d.o.e.com
isposable.style.email.with+symbol@example.com
john@localhost
j.o.h.n@my.c
"Fred Bloggs"@example.com
```

### Tests

Let us proceed for tests

```php
<?php

function validateViaRegexp($email, $strict = false)
{
    if (mb_strlen($email) > 254) {
        return false;
    }

    if ($strict) {
        $qtext = '[^\\x0d\\x22\\x5c\\x80-\\xff]';
        $dtext = '[^\\x0d\\x5b-\\x5d\\x80-\\xff]';
        $atom  = '[^\\x00-\\x20\\x22\\x28\\x29\\x2c\\x2e\\x3a-\\x3c\\x3e\\x40\\x5b-\\x5d\\x7f-\\xff]+';
        $pair  = '\\x5c[\\x00-\\x7f]';

        $domain_literal = "\\x5b($dtext|$pair)*\\x5d";
        $quoted_string  = "\\x22($qtext|$pair)*\\x22";
        $sub_domain     = "($atom|$domain_literal)";
        $word           = "($atom|$quoted_string)";
        $domain         = "$sub_domain(\\x2e$sub_domain)*";
        $local_part     = "$word(\\x2e$word)*";

        $expression     = "/^$local_part\\x40$domain$/D";
    } else {
[...]

PR #3234, opened 2015-01-15
[1.3.x] Fix #3124 (check email with filter_var)

### LINKED ISSUE #3124: [bug]email valid
I write a test case:
```php
$model2 = new AccountInfo();
$model2->email = 'sdfsaf@asfd$d.+*';
$model2->nickName = 'aaaaa';
$model2->validation();
$this->tester->assertEquals(false, filter_var($model2->email,FILTER_VALIDATE_EMAIL));
$this->assertNotNull($model2->getMessages());
$this->assertCount(1,$model2->getMessages());
```
result:
```php
$this->tester->assertEquals(false, filter_var($model2->email,FILTER_VALIDATE_EMAIL));   # is ok
```
but 
```
$this->assertNotNull($model2->getMessages());     # fail  $model2->getMessages() return null
```

I use Phalcon 1.3.4 and codeception testing framework 2.0.8

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/6941668-bug-email-valid?utm_campaign=plugin&utm_content=tracker%2F50707&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F50707&utm_medium=issues&utm_source=github).
</bountysource-plugin>

maxgalbu:
#3124

### DIFF
--- a/ext/mvc/model/validator/email.c
+++ b/ext/mvc/model/validator/email.c
@@ -87,8 +87,8 @@ PHALCON_INIT_CLASS(Phalcon_Mvc_Model_Validator_Email){
  */
 PHP_METHOD(Phalcon_Mvc_Model_Validator_Email, validate){
 
-	zval *record, *option = NULL, *field_name = NULL, *regs, *invalid = NULL;
-	zval *value = NULL, *pattern, *match_pattern, *match_zero;
+	zval *record, *option = NULL, *field_name = NULL;
+	zval *value = NULL, *email_flags, *valid_email = NULL;
 	zval *message = NULL, *type, *is_set_code = NULL, *code = NULL;
 	zval *allow_empty = NULL;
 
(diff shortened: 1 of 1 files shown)


sergeyklay:
About regex vs `FILTER_VALIDATE_EMAIL` validation

phalcon:
But this is a validator not a filter, we need a regex there

maxgalbu:
In the issue #3124 i saw that phalcon 2 uses filter_var():

https://github.com/phalcon/cphalcon/blob/dc8ced6a83da2d9e148c5ca222abc4db9e4a1604/phalcon/mvc/model/validator/email.zep#L80

Filter_var() return false if the email is not valid, so it should actually act as a validator, if I'm not mistaken

phalcon:
It should use regex on both, filter_var doesn't validate a format correctly

maxgalbu:
Do you agree with #3227?
I think filter_var() validates most of the emails except some "obscure" emails ("string with spaces"@example.com, test@localhost, etc which IMHO are very obscure), and is accepted by the php community.

We can add a "strict" boolean that when enabled makes the function checks the emails with the regex, otherwise it checks them using filter_var().

phalcon:
Maybe adding an extra option to avoid localhost and obscure emails would be fine :+1:

PR #3236, opened 2015-01-15
[2.0.0] fix key typo error of Column definition

SachaMorard:
Fix an error when defining typeValues of a column.


sergeyklay:
Good catch :+1:

PR #3237, opened 2015-01-15
Fixes issue where issent() is called on non-response objects in micro apps -- Micro::handle()

emilh91:
Only calling issent() if the returned value is an instance of Phalcon\Http\ResponseInterface


phalcon:
We don't add new commits directly to master, could you please submit this to the 1.3.5 branch

emilh91:
Submitted to 1.3.5 branch. Closing this PR.

PR #3243, opened 2015-01-19
replacing array->merge with array_merge (last ones)

vlad-khramov:


ovr:
Why remove built-in methods?

PR #3261, opened 2015-01-21
[2.0.0] form default value fix

KoriSeng:
Form was not calling self getters, fixed it so that adding csrf to the form would not cause problems.

csrf, if set as default for element would cause it to update everytime a new instance is created.

now it will use getCsrf().


sergeyklay:
:+1:

phalcon:
Could you please use rebase instead of merge, last commit contains many unrelated changes

7thcubic:
https://github.com/phalcon/cphalcon/pull/3330
use that instead

PR #3268, opened 2015-01-22
[2.0.0] model named scope

KoriSeng:
Namescope and removal of doc blocks

usage is

<code>
public function scope()
    {
        return ["published" => ["conditions" => "type = 1"],
                 "draft" => ["conditions" => "type = ?", "bind" => 2], "jkrowling" => ["conditions" => "type = 1 & author_id = 6"]];
    }
</code>

<code>
Books::model()->published()->draft()->findAll();
Books::model()->jkrowling()->draft()->findAll(["limit" => 20]);
Books::model()->findAll(["conditions" => "author_id = ?", "bind" => 4]);

Books::model() == new Books();
</code>

It would have to create a new instance, in order to daisy chain the scopes.
All scopes will be saved until findAll() is called.

Scopes function will return the model itself.

So unless you guys prefer

Books::published()::draft()::findAll()

i guess this is it

### DIFF
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -502,13 +502,8 @@ abstract class Model implements ModelInterface, ResultInterface, InjectionAwareI
 	 *</code>
 	 *
 	 * @param Phalcon\Mvc\ModelInterface|Phalcon\Mvc\Model\Row base
-	 * @param array data
-	 * @param array columnMap
-	 * @param int dirtyState
-	 * @param boolean keepSnapshots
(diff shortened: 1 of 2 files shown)


quantum13:
-1
Incorrect logic

Books::model() - model instance
so why Books::model()->findAll must return some filtered models? Models must return manager!

See scopes in Yii2: http://www.yiiframework.com/doc-2.0/guide-db-active-record.html#scopes

7thcubic:
oooo damn i knew it was too easy to implement. LOL ok sure let me fix it

Issue #3249, opened 2015-01-19
Phalcon\Di must use InjectionAwareInterface and can also use EventsAwareInterface


olivier-monaco:
From https://github.com/phalcon/cphalcon/blob/2.0.0/phalcon/di.zep#L263
<pre>
		if typeof instance == "object" {
			if method_exists(instance, "setDI") {
				instance->setDI(this);
			}
		}
</pre>

DI is injected if the newly created instance has a setDi method and not if it implements InjectionAwareInterface. So this interface is "useless"... Maybe replacing this with a "instanceof" check?

I think Di must do the same with EventsAwareInterface: if the newly created interface implements EventsAwareInterface and the DI has an EventsManager, it must call setEventsManager on the instance. So, the EventsManager is automatically available.

olivier-monaco:
@phalcon it seems to be a design flaw.

All methods from the Phalcon\Tag class are static. But these methods are often used has instance method (ex. http://docs.phalconphp.com/en/latest/reference/tags.html#tag-service). This is not really an issue... except for setDI method...

If no instance of Tag is created, Tag::getDI() returns NULL. Once one instance of Tag is created using a DI, Tag::setDI() is called by the DI with itself as a parameter. This will not be right anymore if DI use the InjectionAwareInterface as proposed above.

If Tag::setDI/getDI required? If yes, I propose to call Tag::setDI from Di::setDefault.

Current implementation : https://github.com/olivier-monaco/cphalcon/commit/ea33a127f30a0a1c1b9c41881c3347a5f6bae8c8

phalcon:
@olivier-monaco we want to remove all static methods in Phalcon\Tag but probably in 2.1

olivier-monaco:
Ok so I propose to change Tag::getDI to something like:
<pre>
	public static function getDI() -> <DiInterface>
	{
		var di;
		let di = self::_dependencyInjector;
		if typeof di != "object" {
				let di = \Phalcon\Di::getDefault();
		}
		return di;
	}
</pre>

This will be a temporary workaround.

7thcubic:
Whats the plan for rewriting tag method?

Create a helper namespace and shift all helpers into a helper manager? (url,tag,html)? 

or place them all as a service?

PR #3270, opened 2015-01-22
Phalcon\Di must use InjectionAwareInterface and can also use EventsAwareInterface

olivier-monaco:
Issue #3249


olivier-monaco:
@phalcon Of course. Done (and it's better ;)).

PR #3276, opened 2015-01-24
[2.0.0] toArray Bug

KoriSeng:
Checking of column map always fail due to incorrect check, fixed so toArray works
remove clean up doc blocks too


7thcubic:
```php
class Model1 extends Model
{
     public function columnMap()
    {
         return ["id" => "id", "field1" = > "field1"];
    }
} 

$m = Model1::find()->toArray(); //doesnt work error "id" not part of column map.
```

7thcubic:
fix #3287

Issue #3274, opened 2015-01-23
Phalcon v1.3.x deletes url parameters in some situations (namespace/controller/action/0 results in params==NULL)


fsando:
$ php -version
PHP 5.5.20-1+deb.sury.org~trusty+1 (cli) (built: Dec 21 2014 19:56:00) 
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.5.0, Copyright (c) 1998-2014 Zend Technologies
    with Zend OPcache v7.0.4-dev, Copyright (c) 1999-2014, by Zend Technologies

Tested in Phalcon 1.3.4 and 1.3.5

**Description:**
Consider an action with two optional parameters
```php

public function editAction($var1=null,$var2=null){
   //... 
}

```
**Steps to reproduce:**

1) When passing an url paramter like this:

    /namespace/controller/edit/0

**Expected behavior:**
The expected behavior is that $var1 == 0

**Actual behavior:**
$var1 == null

2) **HOWEVER** when passing two parameters the expected result equals the actual result:

    /namespace/controller/edit/0/0

Expected (and actual) result:
$var1 == 0, $var2 == 0

I also have a thread on the phalcon forum about the issue: http://forum.phalconphp.com/discussion/4554/phalcon-converts-zero-to-null-when-passed-as-last-parameter-in-u

7thcubic:
actually if u do
/0/ 
it will work too. cause strsplit it will give 1 params

fsando:
@7thcubic
re strsplit, unfortunately I don't know c. I searched all the code of cphalcon (v1.3.5) for "strsplit" but didn't find anything.

7thcubic:
https://github.com/phalcon/cphalcon/blob/1.3.5/ext/mvc/router.c#L944

fsando:
Thanks for taking so much time.
I don't know c so I don't really know what I'm looking for or what I'm looking at but looking here

https://github.com/phalcon/cphalcon/blob/1.3.5/ext/kernel/string.c#L313

doesn't seem to be an issue assuming that php_explode() behaves like php's explode(). 

Remember that these two urls produce the same result
ns/crtl/act
ns/crtl/act/0

While these two produce different results:
ns/crtl/act
ns/crtl/act/1

Further, if you look at the output of 
router->getMatches()
You'll see that ns/cctrl/act/0 results in the last entry having the string "/0"

[...]

PR #3277, opened 2015-01-24
Phalcon v1.3.x deletes url parameters in some situations #3274 - possible solution

fsando:
As described here: https://github.com/phalcon/cphalcon/issues/3274
And discussed in the forum here: http://forum.phalconphp.com/discussion/4554/phalcon-converts-zero-to-null-when-passed-as-last-parameter-in-u#C15206

There is an issue where Phalcon converts a single url param to null if the param is 0.
Thanks to 7thcubic who pointed to the possible problem area 
https://github.com/phalcon/cphalcon/blob/2.0.0/phalcon/mvc/router.zep#L575

I have created a PR that may solve this.


phalcon:
This have some many commits

fsando:
> fsando wants to merge 2,297 commits

I don't know how that happened, there was one small change - 5 lines, one commit. Sorry about that,

PR #3279, opened 2015-01-24
Check length of strParams and don't use it as a condition

olivier-monaco:
Issue #3274

### DIFF
--- a/phalcon/mvc/router.zep
+++ b/phalcon/mvc/router.zep
@@ -574,7 +574,7 @@ class Router implements RouterInterface
 			 */
 			if fetch paramsStr, parts["params"] {
 				let strParams = substr(paramsStr, 1);
-				if strParams {
+				if strlen(strParams) > 0 {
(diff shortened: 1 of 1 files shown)


olivier-monaco:
Better code on #3280

phalcon/mvc/router.zep
```diff
@@ -574,7 +574,7 @@ class Router implements RouterInterface
 			 */
 			if fetch paramsStr, parts["params"] {
 				let strParams = substr(paramsStr, 1);
-				if strParams {
+				if strlen(strParams) > 0 {
```
sergeyklay: or `strParams->length()` :)

olivier-monaco: But substr will return "false" when paramsStr is only 1 character length. So pull request #3280 is better...

sergeyklay: :+1:

PR #3280, opened 2015-01-25
Don't lose uri params when param==0 (#3274)

fsando:
https://github.com/phalcon/cphalcon/issues/3274


fsando:
I changed the patch slightly so the only change from the original code is that a weak test for falsyness is changed to strict test for false.

PR #3286, opened 2015-01-26
[2.0.0] Image crop bug

tmihalik:
0 can be a desired value for offsets, e.g. if you want to crop from top of the image

``` php
$image->crop(150, 193, null, 0); //crop horizontally center, vertically top 
```

It was OK in 1.3:
https://github.com/phalcon/cphalcon/blob/1.3.4/ext/image/adapter.c#L429

### DIFF
--- a/phalcon/image/adapter.zep
+++ b/phalcon/image/adapter.zep
@@ -185,7 +185,7 @@ abstract class Adapter
  	 */
 	public function crop(int width, int height, int offset_x = null, int offset_y = null) -> <Adapter>
 	{
-		if !offset_x {
+		if is_null(offset_x) {
 			let offset_x = ((this->_width - width) / 2);
 		} else {
 			if offset_x < 0 {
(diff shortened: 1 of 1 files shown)


phalcon:
Thanks

tmihalik:
@phalcon 
@andresgutierrez 
There's still some problem with this crop method.
It's impossible to pass `null` value to `offsetX` or `offsetY` params, because zephir makes `null` to `0`.

```php
$image->crop(150, 193, null, null); //crop from both horizontally and vertically center (should be but nulls becomes 0)
$image->crop(150, 193, 0, 0); //crop from horizontally left and vertically top
```

tmihalik:
ping @andresgutierrez

tmihalik:
ping @sergeyklay

sergeyklay:
@tmihalik Create please separated issue with scrip to reproduce

Issue #3271, opened 2015-01-22
PHALCON_INIT_VAR in 2.0.0


olivier-monaco:
See:
https://github.com/phalcon/cphalcon/blob/2.0.0/ext/phalcon/annotations/base.c#L82
https://github.com/phalcon/cphalcon/blob/2.0.0/ext/phalcon/annotations/base.c#L374
https://github.com/phalcon/cphalcon/blob/2.0.0/ext/phalcon/annotations/base.c#L410

maxgalbu:
What about it? Should MAKE_STD_ZVAL be used instead?

olivier-monaco:
I think so but not sure. I've planed to dig into but did not find time...

olivier-monaco:
Hum... ext/phalcon/annotations/base.c seems to not be compiled. Introducing some compilation error don't stop compilation.

Issue #3195, opened 2015-01-08
Phalcon 2 hangs and keeps loading infinitely


4ley:
I have Phalcon 2, PHP5-FPM and Ngnix installed.
When I execute `SomeModel::findFirst('{{ query.var }}');` PHP starts to load infinitely. It seems Phalcon is trapped into something. Obviously '{{' is killing the parser. I came into the issue, when accidently passed this string through URL params.
This error wasn't in Phalcon 1.

olivier-monaco:
This issue happens for any syntax error in phql. There seems to be a ZEPHIR_MM_GROW missing at start of phql_internal_parse_phql function.

Patch:
<pre>
diff --git a/ext/phalcon/mvc/model/query/parser.c b/ext/phalcon/mvc/model/query/parser.c
index 031bc88..fd4e3e6 100644
--- a/ext/phalcon/mvc/model/query/parser.c
+++ b/ext/phalcon/mvc/model/query/parser.c
@@ -3077,6 +3077,8 @@ int phql_internal_parse_phql(zval **result, char *phql, unsigned int phql_length
        char *error;
        zval **temp_ast;

+       ZEPHIR_MM_GROW();
+
        if (!phql) {
                ZEPHIR_INIT_VAR(*error_msg);
                ZVAL_STRING(*error_msg, "PHQL statement cannot be NULL", 1);
</pre>

maxgalbu:
I don't think the memory management functions are used in that file, from what I know those macros (*_MM_GROW, *_MM_RESTORE) are only needed when there are zvals in the body of the function. There is one zval but is initialized with ZVAL_ZVAL

olivier-monaco:
Normal compilation:
<pre>
*** glibc detected *** /usr/sbin/apache2: free(): invalid pointer: 0x00007f36cb4fab5f ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x76a16)[0x7f36c9265a16]
/lib/x86_64-linux-gnu/libc.so.6(cfree+0x6c)[0x7f36c926a7bc]
/usr/lib/php5/20121212/phalcon.so(+0x2b0e56)[0x7f36b98c8e56]
/usr/lib/apache2/modules/libphp5.so(+0x3a65a4)[0x7f36c5f325a4]
...
</pre>

With "-O0 -g3 -Wall": no more bug...

olivier-monaco:
File query.zep, line 2143:
<pre>
[...]

PR #3294, opened 2015-01-28
Sync ext/phalcon/annotations/base.c with parser.c and add generate file

olivier-monaco:
Issue  #3271


phalcon:
We absolutely need this file

olivier-monaco:
Phalcon 2 includes 3 base.c files:
<pre>
./ext/phalcon/annotations/base.c
./ext/phalcon/mvc/view/engine/volt/base.c
./ext/phalcon/mvc/model/query/base.c
</pre>

Only 2 of then seems to be used for parsers. Ouput from a grep:
<pre>
ext/phalcon/mvc/view/engine/volt/generate:re2c -o scanner.c scanner.re && ./lemon -s parser.lemon && cat base.c >> parser.c
ext/phalcon/mvc/model/query/generate:re2c -o scanner.c scanner.re && ./lemon parser.lemon && cat base.c >> parser.c
</pre>

The file ext/phalcon/annotations/base.c is in the "exclusions" list of "scripts/gen-build.php".

Phalcon 2 compiles anyway.

Are you sure it's needed?

olivier-monaco:
If it works like other base.c files, there's a generate file missing.

We can also see old code from base.c which is not in parser.c:
- use of PHALCON_INIT_VAR
- use of PHALCON_THROW_EXCEPTION_STR

olivier-monaco:
But now we may have the same problem than #3195...

olivier-monaco:
We have... Use case:
<pre>
<?php

/**
 * @Invalid(")
 */
class MyTest {
}

$reader = new \Phalcon\Annotations\Adapter\Memory();

$reflector = $reader->get('MyTest');
$annotations = $reflector->getClassAnnotations();
foreach ($annotations as $annotation) {
    echo $annotation->getName(), PHP_EOL;
    echo $annotation->numberArguments(), PHP_EOL;
    print_r($annotation->getArguments());
}
</pre>

olivier-monaco:
Rewriting this pull request from scratch.

PR #3299, opened 2015-01-29
[2.0.0] Forms: Fix entity binding before validation

brainformatik:
If entity is not passed to isValid method but previously set via constructor we should bind this one.

Now it's additionally possible:

``` php
$robot = Robots::findFirst();
$form = new Form($robot);

// works because we use for entity binding the model which is passed to constructor
if ($form->isValid($_POST)) { 
    $robot->save();
}
```


andresgutierrez:
Thanks

PR #3300, opened 2015-01-29
[2.0.0] HTTP Request: Fix header field handling

brainformatik:
1) 
Method **getHeader** extended to allow to retrieve values with regular HTTP header names like Accept-Header, Content-Type etc. It's compatible with 1.3.x code. 
_(Because the method getHeaders returns an array with keys that has regular HTTP header names. We should have the possibility to retrieve one value via HTTP header name from getHeader)_

2)
Method **getHeaders** refactored to add common used HTTP header field names, like Content-Type and Content-Length. It's compatible with 1.3.x code.


phalcon:
Thanks

Issue #3295, opened 2015-01-29
[2.0] Segault in annotation parser when annotation is malformed


olivier-monaco:
See #3294
Like #3195 

Use case:
<pre>
/**
 * @Invalid(")
 */
class MyTest {
}

$reader = new \Phalcon\Annotations\Adapter\Memory();

$reflector = $reader->get('MyTest');
$annotations = $reflector->getClassAnnotations();
foreach ($annotations as $annotation) {
    echo $annotation->getName(), PHP_EOL;
    echo $annotation->numberArguments(), PHP_EOL;
    print_r($annotation->getArguments());
}
</pre>

olivier-monaco:
Optimizer don't manager failure of annotation parse.

PR #3314, opened 2015-02-02
Segault in annotation parser when annotation is malformed

olivier-monaco:
Issue #3295 (rewrite of pull request #3294)


olivier-monaco:
Handle also #3271

Issue #3222, opened 2015-01-13
Small issue with urls from \Phalcon\Mvc\Url


dompie:
In the router I use:

```php
$router->addGet('/activation/{activationToken:([a-zA-Z0-9_]{19,45})}', array(
	'controller' => 'index',
	'action' => 'activation',
))->setName('activationUrl');
```

When generating an URL for this action with the following code:

```php
$url = Di::getDefault()->get('url'); // This is \Phalcon\Mvc\Url
$activationUrl = $url->get(array(
	'for' => 'activationUrl',
	'activationToken' => $activationToken
));
```
The generated url looks like this for base uri `'http://' . $_SERVER['SERVER_NAME']` (which is "www.local.dev"):

`http://www.local.devactivation/u9o6xzwql3pirc42k9m04fv`

And I can not add a trailing slash after $_SERVER['SERVER_NAME'], because this leads to multiple slashes in some cases.
Can this be fixed by reconfiguration?

I use phalcon 1.3.4

dompie:
Could workaround this issue by changing:
```php
$router->addGet('/activation/{activationToken:([a-zA-Z0-9_]{19,45})}', array(
```
to
```php
$router->addGet('//activation/{activationToken:([a-zA-Z0-9_]{19,45})}', arr...
```

Other suggestions / solutions welcome.

7thcubic:
what did you set up for ur base url?

$di->set('url', function () use (&$config) {
    $url = new UrlResolver();
    $url->setBaseUri($config->application->baseUri); // << this one
    return $url;
}, true);

dompie:
I use $_SERVER['SERVER_NAME'] for baseUri

7thcubic:
If you configured your base url to be

http://.$_SERVER["SERVER_NAME"]."/"

when ever you use the functionality from url. do not include the prevailing slash.
eg.

$url->get("index/login");

if its http://.$_SERVER["SERVER_NAME"]

do it as 

$url->get("/index/login");

so you really have to set the standard of coding.

dompie:
That's what I'm saying: I already use 
```php
define('BASE_URI', 'http://' . $_SERVER['SERVER_NAME']);
...
$url->setBaseUri(BASE_URI);
```
and I define all my routes like this *always*

```php
$router->addGet('/activation/{activationToken:([a-zA-Z0-9_]{19,45})}', array(
	'controller' => 'index',
	'action' => 'activation',
))->setName('activationUrl');

[...]

PR #3315, opened 2015-02-02
URL::get: Avoid double slash when joining baseUri to provided uri

olivier-monaco:
Issue #3222


sergeyklay:
phalcon/mvc/url.zep 100644 → 100755

olivier-monaco:
Sorry, I definitely need to fix my samba issues ;).

sergeyklay:
:+1: Good work

PR #3342, opened 2015-02-19
[2.0.0] Model/Validator/Url: Comparison fix

brainformatik:
If using validator of type url you get an error, because the comparison is wrong here 

```
typeof field == "string"  => throw Exception ("Field name must be a string")
```

It is fixed to 

```
typeof field != "string"
```


andresgutierrez:
Thanks

PR #3345, opened 2015-02-19
[2.0.0] Add __isset magic method to view for view params

onshorechet:
Adds the magic isset method the the view object so that view params can be checked with isset().

ie:
echo isset($this->view->products);

### DIFF
--- a/phalcon/mvc/view.zep
+++ b/phalcon/mvc/view.zep
@@ -1442,7 +1442,7 @@ class View extends Injectable implements ViewInterface
      * @param string key
      * @return boolean
      */
-    public function __isset(string! key)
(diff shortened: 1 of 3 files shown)


Timethor:
+1

andresgutierrez:
Could you please rebase against 2.0.0?

chet0xhenry:
With pleasure.

chet0xhenry:
I added the boolean return type in the method declaration, but you don't want the @params and @return in the method comment.  It won't match the style of this file.  It doesn't bother me to do that, I just wanted confirmation.

andresgutierrez:
Thanks

PR #9914, opened 2015-03-05
[2.0.0] Fix compatibility with Micro\Collection::map() when no verb used

ovide:
The method map() sends a null verb to _addMap and it's fine when you add a controllers as handlers.

This worked with 1.\* but in 2.0 it's casted to an empty string.
This patch solves the problem when the route tries to add null verbs (route->via(""))


andresgutierrez:
Thanks

PR #9919, opened 2015-03-09
[2.0.0] Small Logger review

ovr:
- remove unneeded docblocks
- NS - import Logger
- added inline for log (will speedup soon)
- context must be an array!


ovr:
cC @phalcon @andresgutierrez

andresgutierrez:
Thanks

PR #9927, opened 2015-03-10
Minor fix in method's doc

ppietak:
Method "hasChanged" takes string as a parameter, and returns boolean value.


andresgutierrez:
Thanks

PR #9929, opened 2015-03-11
[2.0.0] Updated Memcached client option to optional

kjdev:
unit-tests/CacheTest.php(libmemcached) becomes Segmentaion Fault.

It seems there is no client set to option of test. 

If you are not a required option than to add a client option to test has been fixed.


andresgutierrez:
Thanks

PR #9940, opened 2015-03-20
[2.0.0] Reduced the complexity of various methods

SidRoberts:


SidRoberts:
I think it's failing due to https://travis-ci.org/phalcon/zephir/builds/55272761. On my machine, it compiles and all tests relating to these classes are passing.

PR #9947, opened 2015-03-25
Review Crypt

ovr:
- remove unneeded docblocks
- added more return type hints


ovr:
ping @phalcon @andresgutierrez

ovr:
Thanks

PR #9949, opened 2015-03-25
[2.0.0] Fixed type checking for model messages

SidRoberts:
With `if typeof message != "object" { ... }`, `message` has to be an object in order to use the `setModel()` method.

With `if typeof record == "object" { ... }`, we know that `record` will be an object because we throw an Exception if it's not on line 2434 and we still need to check that `message` is an object to use the `setModel()` method.

### DIFF
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -2458,7 +2458,7 @@ abstract class Model implements ModelInterface, ResultInterface, InjectionAwareI
 							/**
 							 * Set the related model
 							 */
-							if typeof record == "object" {
+							if typeof message == "object" {
 								message->setModel(record);
 							}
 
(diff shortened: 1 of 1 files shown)


andresgutierrez:
Actually checking if record is ok, maybe it needs both conditions?

SidRoberts:
No, `record` will always definitely be an object as we check on [line 2434](https://github.com/phalcon/cphalcon/pull/9949/files#diff-95da5fc282ff73cf517e37ee54f1613bR2434) and we're also in a [for loop](https://github.com/phalcon/cphalcon/pull/9949/files#diff-95da5fc282ff73cf517e37ee54f1613bR2456) that depends on `record` being an object: `for message in record->getMessages() { ... }`. :smile:

andresgutierrez:
Right!

PR #9952, opened 2015-03-26
[2.0.0] Tidied up the Image component

SidRoberts:
All variables are now camel-cased and I've removed all the Docblocks (param/return statements) that I can without affecting the IDE stubs output.


andresgutierrez:
Thanks

PR #9953, opened 2015-03-27
[1.3.5] [minor] Bad return type in getPaginate() for all Paginator\Adapter's

Flygenring:
The stubs are declared with `stdClass` as return type, which expands in their namespace to be `Phalcon\Paginator\Adapter\stdClass`, which in turn does not exist.
Quick fix to change it to `\stdClass`.

### DIFF
--- a/ext/paginator/adapter/model.c
+++ b/ext/paginator/adapter/model.c
@@ -108,7 +108,7 @@ PHP_METHOD(Phalcon_Paginator_Adapter_Model, setCurrentPage){
 /**
  * Returns a slice of the resultset to show in the pagination
  *
- * @return stdClass
+ * @return \stdClass
  */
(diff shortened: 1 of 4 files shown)


andresgutierrez:
Thanks

Flygenring:
You're very welcome :)
My local fix to make auto-complete, hinting, etc. work properly was (a big no-no) to make a completely new class, but I'm not deeply enough into the C-way, so it's just PHP:

```php
namespace Phalcon\Paginator {

	class Page {

        /** @type stdClass[] $items */
        public $items;

        /** @type int $before */
        public $before;

        /** @type int $first */
        public $first;

        /** @type int $next */
        public $next;

        /** @type int $last */
        public $last;

        /** @type int $current */
        public $current;

        /** @type int $total_pages */
        public $total_pages;

        /** @type int $total_items */
        public $total_items;

	}
}
```

PR #9959, opened 2015-03-31
Fix «in» operator because of strpos «0» answer

dugwood:
strpos returns 0 if the needle is found at position 0. So the test should be made against «false», and not return the position of the string, which won't work for examples like 'test' in 'test1234'.


andresgutierrez:
Thanks

PR #9971, opened 2015-04-07
[2.0.0] DB Docblock Cleanup

SidRoberts:
I've removed a lot of the unnecessary docblocks from the DB component. I've purposely left alone any docblocks that don't have Zephir type hinting as I don't want to change the behaviour of the methods.


andresgutierrez:
Thanks

PR #9978, opened 2015-04-08
[2.0.0] fixed return null at Phalcon\Mvc\Model\Resultset:toArray() when the resu...

alu:
fixed return null at Phalcon\Mvc\Model\Resultset:toArray() when the result is empty.

### DIFF
--- a/phalcon/mvc/model/resultset/simple.zep
+++ b/phalcon/mvc/model/resultset/simple.zep
@@ -228,6 +228,8 @@ class Simple extends Resultset
 					 * Update the row count
 					 */
 					let this->_count = count(records);
+				} else {
+					let records = [];
 				}
 			}
 		}


andresgutierrez:
Phalcon 1.3.x always return an array no matter if the resultset is empty or not

alu:
@andresgutierrez
This commit is intended to match 2.0.0 with 1.3.x.
Check [my code](https://github.com/phalcon/cphalcon/blob/a92db5fb31133b5840f50e4a870c3b52b111aa33/phalcon/mvc/model/resultset/simple.zep#L207-L234) AND [original code](https://github.com/phalcon/cphalcon/blob/master/ext/mvc/model/resultset/simple.c#L323-L358).

andresgutierrez:
Thanks

PR #9980, opened 2015-04-08
[NFR] Phalcon\Tag class extension by closure 

kjdev:
Added Phalcon\Tag::addFunction() and hasFunction() and calls from Volt Compiler.

``` php
Phalcon\Tag::addFunction('shuffle', function($str){
    return str_shuffle($str);
});

$volt = new Phalcon\Mvc\View\Engine\Volt\Compiler();
$compilation = $volt->compileString('{{ shuffle("hello") }}');
// compilation: <?php echo $this->tag->shuffle('hello'); ?>
```

### DIFF
--- a/phalcon/mvc/view/engine/volt/compiler.zep
+++ b/phalcon/mvc/view/engine/volt/compiler.zep
@@ -618,6 +618,15 @@ class Compiler implements InjectionAwareInterface
 			}
 
 			/**
+			 * Check if it's a function in Phalcon\Tag
+			 */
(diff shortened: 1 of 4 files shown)


andresgutierrez:
I think this violates the [single responsibility principle](http://en.wikipedia.org/wiki/Single_responsibility_principle) of Phalcon\Tag as its goal is not that

kjdev:
Thanks.
Cancel to Pull Request .

Issue #9930, opened 2015-03-11
Markdown Support Request in Phalcon 2.0


ghost:
Request for markdown support in phalcon. So we can use something like this;
Preferably should support all of this; https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet

``` php
'.md' => function($view, $di) {
    $markdown = new \Phalcon\Mvc\View\Engine\Markdown($view, $di);
    $markdown->setOptions([
        'compiledPath' => '../app/cache/',
        'compiledExtension' => '.md',
    ]);
    return $markdown;
}
```

Thank You!

maxgalbu:
But how do you interpolate variables from php into markdown? Php, Twig, smarty all have a syntax for it

ovr:
It's not needed in main core, but you can implement it in your library
and share on GitHub

wetux:
If I could implement it my self I would request it here :) Implementing markdown in phalcon core will be very useful. Because more and more people are switching to static site generators and having one in phalcon core would just make the life that much easier for phalcon developers. I would guess it would also be faster then using separate library.

kjdev:
It seems there is no need in the main core, but I tried incorporated into the trial.

I tried crowded built a [php-ext-hoedown](https://github.com/kjdev/php-ext-hoedown.git).

repos: https://github.com/karakurihiden/cphalcon/tree/2.0.0-markdown

patch file: [phalcon-2.0.0-view-engine-markdown.patch](https://gist.github.com/kjdev/2ecc58afe95e9ffbf105)

``` php
use Phalcon\Mvc\View\Engine\Markdown as Markdown;

'.md' => function($view, $di) {
    $markdown = new Markdown($view, $di);
    $markdown->setOptions([
        'compiledPath' => '../app/cache/',
        'compiledSeparator' => '_',
        /*
        // Default enable options
        'compileOption' => [
            Markdown::TABLES, Markdown::FENCED_CODE, Markdown::AUTOLINK,
            Markdown::STRIKETHROUGH, Markdown::NO_INTRA_EMPHASIS,
            Markdown::TASK, Markdown::SCRIPT_TAGS,
        ],
        */
        /*
        // Set normal Markdown
        'compileOption' => [],
        */
        /*
        // Set user options
        'compileOption' => [
            Markdown::TABLES, Markdown::FENCED_CODE, Markdown::AUTOLINK,
[...]

PR #9983, opened 2015-04-08
[NFR] Added Phalcon\Mvc\View\Engine\Markdown

kjdev:
see: #9930


andresgutierrez:
Thanks, we currently have code freeze for 2.0.0, only PR with bug fixes will be accepted

kjdev:
All right.
Please consider also after the 2.0.0 release.

wetux:
This please add this feature at least after 2.0.0 released, because this is very useful to many people who use Markdown. There is also something called ParseDown.org which is even faster.

andresgutierrez:

Issue #2927, opened 2014-10-21
Confirmation validator is able to compare strings in case insensitive mode


Samhayn:
Hello, phalcon team.

It would be great to add an option to Confirmation validator, to switch on/off case-sensitive mode of comparing 2 values.

**Example:**
new Confirmation(array('with' => 'email', 'message' => 'please, confirm your email', 'caseSensitive' => false));

**Reason:**
Such things as a email agents are case insensitive.

**Additional info:**
I know, that it's possible to just extend standard validator (and I'll do it), but, I think it's better to support such simple configs "from the box"

andresgutierrez:
Implemented in 2.0.x

PR #9987, opened 2015-04-08
[2.0.0] new feature  #2927

KorsaR-ZN:


andresgutierrez:
I think the new behaviour should be only enabled passing a new option instead of changing the default behaviour.

KorsaR-ZN:
@andresgutierrez 
The default behavior not has changed. It changes only when the option "caseSensitive" == false

andresgutierrez:
Thanks

Issue #3247, opened 2015-01-19
Data not update when update value from integer to decimal


ohmcoe:
I use phalcon 1.3.4 build from source.
database: mysql
And use models with useDynamicUpdate(true);

Example.
When update model value from 4 to 4.00 (type of field in database is longtext) the value in database is not change.

sergeyklay:
@ohmcoe Can you give an small example of your code please? Also provide table schema.

ohmcoe:
Table export from mysql

```
CREATE TABLE IF NOT EXISTS `test` (
`id` int(10) unsigned NOT NULL,
  `value` longtext NOT NULL
) ENGINE=InnoDB  DEFAULT CHARSET=utf8 AUTO_INCREMENT=2 ;

INSERT INTO `test` (`id`, `value`) VALUES
(1, '4');

ALTER TABLE `test`
 ADD PRIMARY KEY (`id`);
```

This models file.
```
class Test extends \Phalcon\Mvc\Model
{
    public $id;
    public $value;

    public function initialize()
    {
        $this->useDynamicUpdate(true);
    }
}
```

this code to change value
```
      $test = Test::findFirst("id='1'");
      $test->value = '4.00';
      $test->save();
```

the record in table test id 1 value is not change from 4 to 4.00

olivier-monaco:
Your problem is here: https://github.com/phalcon/cphalcon/blob/2.0.0/phalcon/mvc/model.zep#L2317

The beauty of low typed language is to be able to compare apple with birds... Try this and you're be suprised !
<pre>
$a = '4';
$b = '4.0';
var_dump($a, $b, $a == $b, $a === $b);
</pre>

But the bugfix is not so easy... What about putting an INT column with 4 as value replaced by '4'? No update is required...

olivier-monaco:
I try do review code of 2.0.0 to better detect changes. But maybe it is too strict now...

olivier-monaco:
Use the last 2.0.0. This bug is fixed.

Green-Cat:
@andresgutierrez Fixed in #9996

PR #9996, opened 2015-04-08
Data not update when update value from integer to decimal (again)

olivier-monaco:
Issue #3247

Strict type comparisons to detect changes. Need someone to check my code...

@andresgutierrez This is the new pull request.


andresgutierrez:
Thanks

Issue #9974, opened 2015-04-07
Phalcon\Http\Response _statusCodes


Kachit:
Please move status codes with messages to constants class for example: https://github.com/teapot-php/status

PR #9997, opened 2015-04-08
Proposal: enum like Java Enums

olivier-monaco:
I'm using this code for year to use a PHP class as en enum like in Java.

Here is the Phalcon\Enum class, a base class for enum.Child classes must implements the static _init method to returns an array of class instances.

```
class MyEnum extends Phalcon\Enum
{
    protected static function _init()
    {
        return array(
            new self('First'),
            new self('Second'),
        );
    }
 }
```

Then, each instance can be accessed using a static method equals to the instance
name (ie. the constructor argument).

```
MyEnum::First()
```

The same instance if returned for each call so we can use the = operator.

Child classes can add other properties.

```
class Country extends Phalcon\Enum
{
    private $_code;
    private $_language;

    protected function __construct($name, $code, $language)
    {
        parent::__construct($name);
        $this->_code = $code;
        $this->_language = $language;
    }

    public function getCode() ...

    public function getLanguage() ...

    protected static function _init()
    {
        return array(
            new self('France', 'fr_FR', 'French'),
            ...
        );
    }
}
```

This is an example:

```
<?php

class MyEnum extends Phalcon\Enum
{
    protected static function _init()
    {
        return [
            new self('First'),
            new self('Second')
        ];
    }
}

$a = MyEnum::First();
$b = MyEnum::Second();
$c = MyEnum::First();
$d = $a;
$e = MyEnum::valueOf('First');

echo '$a / $b'.PHP_EOL;
var_dump($a == $b, $a === $b); // false, false
echo '$a / $c'.PHP_EOL;
var_dump($a == $c, $a === $c); // true, true
echo '$a / $d'.PHP_EOL;
var_dump($a == $d, $a === $d); // true, true
echo '$a / $e'.PHP_EOL;
var_dump($a == $e, $a === $e); // true, true

var_dump(MyEnum::values()); // array with 2 instances

var_dump($a == 'First'); // true
```

It's more useful than constant because you can add "data" to each value and you can use type checking.

Implementing #9974 could be simple and powerful like:

```
class StatusCode extends Phalcon\Enum
{
    private $_label;
    private $_code;

    protected function __constuct($name, $label, $code) ...

    public getLabel() ...

    public getCode() ...

    protected static function _init()
    {
        return array(
            new self('CONTINUE', 'Continue', 100),
            new self('SWITCHING_PROTOCOLS', 'Switching Protocols', 101),
            new self('PROCESSING', 'Processing', 102),
            new self('OK', 'OK', 200),
            ...
        );
    }
}
```

Then the enum can replace the statusCodes array in https://github.com/phalcon/cphalcon/blob/2.0.0/phalcon/http/response.zep#L146.

And the Response::setStatusCode can accept a StatusCode instance instead of an int in https://github.com/phalcon/cphalcon/blob/2.0.0/phalcon/http/response.zep#L121


andresgutierrez:
Thanks, we currently have code freeze for 2.0.0, only PR with bug fixes will be accepted

olivier-monaco:
@andresgutierrez I can make a new PR with this feature. Do you want?

PR #9998, opened 2015-04-08
Fixed compile warning noneexistent-class

kjdev:
Warning: Class "Phalcon\Validation\Validator\Exception" does not exist at compile time in cphalcon/phalcon/validation/validator/confirmation.zep on 94 [nonexistent-class]


andresgutierrez:
It's better to add at the top:

```zephir
use Phalcon\Validation\Exception;
```

kjdev:
Thanks.

PR #9999, opened 2015-04-08
[2.0.0] [BUG] fixed no import Phalcon\Validation\Exception

KorsaR-ZN:


andresgutierrez:
This was fixed in #9998

PR #10002, opened 2015-04-09
[2.0.0] [BUG] fixed compatibility with 1.3.x

KorsaR-ZN:
Added translate adapter - Gettext and more enhancements


andresgutierrez:
Thanks!

PR #10005, opened 2015-04-09
[2.0.0] Gets Criteria instance from DI container

KorsaR-ZN:


andresgutierrez:
I think also this can be moved there: 

```php
https://github.com/phalcon/cphalcon/blob/2.0.0/phalcon/mvc/model/manager.zep#L1820
https://github.com/phalcon/cphalcon/blob/2.0.0/phalcon/mvc/model/manager.zep#L1847
```

KorsaR-ZN:
@andresgutierrez 
I think so too.
I'll now do it

KorsaR-ZN:
@andresgutierrez 
Who broke the CI build?
My build is passed
https://travis-ci.org/KorsaR-ZN/phalcon-core/builds/57862567

andresgutierrez:
Thanks

PR #10019, opened 2015-04-12
Debug docblock cleanup

Green-Cat:


Green-Cat:
@andresgutierrez fixed.

andresgutierrez:
Thanks

Issue #3209, opened 2015-01-11
Pagination::getPaginate() doesn't contain limit in result object


fizzka:
It is not big problem, but it'll be more accurate to provide full information to result $page object.

7thcubic:
i presume you are talking about the total amount of items in the database/array right?

fizzka:
I think we are talking about different subjects.

I mean _$limit_ from this construction:

```php
//from docs
$paginator = new \Phalcon\Paginator\Adapter\Model(
    array(
        "data" => $robots,
        "limit"=> 10,
        "page" => $currentPage
    )
);
```

When I call
```php
//from docs
$page = $paginator->getPaginate();
```

or pass reslults to view:

```php
$this->view->page = $paginator->getPaginate(); 
```

I can't access _"limit"_ ( => 10) property from _$page_.

Workaround I found is:

```php
//set limit
$limit = 10;
$paginator = new \Phalcon\Paginator\Adapter\Model(
    array(
        "data" => $robots,
        "limit"=> $limit,
        "page" => $currentPage
    )
);

$page = $paginator->getPaginate();
$page->limit = $limit;

$this->view->page = $page;
```

I suppose result must contain _limit_ value already ;)

sergeyklay:
@fizzka 
I too use this workaround :)
also you can

```php
        $this->view->setVars([
            'page' => $page,
            'limit' => $limit
        ]);
```

fizzka:
@sergeyklay 
where are many ways to get success, but I raised issue couse I'm sure that result _page_ object must be added with _limit_ =)

It is clean way - result object must incapsulate all neccessary data.

7thcubic:
I had suspected it was the manual limit. 

If it is that i suggest the work around. It isnt that much of a deal breaker.

Even if you used a global service to return the paginator object. The global service would already contain the limit information and could pass it back to you similarly. And since you initialize it on the fly, the limit info is something you already have, so it is my guess that it was intended to be this way. 

Again though extending pagination and adding a custom magic __get for limit isnt impossible. 

One scenario that i can think off when u pass the entire pagination object into view and you are lazy to pass limit too. I guess that would be needed. 

But most pagination applied to web normally allow the users to set the limit they want. 5/10/50 in the get parameter. So again there is redundancy.

Green-Cat:
Implemented in the merged pull request #10020.
@phalcon @andresgutierrez could you close this issue please.

PR #10020, opened 2015-04-12
Implements ability to retrieve limit from paginate object

Green-Cat:
Adds a property called "limit" to the paginate object of the paginator in all three adapters.
This is rather useful, especially to display the limit in the view.
Related issue #3209.
The travis build error is not from this pull request, it's from a commit on the 2.0 branch. See #10021.


Green-Cat:
Closed and reopened to have travis rerun.

PR #10025, opened 2015-04-13
Crypt Chaining Fix & Docblock Cleanup

Green-Cat:
Generally cleans up the unneeded docblocks but also:
- removes mentions of the non-existing EncryptInterface
- makes the interface match implementation in terms of variable types
- fixes setter chaining when using setPadding method

### DIFF
--- a/phalcon/crypt.zep
+++ b/phalcon/crypt.zep
@@ -19,6 +19,7 @@
 
 namespace Phalcon;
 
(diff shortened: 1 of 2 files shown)


Green-Cat:
@andresgutierrez fixed.

Green-Cat:
@phalcon @andresgutierrez ping

Issue #3170, opened 2014-12-29
[Phalcon 2] setViewsDir does not add trailing slash


olivier-monaco:
While porting Phalcon\Mvc\View::setViewsDir to Zephir, the following line was removed:
<pre>
	phalcon_add_trailing_slash(views_dir);
</pre>

So, the following code does not work as before:
<pre>
$view->setViewsDir('views');
</pre>

It must become:
<pre>
$view->setViewsDir('views/');
</pre>

andresgutierrez:
Could you please try again using latest Phalcon 2?

PR #10028, opened 2015-04-13
Adding final DIRECTORY_SEPARATOR if needed

olivier-monaco:
#3170


olivier-monaco:
Only tested on linux box. Using DIRECTORY_SEPARATOR for cross plateform...

Issue #6788, opened 2015-02-25
[2.0.0] Mvc/Model: Decimal and null fileds cause '0' values in unrelated columns


meteoblue:
Hello,

I recently upgraded my development system from 1.3.4 to newest 2.0.0. One mvc/model (mysql) now overwrites certain columns to '0' while creating.
I reduced the table structure and code to a minimum. If I change the order of columns the issue does not appear. It requires a null-able, a decimal and a varchar/text/char column.
If decimal is set, the char column will store '0' in the database instead of 'test'.

Could someone try to reproduce this issue? Thanks!

Regards,
Patrick

```SQL
CREATE TABLE `test` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `null_field` int(1) DEFAULT NULL, 
  `decimal` decimal(2,1) NOT NULL,
  `char` char(10) NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB  DEFAULT CHARSET=utf8;
```

```PHP
class Test extends \Phalcon\Mvc\Model { }

$t = new Test();
$t->decimal = 3; // has to be set
$t->char = 'test';
$t->save(); // 'char' is stored as '0' in DB, but local scope is still 'test'
$t->refresh(); // after reload, 'char' is now '0'

if($t->char != 'test') { 
    echo 'BUG';
} else {
    echo 'OK';
}
```

olivier-monaco:
See #3247. I think it's related. Can you use the latest 2.0.0?

olivier-monaco:
Oups... wrong issue... I'll take a look...

olivier-monaco:
Problem comes from Phalcon\Db\Adapter::insert method. Use this code:
<pre>
$values = [null, 3, 'test'];
$fields = ['null_field', 'decimal', 'char'];
$types = [1024, 32, 2];
$cnx->insert('test2', $values, $fields, $types);
</pre>

And the bindDataTypes passed to the execute method will be: [1024, 32, 2, 32, 2]...

olivier-monaco:
Please try with latest 2.0.0.

meteoblue:
Works! Thanks!

PR #10031, opened 2015-04-13
Phalcon\Db\Adapter::insert build bas data types array

olivier-monaco:
See #6788


andresgutierrez:
But this removes the bind types that come from parameters

olivier-monaco:
The array is then built by the following for. See line 335 where bindDataTypes is effectively build, field by field.

Keeping the initialization results in duplicate types.

olivier-monaco:
Probably same problem in update function... I need to test.

Are you okay with my pull request?

andresgutierrez:
Thanks

PR #10036, opened 2015-04-14
Acl Docblock Cleanup

Green-Cat:


andresgutierrez:
Build is broken, could you please check?

Green-Cat:
@andresgutierrez it seems like Phalcon\Acl\Role does not implement RoleInterface.
Is that by design?

Green-Cat:
@phalcon @andresgutierrez I fixed the build.
Should I squash to a single commit and create a new pull request or will this do?

andresgutierrez:
Thanks

Issue #9963, opened 2015-04-02
[2.0.0] Phalcon\Di\Exception: Service 'eventsManager' cannot be resolved for existing service


arius86:
Part of dump from my DI:
````
[
//...
 'eventsManager' =>
    class Phalcon\Di\Service#5924 (5) {
      protected $_name =>
      string(13) "eventsManager"
      protected $_definition =>
      string(21) "Phalcon\Events\Manage"
      protected $_shared =>
      bool(true)
      protected $_resolved =>
      bool(false)
      protected $_sharedInstance =>
      NULL
    }
//...
]
````

Error when calling:

````
$eventsManager = $this->di->getShared('eventsManager');
````

andresgutierrez:
Can you please try again?

arius86:
Still the same.

andresgutierrez:
I think we need more information to reproduce the issue, DI tests are passing in Travis and the forum and the website is running with Phalcon 2. Could you please share a repository with the code to reproduce the problem?

arius86:
Code: https://github.com/arius86/core/tree/v2.0
Travis CI log for current build: http://archive.travis-ci.org/jobs/57795446/log.txt (search for string
"Phalcon\Di\Exception: Service 'eventsManager' cannot be resolved")

olivier-monaco:
Can you create a small use case?

olivier-monaco:
I think I've found your problem. There's a missing r in constructor of Phalcon\DI\DefaultFactory\Cli for the event manager class name.

arius86:
Hah that was unexcpected :)

olivier-monaco:
Please check with latest 2.0.0.

PR #10039, opened 2015-04-14
Some mistake in Phalcon\DI\DefaultFactory\CLI

olivier-monaco:
See #9963.

I took code from Phalcon\DI\DefaultFactory to check all services defined in CLI class.


sergeyklay:
WoW :+1:

Green-Cat:
:+1:

andresgutierrez:
Thanks

Issue #2502, opened 2014-06-05
getFirst then iteration results in extra queries


mrjameshamilton:
Probably related to #2391 

If I call getFirst on a Resultset or use the ArrayAccess interface a query is performed, then when iterating through the resultset a further query is performed for each iteration.

Green-Cat:
@andresgutierrez Implemented in 2.0 (#10100)

PR #10050, opened 2015-04-16
[2.0.0] Refactored mvc/model/resultset to work with AppendIterator

patrick-zippenfenig:
The current resultset implementation does not work with appendIterator (added unit-test).

"valid()" must not move the cursor to the next position. Now: "next()" and "seek()" move the cursor and prepare the row for hydration. Hydration is handled in "current()" instead of "valid()".

It is now safe to call valid() and current() multiple times. current() keeps the activeRow in memory.

Ran unit-tests locally. In case travis build fails I correct accordingly


patrick-zippenfenig:
Should also fix https://github.com/phalcon/cphalcon/issues/2502

patrick-zippenfenig:
I'll add an additional update to improve forward seeks. Currently array access operations on resultsets > 25 rows trigger a query re-execution
```php
$robots = Robots::find();
for($i = 0; $i < count($robots); ++$i) {
  echo $robots[$i]->name;
}
```

patrick-zippenfenig:
rebase went wrong... reopen on new branch

PR #10058, opened 2015-04-17
Fix exception message in Di\Service

igorgolovanov:


andresgutierrez:

PR #10065, opened 2015-04-18
Add class Phalcon\Http\StatusCode

dreamsxin:


sergeyklay:
:+1:

ovr:
:+1:

Green-Cat:
:+1:

KorsaR-ZN:
Oh, you're faster than me, I was late :)
:+1:

SidRoberts:
Do the constants need to have the "SC_" prefix? And why is the static `_messages` property needed?

andresgutierrez:
We don't add changes directly to master: https://github.com/phalcon/cphalcon/blob/master/CONTRIBUTING.md#pull-request-checklist

sergeyklay:
@dreamsxin Can u please pull to 2.0.0 branch?

phalcon/http/statuscode.zep
```diff
@@ -0,0 +1,182 @@
...
+	 *
+	 * @param int statusCode
+	 * @return string
+	 */
+	public static function getStatusText(int statusCode)
+	{
+		if typeof self::_messages != "array" {
+			let self::_messages = [
+				// INFORMATIONAL CODES
+				100 : "Continue",
```
ogarbe: for better clarity, you should use the const you defined hundred lines upper :)

dreamsxin: If really necessary?

sergeyklay:
I agree with @SidRoberts

dreamsxin:
@SidRoberts @sergeyklay Because `CONTINUE` is reserved keyword, so add prefix.

KorsaR-ZN:
I think the prefix must be such:
- 1xx - INFO_
- 2xx - SUCCESS_
- 3xx - REDIRECT_
- 4xx - CLIENT_
- 5xx - ERROR_

PR #10068, opened 2015-04-18
Removes dead code in Di\Injectable

Green-Cat:


SidRoberts:
There's another instance of this in `\Phalcon\Mvc\View\Engine\Volt\Compiler::setDI()`. :wink:

andresgutierrez:

Green-Cat:
@SidRoberts Thanks, will include in my new commit.

PR #10073, opened 2015-04-18
Iteration and array access feature on session bag like on 1.3.4 version

DestinyMKas:
Missing functionality in 2.0.0 version on session bag:
iteration
arrayaccess
count


sergeyklay:
@ogarbe :+1: good catch!

andresgutierrez:
Thanks

PR #10074, opened 2015-04-18
Add late state binding on model

ogarbe:
Conflicts:
phalcon/mvc/model/resultset/simple.zep

cf #3107

### DIFF
--- a/config.json
+++ b/config.json
@@ -76,6 +76,10 @@
         "orm.enable_literals": {
             "type": "bool",
             "default": true
+        },
+        "orm.late_state_binding": {
+            "type": "bool",
+            "default": false
         }
     },
     "destructors": {
--- a/phalcon/mvc/collection.zep
+++ b/phalcon/mvc/collection.zep
@@ -417,7 +417,7 @@ abstract class Collection implements CollectionInterface, InjectionAwareInterfac
 			/**
 			 * Assign the values to the base object
 			 */
-			return self::cloneResult(base, document);
+			return static::cloneResult(base, document);
 		}
 
 		/**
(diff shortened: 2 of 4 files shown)


KorsaR-ZN:
:+1:

Green-Cat:
:+1:

andresgutierrez:
Could you please rebase this to latest 2.0.x?

Green-Cat:
@ogarbe Quick question, why is the setting even necessary? Are there any situations when someone would not want this behaviour? oO

andresgutierrez:
@Green-Cat I asked him to implement it that way, dynamic calls are slower compared to methods well known at compile time, as this implementation is key when hydrating records this allows to reduce the overhead if someone is not using that feature.

sergeyklay:
@Green-Cat `"default": true` is enough :)

andresgutierrez:
@ogarbe it should be `"default": false`

ogarbe:
@andresgutierrez  @sergeyklay it's done

andresgutierrez:
Sorry :'(, could you please rebase again?

ogarbe:
@andresgutierrez done :-)

Green-Cat:
:+1:

andresgutierrez:
Thanks

KorsaR-ZN:
@ogarbe Your PR broke build. Please fix it

vitaliy-w:
@ogarbe Is there any reason why `orm.late_state_binding` is defaulted to `false` in this PR, but was originally defaulted to `true` in https://github.com/phalcon/cphalcon/pull/3107?

ogarbe:
yes, it's a request from @andresgutierrez who though this functionality should be disabled by default because of the extra operations it request.

You can enable it globally in php.ini

``ìni
extension=phalcon.so
phalcon.orm.late_state_binding=1
```

vitaliy-w:
@ogarbe Thank you for explaining. This is a very useful feature to have.

PR #10075, opened 2015-04-18
[2.0.x] [NFR] Global container

KorsaR-ZN:
It will be convenient to reduce communication between classes.
See:
- #10005
- https://github.com/phalcon/cphalcon/blob/2.0.x/phalcon/forms/manager.zep#L44
- https://github.com/phalcon/cphalcon/blob/2.0.x/phalcon/db/adapter.zep#L132
- https://github.com/phalcon/cphalcon/blob/2.0.x/phalcon/queue/beanstalk.zep#L165
- etc...

**Examples**

```
   // Gets CLASS instance
   instance = Phalcon\Factory::di()->get(CLASS);

   // Gets Application instance
   instance = Phalcon\Factory::app();

   // Gets Dispatcher service
   instance = Phalcon\Factory::getDispatcher();

   // Gets ModelManager service
   instance = Phalcon\Factory::getModelsManager();
```

### DIFF
--- a/phalcon/factory.zep
+++ b/phalcon/factory.zep
@@ -20,21 +20,35 @@
 namespace Phalcon;
 
 use Phalcon\Di;
+use Phalcon\DiInterface;
+use Phalcon\Events\ManagerInterface as EventsManagerInterface;
+use Phalcon\Mvc\Application as ApplicationInterface;
 
 /**
- * Phalcon\Filter\Exception
+ * Class Phalcon\Factory
  *
- * Exceptions thrown in Phalcon\Filter will use this class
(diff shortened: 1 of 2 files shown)


andresgutierrez:
I'll leave this open for discussion, however I think it would be better not to use static methods as they introduce potential problems to test applications since there is no way to overrider the called method dependency.

sergeyklay:
I agree with @andresgutierrez

Atriedes:
I think this is bad idea, every dependency should be injected via constructor or setter method, it can make application more loose coupling.

PR #10078, opened 2015-04-18
[2.0.x] fix "call to undefined method" when using Phalcon\Mvc\Dispatcher::getControllerClass()

stecman:
The method `Phalcon\Mvc\Dispatcher::getControllerClass()` doesn't work as it references a method at runtime that doesn't exist:

> Fatal error: Uncaught exception 'RuntimeException' with message 'Call to undefined method Phalcon\Mvc\Dispatcher::gethandlername()'

``` php
public function getControllerClass() -> string
{
    return this->{"getHandlerName"}();
}
```

This PR fixes that error by refactoring the building of the handler class name from `Phalcon\Dispatcher::dispatch` into `Phalcon\Dispatcher::getControllerClass` method. Note that this adds an additional public method to `Phalcon\Dispatcher`, as the broken method was on `Phalcon\Mvc\Dispatcher`.

**Small issue introduced with this change:** The method `getControllerClass` depends on the resolving of defaults in `Dispatcher::dispatch()` to work correctly which isn't great, however fixing this would require reworking how defaults are set, which is outside the scope of this change. As-is, this code won't change the current behaviour of `Dispatcher::dispatch`, but some incorrect behaviour (defaults not used) is possible if calling `getControllerClass` outside of the dispatch loop.

### DIFF
--- a/phalcon/dispatcher.zep
+++ b/phalcon/dispatcher.zep
@@ -207,6 +207,39 @@ abstract class Dispatcher implements DispatcherInterface, InjectionAwareInterfac
 	}
 
 	/**
+	 * Gets the fully qualified controller class name that will be located to dispatch the request
+	 */
+	public function getControllerClass() -> string
+	{
+		var handlerSuffix, handlerName, namespaceName,
+			camelizedClass, handlerClass;
+
+		let handlerSuffix = this->_handlerSuffix,
+			handlerName = this->_handlerName,
+			namespaceName = this->_namespaceName;
+
+		// We don't camelize the classes if they are in namespaces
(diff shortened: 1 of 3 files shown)


andresgutierrez:
Could you please add unit tests?

stecman:
Tests added, including a failing test for the noted reliance on `Dispatcher::dispatch` having been run. Currently look at fixing this state issue in another PR (dependent on this one).

stecman:
Just went to rebase this against the 2.0.x branch and found this issue has already been fixed in a different PR: #10168.

PR #10083, opened 2015-04-19
[2.0.x] validation refactoring, double array initialization fix

arius86:
## Major changes
- added abstract `validate()` method to `Phalcon\Validation\Validator` class, made this class abstract
- `Phalcon\Validator` class now implements `Phalcon\Validation\ValidatorInterface`
- changed "keys" array initialization in `Phalcon\Cache\Backend\Memory`
## Minor changes
- removed `implements Phalcon\Validation\ValidatorInterface` from all `Phalcon\Validation\Validator\*` classes because parent class (`Phalcon\Validation\Validator`) is implementing it right now
- added "use" statements to  `Phalcon\Validation\Validator\*` instead using fully qualified names

### DIFF
--- a/phalcon/cache/backend/memory.zep
+++ b/phalcon/cache/backend/memory.zep
@@ -148,14 +148,14 @@ class Memory extends Backend implements BackendInterface
 	 */
 	public function queryKeys(var prefix = null) -> array
 	{
-		var data, index;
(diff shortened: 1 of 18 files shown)


andresgutierrez:
Thanks

PR #10088, opened 2015-04-19
[2.0.x] [FIX] PHP Warning - argument 2 not passed to function

KorsaR-ZN:


KorsaR-ZN:
Tests **passed** - [see](https://travis-ci.org/KorsaR-ZN/phalcon-core/builds/59170645)
An error in https://travis-ci.org/phalcon/cphalcon/jobs/59170796#L3035

pablomoreno61:
So, what's the version we should use in a production environment? Phalcon master or phalcon 2.0.x?

Green-Cat:
Master

SidRoberts:
Why not this?:

```
public function query(string! index, placeholders = null, domain = null) -> string
{
	var translation;

	if !domain {
		let translation = gettext(index);
	} else {
		let translation = dgettext(domain, index);
	}

	// ...
}
```

KorsaR-ZN:
@SidRoberts 
it restriction is due to the interface.
See https://github.com/phalcon/cphalcon/blob/master/phalcon/translate/adapterinterface.zep#L46

pablomoreno61:
Moreover, I can write:

<h1>{{ translate.t('Datos del usuario') }}</h1>

<h1>{{ translate.t('Datos del usuario', null) }}</h1>

And both lines gives me the same warning:
Warning: func_get_arg(): Argument 2 not passed to function in
/var/www/frontend/cache/volt_var_www_frontend_apps_modules_frontend_views_profile_index.phtml.php
on line *2*

2015-04-21 16:08 GMT+02:00 Stanislav Kiryukhin <notifications@github.com>:

KorsaR-ZN:
@pablomoreno61
upgrade from this branch and the error will go away.

Or apply patch https://github.com/phalcon/cphalcon/commit/39a399cc26b41439512591c0bd91efebde09263d.patch for master branch

pablomoreno61:
Hi, is this solution applied into the master branch?

andresgutierrez:
@pablomoreno61 not yet

PR #10090, opened 2015-04-20
[NFR] Added Phalcon\Mvc\View\Engine\Markdown

kjdev:
Re-submit. (#9983)


andresgutierrez:
Keeping this open for discussion

dedalozzo:
I'm using hoedown and it works. I use it to render just a part of a view (markdown articles and comments), but it can be useful to render static text or build a static site generator like Jekyll.

sergeyklay:
:+1:

sergeyklay:
I would like see this library in incubator. Thanks for contributing.

PR #10048, opened 2015-04-16
[2.0.0] Updated Phalcon\Session

kjdev:
When you use the "Phalcon\Session\Adapter\Files" session disappears in the destroy.
When not otherwise(Phalcon\Session\Adapter\Libmemcached) remains without disappearing session.

Example DI:

``` php
$di->setShared('session', function() {
    // Use files
    $session = new Phalcon\Session\Adapter\Files();

    /*
    // Use libmemcached
    $session = new Phalcon\Session\Adapter\Libmemcached([
        'servers' => [
            ['host' => 'localhost', 'port' => 11211, 'weight' => 1],
        ],
        'client' => [],
        'lifetime' => 3600,
    ]);
    */

    $session->start();
    return $session;
});
```

Example Controller Action:

``` php
public function testAction()
{
    // Session write
    $this->session->set("name", "Michael");

    // Session read
    if ($this->session->has("name")) {
        echo 'Name: ', $this->session->get("name");
    }

    // Session destroy
    $this->session->destroy();

    // Redirect
    return $this->response->redirect('dump');
}

public function dumpAction()
{
    var_dump($_SESSION);

    /*
    // Use files
    array (size=0)
      empty

    // Use libmemcached
    array (size=1)
      'name' => string 'Michael' (length=7)    
    */    
}
```

session_destroy has not been called in "Phalcon\Session\Adapter\ XXX" of class destory.

Proposal:

Enclosing the Adapter to Phalcon\Session class.

``` php
$di->setShared('session', function() {
    $session = new Phalcon\Session([
      // Use files (default)
      'adapter' => 'files',
      /*
      // Use libmemcached
      'adapter' => 'libmemcached',
      'options' => [
        'servers' => [
            ['host' => 'localhost', 'port' => 11211, 'weight' => 1],
        ],
        'client' => [],
        'lifetime' => 3600,
      ],
      */
    ]);
    $session->start();
    return $session;
});
```

PR #10091, opened 2015-04-20
[NFR] Updated Phalcon\Session

kjdev:
Re-submit. (#10048)


andresgutierrez:
Keeping this open for discussion

KorsaR-ZN:
:+1:

PR #10113, opened 2015-04-21
[2.0.x] Removed manual type checking.

SidRoberts:
In a lot of these cases, these Exceptions won't ever get thrown because the parameter types will require the variable to have that type anyway.


Green-Cat:
:+1:

andresgutierrez:
Thanks

PR #10119, opened 2015-04-21
Check array key exists before accessing. Support mysql double column types

masenocturnal:
Hello,
I ran into a few issues when porting a 1.3.x project to 2. I was hoping you could include the following changes. 

1) Check array key exists before accessing. 
2) Support mysql double column types

### DIFF
--- a/phalcon/db/column.zep
+++ b/phalcon/db/column.zep
@@ -103,6 +103,13 @@ class Column implements ColumnInterface
 	const TYPE_BOOLEAN = 8;
 
 	/**
+	 * Double abstract data type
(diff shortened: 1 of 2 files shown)


andresgutierrez:
Thanks

DusanBrejka:
Hi @andresgutierrez, thanks for fixing a bug with double column types.
Unfortunately the same problem still exists for Bit columns. Could you please fix this as well?

masenocturnal:
Yep will do. 
Also need to modify the introspection MetaData manager to deal with these.

PR #10129, opened 2015-04-22
Add class Phalcon\Http\StatusCode

dreamsxin:


KorsaR-ZN:
I think the prefix must be such:

1xx - INFO_
2xx - SUCCESS_
3xx - REDIRECT_
4xx - CLIENT_
5xx - SERVER_

@andresgutierrez What do you think about this?

Green-Cat:
@KorsaR-ZN I also think so.

sergeyklay:
@KorsaR-ZN +1

dreamsxin:
@KorsaR-ZN Have been using the prefix.

phalcon/http/statuscode.zep
```diff
@@ -0,0 +1,182 @@
...
+				504 : "Gateway Time-out",
+				505 : "HTTP Version not supported",
+				506 : "Variant Also Negotiates",
+				507 : "Insufficient Storage",
+				508 : "Loop Detected",
+				511 : "Network Authentication Required"
+			];
+		}
+
+		if (array_key_exists(statusCode, self::_messages)) {
```
sergeyklay: isset is faster than array_key_exists also because is a language construct, not a function

phalcon/http/statuscode.zep
```diff
@@ -0,0 +1,182 @@
...
+ */
+abstract class StatusCode
+{
+	// INFORMATIONAL CODES
+	const INFO_CONTINUE = 100;
+	const INFO_SWITCH_PROTOCOL = 101;
+	const INFO_PROCESSING = 102;
+
+	// SUCCESS CODES
+	const SUCCESS_CREATED = 201;
```
sergeyklay: If u created a class for status codes, can u please provide all codes, for example `200` too

phalcon/http/statuscode.zep
```diff
@@ -0,0 +1,182 @@
...
+	const SUCCESS_ACCEPTED = 202;
+	const SUCCESS_NON_AUTHORATIVE_INFORMATION = 203;
+	const SUCCESS_NO_CONTENT = 204;
+	const SUCCESS_RESET_CONTENT = 205;
+	const SUCCESS_PARTIAL_CONTENT = 206;
+	const SUCCESS_MULTI_STATUS = 207;
+	const SUCCESS_ALREADY_REPORTED = 208;
+
+	// REDIRECTION CODES
```
sergeyklay: 226 ? 
see RFC3229

KorsaR-ZN: @sergeyklay :+1: 
@dreamsxin Please add all HTTP-codes (see [wiki](http://en.wikipedia.org/wiki/List_of_HTTP_status_codes))

andresgutierrez:
-1 There are many constants to add, constants must be allocated in each request startup

PR #10131, opened 2015-04-22
[2.0.x] [BUG] fixed compatibility with 1.3.x

KorsaR-ZN:
This fixes bug, when the method returns NULL always:

```
class SomeModel extend Phalcon\Mvc\Model
{
    /**
     * @return string
     */
    public function getManufacturer()
    {
        if ($manufacturer = $this->manufacturer) {
            return $manufacturer->name;
        } else {
            return null;
        }
    }

    public function initialize()
    {
        $this->belongsTo('manufacturer_id', 'App\Models\Manufacturer', 'id', [
            'alias'    => 'manufacturer',
            'reusable' => true
        ]);
    }
}
```


KorsaR-ZN:
@akaNightmare Thanks!

PR #10145, opened 2015-04-23
[2.0.x] CLI docblock cleanup

SidRoberts:


phalcon/cli/router.zep
```diff
@@ -407,7 +391,7 @@ class Router implements \Phalcon\Di\InjectionAwareInterface
 	 * @param string/array paths
 	 * @return Phalcon\Cli\Router\Route
 	 */
-	public function add(string! pattern, paths = null) -> <Router>
+	public function add(string! pattern, paths = null) -> <Route>
```
sergeyklay: In this case
```php
 	 * @param string pattern
 	 * @param string/array paths
 	 * @return Phalcon\Cli\Router\Route
```
not needed

sergeyklay: And use `-> <\Route>`

SidRoberts: I don't want to get rid of `string|array` for the `paths` parameter. And there's a use statement mapping `Route` to `\Phalcon\Cli\Router\Route` - is this what you mean?

sergeyklay: Sry, about `<\Route>`
had not noticed that you are in the same namespace

PR #10146, opened 2015-04-23
[2.0.x] ACL docblock cleanup

SidRoberts:


phalcon/acl/adapter/memory.zep
```diff
@@ -612,20 +613,16 @@ class Memory extends Adapter
...
 		return this->_roles;
 	}
 
 	/**
 	 * Return an array with every resource registered in the list
-	 *
-	 * @return Phalcon\Acl\Resource[]
 	 */
-	public function getResources()
+	public function getResources() -> <$Resource[]>
```
sergeyklay: hmm.. $ ?

SidRoberts: `Resource` is a keyword in Zephir (as is `Inline`). See https://github.com/phalcon/cphalcon/blob/2.0.x/phalcon/assets/resource.zep#L31 and https://github.com/phalcon/cphalcon/blob/2.0.x/phalcon/assets/inline.zep#L31.

sergeyklay: Oh, sry :open_mouth:

PR #10147, opened 2015-04-23
[2.0.x] Fix for #10131

KorsaR-ZN:
see #10131

### DIFF
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -40,6 +40,7 @@ use Phalcon\Mvc\Model\MetadataInterface;
 use Phalcon\Mvc\Model\MessageInterface;
 use Phalcon\Events\ManagerInterface as EventsManagerInterface;
 use Phalcon\Db\Column;
+use Phalcon\Text;
 
 /**
  * Phalcon\Mvc\Model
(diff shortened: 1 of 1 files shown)


sergeyklay:
:+1: 

but now we have  +1 dependence

KorsaR-ZN:
@sergeyklay 
Maybe it is better?
```
...
let method = "get" . this->_camelize(property);
...
protected function _camelize(string! str) -> string
{
   return Text::camelize(str);
}
```

sergeyklay:
yes  it is better?

phalcon:
Why an extra method just to call a method?

KorsaR-ZN:
@phalcon That's better, code becomes more flexible

```
....

protected function getGettersProperty(string! str) -> string
{
    return "get" . Text::camelize(str);
}

....

let method = this->getGettersProperty(property);

....
```

andresgutierrez:
Why an extra method just to call a method???

KorsaR-ZN:
@andresgutierrez So anyone can change logic create getters for property

andresgutierrez:
How often do you change how getters look like? Most applications just have get + name field, what would be a different common case?

KorsaR-ZN:
@andresgutierrez I agree with you. But I think better to leave the last option. So the code is improved.
An extra method is not slowing code.

Or do you think I should do reset?

andresgutierrez:
@KorsaR-ZN Could you please remove that method?

KorsaR-ZN:
@andresgutierrez I have to do rebase?

andresgutierrez:
Thanks

PR #10153, opened 2015-04-23
Reenabling disabled unit tests

Green-Cat:
This commit:
- enables disabled tests in the asset filter
- fixes and reenables the version tests
- reenables an escaper test


sergeyklay:
excelent!

KorsaR-ZN:
Maybe should `git rebase`?

andresgutierrez:
Thanks

PR #10158, opened 2015-04-24
Updated Phalcon\Http\Response::redirect

dreamsxin:


phalcon/http/response.zep
```diff
@@ -450,7 +450,9 @@ class Response implements ResponseInterface, InjectionAwareInterface
 
 		if dependencyInjector->has("view") {
 			let view = <ViewInterface> dependencyInjector->getShared("view");
-			view->disable();
+			if view instanceof \Phalcon\Mvc\ViewInterface {
```
akaNightmare: may be move to ```use``` statement (\Phalcon\Mvc\ViewInterface) ?

Green-Cat: :+1:

andresgutierrez:
Thanks

PR #10142, opened 2015-04-22
Updated security.zep with fixes and new functionality

agim-a:
**checkToken** function makes sure that _tokenKey_ exists in session, if not provided

 Added a utility function **destroyToken** to be used when checkToken
returns true

Added a 3rd variable to _checkToken_
destroyIfValid, which removes token key and value from session

Moved token key and value session ids as protected properties

PR #10165, opened 2015-04-24
Updated security.zep with fixes and new functionality #10142

agim-a:
submitted #10142 to 2.0.x branch

checkToken function makes sure that tokenKey exists in session, if not provided

 Added a utility function destroyToken to be used when checkToken
returns true

Added a 3rd variable to checkToken
destroyIfValid, which removes token key and value from session

Moved token key and value session ids as protected properties

### DIFF
--- a/phalcon/security.zep
+++ b/phalcon/security.zep
@@ -50,6 +50,10 @@ class Security implements InjectionAwareInterface
 
 	protected _numberBytes = 16;
 
+	protected _tokenKeySessionID = "$PHALCON/CSRF/KEY$";
+
+	protected _tokenValueSessionID = "$PHALCON/CSRF$";
+
 	protected _csrf;
 
 	/**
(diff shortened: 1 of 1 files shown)


agim-alb:
$2a$ means that it is not Legacy, right? Because it is current
So isLegacyHash will return *true* for hashes like md5,sha1 etc. and *false* if crypt hash that starts with $2a$
``` php
isLegacyHash(md5('password')); should return TRUE
```
while

``` php
isLegacyHash('$2a$.....')); should return FALSE
```

Thoughts?

andresgutierrez:
No, it should returns true if the hash starts by $2a$

This is how it's implemented in 1.3.5 https://github.com/phalcon/cphalcon/blob/1.3.5/ext/security.c#L670-L672

agim-alb:
Ok, updated it , so that isLegacyHash returns true, if hash starts with $2a$

andresgutierrez:
Thanks

PR #10168, opened 2015-04-25
Add method `Phalcon\Dispatcher::getHandlerClass`

dreamsxin:
Add method `Phalcon\Dispatcher::getHandlerClass`
Updated method `Phalcon\Mvc\Dispatcher::getControllerClass`
Updated method `Phalcon\Mvc\Application::handle`

If controler action return false, will don't render view.


dreamsxin:
@andresgutierrez It's ok.

andresgutierrez:
Thanks

PR #10141, opened 2015-04-22
Fixed checking case insensitive image extensions

agim-a:
Replaced **strcmp** with **strcasecmp** for checking image extenstions

PR #10173, opened 2015-04-25
[2.0.x] Fixed checking case insensitive image extensions

agim-a:
Also fixed a bug using _imagepng_ instead of _imagejpeg_ to create png images

From pull request #10141


andresgutierrez:
Thanks

PR #10183, opened 2015-04-27
More double support

masenocturnal:
Hello,
My previous work didn't go far enough to supporting double column types in MySQL. This appears to work with the introspection strategy and other parts of Phalcon.


andresgutierrez:
We don't add changes directly to master, could you please submit this to the 2.0.x branch?

https://github.com/phalcon/cphalcon/blob/2.0.x/CONTRIBUTING.md#pull-request-checklist

PR #10186, opened 2015-04-27
[2.0.x] Tidied up Image component (docblocks, comments, use statements).

SidRoberts:


phalcon/image/adapter/gd.zep
```diff
@@ -19,11 +19,14 @@
 
 namespace Phalcon\Image\Adapter;
 
+use Phalcon\Image\Adapter;
+use Phalcon\Image\Exception;
+
 class Gd extends \Phalcon\Image\Adapter implements \Phalcon\Image\AdapterInterface
```
Green-Cat: You don't need to use the namespace here since it's already in the use statement above.
You should probably also move the Interface namespace to a use statement.

SidRoberts: Whoops, I missed that. :smile:

PR #10188, opened 2015-04-27
2.0.x More double support

masenocturnal:
Hi resubmitting the patches to the correct branch. 

I have some unit tests for these prepared for this patch but I am having trouble getting the unit tests to actually execute.

I believe I have following the instructions correctly from:
https://github.com/phalcon/cphalcon/blob/master/unit-tests/README.md

What is the best way to get assistance with this ? 
am@portland:~/cphalcon/unit-tests$ phpunit --debug -c phpunit.xml 
PHPUnit 4.2.6 by Sebastian Bergmann.

Configuration read from /home/am/cphalcon/unit-tests/phpunit.xml

Time: 23 ms, Memory: 1.75Mb

No tests executed!
am@portland:~/cphalcon/unit-tests$

### DIFF
--- a/phalcon/db/adapter/pdo/mysql.zep
+++ b/phalcon/db/adapter/pdo/mysql.zep
@@ -195,6 +195,16 @@ class Mysql extends PdoAdapter implements AdapterInterface
 				}
 
 				/**
+				 * Doubles are float
+				 */
+				if memstr(columnType, "double") {
+					let definition["type"] = Column::TYPE_DOUBLE,
+						definition["isNumeric"] = true,
+						definition["bindType"] = Column::TYPE_DECIMAL;
+					break;
+				}
+				
+				/**
 				 * By default is string
 				 */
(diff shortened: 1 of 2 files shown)


SidRoberts:
Have a look at [run-tests.sh](https://github.com/phalcon/cphalcon/blob/2.0.x/run-tests.sh).

It's a very subtle difference, but to run PHPUnit, you need to run this command from the main directory (`cphalcon`):

```bash
php unit-tests/ci/phpunit.php --debug -c unit-tests/phpunit.xml --testsuite=stable
```

It also includes the bootstrap file ([unit-tests/ci/phpunit.php](https://github.com/phalcon/cphalcon/blob/2.0.x/unit-tests/ci/phpunit.php)).

Just running `sh run-tests.sh` should be OK though.

andresgutierrez:
Could you please add tests and update the CHANGELOG?

PR #10193, opened 2015-04-28
Fixes #1482

### LINKED ISSUE #1482: PhalconException: Column type does not support scale parameter
when i try to find data from a model .
how can i set the model's column 's scale.

the table include int ,double type.

```
    CREATE TABLE `carddatalist` (
  `dataseq` int(11) NOT NULL,
  `data_type` varchar(255) DEFAULT NULL,
  `brclvl` varchar(1) DEFAULT NULL,
  `brc` varchar(8) DEFAULT NULL,
  `desc1` varchar(16) DEFAULT NULL,
  `desc2` varchar(32) DEFAULT NULL,
  `numval` int(11) DEFAULT NULL,
  `amtval` double(21,2) DEFAULT NULL,
  `datatype2` varchar(2) DEFAULT NULL,
  KEY `datalist_idx1` (`dataseq`,`brclvl`)
) ENGINE=MyISAM DEFAULT CHARSET=utf8;
```

my model is simple:

```
class  Carddatalist extends \Phalcon\Mvc\Model{
      public function initialize()
    {
         $this->setConnectionService('mysql');
    }

    public function getSource(){
            return "carddatalist";
    }
```

}

i just `Carddatalist:find();` in `IndexController`

PhalconException: Column type does not support scale parameter

i don's know how can i define the scale of column.

ghost:
Added double column type to mysyl.zep to prevent PhalconException: Column type does not support scale parameter

### DIFF
--- a/phalcon/db/adapter/pdo/mysql.zep
+++ b/phalcon/db/adapter/pdo/mysql.zep
@@ -195,6 +195,16 @@ class Mysql extends PdoAdapter implements AdapterInterface
 				}
 
 				/**
+		                 * Float/Smallfloats/Decimals are float
+		                 */
+		                if memstr(columnType, "double") {
+		                    let definition["type"] = Column::TYPE_DOUBLE,
+		                        definition["isNumeric"] = true,
+		                        definition["bindType"] = Column::TYPE_DOUBLE;
+		                    break;
+		                }
+
+				/**
 				 * By default is string
 				 */
 				let definition["type"] = Column::TYPE_VARCHAR;


andresgutierrez:
I think this is already submitted in #10188

andresgutierrez:
Also we don't add changes directly to master: https://github.com/phalcon/cphalcon/blob/2.0.x/CONTRIBUTING.md#pull-request-checklist

spuy767:
Sorry, I had accidentally navigated away from when I was on the 2.0.0 branch.

PR #10194, opened 2015-04-28
[2.0.x] Throw exceptions first

SidRoberts:
In other words, the methods return the typical use case at the very end of the method and all exceptions are caught before.

For example, instead of this:

``` zephir
public function stop() -> void
{
    if this->_cancelable {
        let this->_stopped = true;
    } else {
        throw new Exception("Trying to cancel a non-cancelable event");
    }
}
```

... do this:

``` zephir
public function stop() -> void
{
    if !this->_cancelable {
        throw new Exception("Trying to cancel a non-cancelable event");
    }

    let this->_stopped = true;
 }

```

### DIFF
--- a/phalcon/events/event.zep
+++ b/phalcon/events/event.zep
@@ -89,11 +89,11 @@ class Event
 	 */
 	public function stop() -> void
 	{
-		if this->_cancelable {
-			let this->_stopped = true;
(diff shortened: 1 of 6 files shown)


Green-Cat:
:+1:

PR #10198, opened 2015-04-29
Updated `is not` syntax for volt

dreamsxin:


andresgutierrez:
Thanks!

KorsaR-ZN:
@dreamsxin Please, fix tests, see https://travis-ci.org/phalcon/cphalcon/jobs/60547625#L2993

Issue #10225, opened 2015-05-03
[2.0.0] Phalcon\Config bug with array access with integer key: "Parameter 'index' must be a string"


dschissler:
I'm receiving **Parameter 'index' must be a string** exception when attempting to do array access with an integer.  It works with property access though.

Data:

```php
array (
    'signin' => 
        array (
            0 => 'signin',
            1 => 'org::session::signin'
        )
)
```

This doesn't work:

```php
$config->['signin'][1];
```

But this does work:

```php
$config['signin']->{1}
```

dschissler:
I'll reopen later if the new code doesn't work.

olivierandriessen:
This doesn't work in Phalcon 2.0

andresgutierrez:
@olivierandriessen It works for me in >= 2.0.2

olivierandriessen:
@andresgutierrez  Never mind, i'm still on 2.0.0.

PR #10226, opened 2015-05-03
[2.0.x] Config now accepts non-string keys (#10225)

SidRoberts:
See #10225 

The constructor has also been refactored using `offsetSet()`.


sergeyklay:
:+1:

PR #10227, opened 2015-05-04
[2.0.x] assets manager fixes

dubhunter:
This fixes the call to `$resource->getRealSourcePath()` by passing the source path created from the Manager and Collection.

This also cleans up an unnecessary full namespace reference to Collection.


andresgutierrez:
Thanks

dubhunter:
Glad to contribute!
Any ideas when this might be released? I never got it to compile locally :cry:

dubhunter:
Never mind on the release, I just realized that happened 4 days ago. Awesome!

PR #10241, opened 2015-05-05
[2.0.x]Fix mysql pdo double type

xyhp915:


andresgutierrez:
Could you please add tests and update the CHANGELOG?

andresgutierrez:
@xyhp915 ping

PR #10246, opened 2015-05-05
[2.0.x] Fix for #10214

### LINKED ISSUE #10214: [2.0.x] \Phalcon\Debug\Dump::var
`\Phalcon\Debug\Dump::var`
https://github.com/phalcon/cphalcon/blob/2.0.x/phalcon/debug/dump.zep#L252

`var` is reserved word in PHP 
http://php.net/manual/en/reserved.keywords.php

Suggest rename it to `dump`

KorsaR-ZN:
Fix for #10214


Green-Cat:
:+1:

andresgutierrez:
Thanks

PR #10263, opened 2015-05-06
[2.0.x] Added more parameter/return types to Http

SidRoberts:


phalcon/http/request.zep
```diff
@@ -835,7 +835,6 @@ class Request implements RequestInterface, InjectionAwareInterface
 
 	/**
 	 * Gets a charsets array and their quality accepted by the browser/client from _SERVER["HTTP_ACCEPT_CHARSET"]
-	 *
 	 */
 	public function getClientCharsets()
```
sergeyklay: @SidRoberts can you add here `-> var` ?

PR #10271, opened 2015-05-06
Phalcon changed my mysql database host address and can't connect to db.

lieshanhe:
PDOException: SQLSTATE[28000] [1045] Access denied for user 'username'@'x.x.x.206' (using password: YES) 

but I set host is x.x.x.117.

so why phalcon said the host is x.x.x.206?

and you can see the screen shot here:

http://www.awesomescreenshot.com/image/189837/39318c098c8da4ffeca70615b4993a32

so, why I can't use $db = new \Phalcon\Db\Adapter\Pdo\Mysql() to connect my db?


andresgutierrez:
If you need help using Phalcon, you can better get help in the forum: http://forum.phalconphp.com/

zikezhang:
Actually, I connect the db server in version 1.3 in my local computer, the problem is after transferring my file to AMS , I can't connect to mysql, and also, why phalcon alert me that Access denied for user 'username'@'x.x.x.206'  but the db address is x.x.x.117.?

PR #10274, opened 2015-05-07
Fix #1608 about Phalcon\Mvc\Router\Annotations

baryshev:
Same as #2332.

Fix for #1608 in 2.0.x

### DIFF
--- a/phalcon/mvc/router/annotations.zep
+++ b/phalcon/mvc/router/annotations.zep
@@ -310,7 +310,11 @@ class Annotations extends Router
 				if value != "/" {
 					let uri = routePrefix . value;
 				} else {
-					let uri = routePrefix;
+					if typeof routePrefix !== "null" {
+						let uri = routePrefix;
+					} else {
+						let uri = value;
+					}
 				}
 			} else {
 				let uri = routePrefix . actionName;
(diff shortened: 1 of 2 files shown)


vladimmi:
```
/**
* Create the route using the prefix
*/
if typeof value !== "null" {
    if value != "/" {
        let uri = routePrefix . value;
    } else {
        let uri = routePrefix;
        if typeof routePrefix !== "null" {
            let uri = routePrefix;
        } else {
            let uri = value;
        }
    }
} else {
    let uri = routePrefix . actionName;
}
```

Why cant we just do something like `routePrefix . value` without additional checks and leave details (like double slashes, etc) to developers? So empty prefix or empty route just had no effect, appended as empty string and `/` path can be defined as:
1) empty prefix and `/` route
or 2) `/` prefix and empty route

`let uri = routePrefix . actionName;` - does this mean that `@Route()` is equal to `@Route('[actionName]')`? Have not seen any reference to this in docs. Is this behavior needed and more intuitive than consider empty route as `@Route('')`?

baryshev:
I agree about additional checks. But now we have broken compatibility with 1.3.x.

Constructions like `@Route('')` or `@RoutPrefix('')` looks as hacks.

vladimmi:
Maybe I was unclear about empty routes... I mean why `@Route()` is equal to `@Route('[action name]')` and not to `@Route('')`? If I leave route empty - that's because I want it to be empty, not equal to action name. And why this behavior is not covered by docs?

That's a question for original developers, however, and I'll post it as an issue, I think.

andresgutierrez:
Thanks

PR #10281, opened 2015-05-07
[2.0.x] Flash\Session::getMessages('key') now returns an empty array if there are no 'key' messages (fixes #10279)

### LINKED ISSUE #10279: Flash\Session::getMessage('key') returns an array
I set a flash success messages in the backend. In the view, i also check with getMessage('error') if there error messages present. the method "getMessages('error')" return an array like this:
Array ( [success] => Array (  [0] => "success Message") ).
There was already a similar Bug #908
I would expect an empty array, null or false

SidRoberts:
Fixes #10279. This is also effectively a _resubmission_ of #920.


andresgutierrez:
Thanks

Issue #10285, opened 2015-05-07
BUG - Volt form fields email_field(...) missing Array in sintax parse


wlisesrivas:
I have an application in production with Phalcon v1.3.4, after update to Phalcon v2.0 I have error with my form.

Volt code:
![email-field](https://cloud.githubusercontent.com/assets/1482092/7526247/0a9eaf5c-f4de-11e4-9819-9dc39f6614f2.PNG)

Result PHP Code (after parse, in Cache folder):
![volt-result](https://cloud.githubusercontent.com/assets/1482092/7526284/487d8758-f4de-11e4-8890-15e84b1909c6.PNG)

As you can see in the above code, when the element is "email" the result is missing the "Array" in the syntax.

andresgutierrez:
This is fixed in the 2.0.x branch

PR #10287, opened 2015-05-07
[2.0.x] The parameters of email fields in Volt are now wrapped in an array (#10285)

SidRoberts:
See #10285


andresgutierrez:
Thanks

Issue #10283, opened 2015-05-07
Fatal error when calling Request::hasPut()


mahelious:
After upgrading our environment to Phalcon 2.0 was running some tests and ran into an error when calling Phalcon\Http\Request::hasPut().  Checking the [documentation](http://docs.phalconphp.com/en/latest/api/Phalcon_Http_Request.html) for Phalcon 2.0 I find references to has, hasGet, hasPost and hasServer, however hasPut is missing.  Why has this method not been included?  If oversight, is there a php fragment I can use until a patched stable is released?  If it has been done away with intentionally why was that not mentioned in the upgrade guide?

andresgutierrez:
Fixed in 2.0.x branch

PR #10289, opened 2015-05-07
[2.0.x] Readded Http\Request::hasPut() (#10283)

SidRoberts:
See #10283 and https://github.com/phalcon/cphalcon/blob/1.3.4/ext/http/request.c#L692

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,7 @@
  - Fixed segfault when a docblock does not have annotations #10301
  - Fixed wrong number of parameters passed when triggering an event in Mvc\Collection
  - Now Mvc\Model checks if an attribute has a default value associated in the database and ignores it from the insert/update generated SQL
+ - Readded Http\Request::hasPut() (#10283)
 
 2.0.1
  - Added missing Phalcon\Debug::listenLowSeverity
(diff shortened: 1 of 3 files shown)


Green-Cat:
:+1:

sergeyklay:
@SidRoberts Can you please provide small test like https://github.com/phalcon/cphalcon/blob/master/tests/unit/Phalcon/Http/Request/HttpRequestTest.php#L401 ?

SidRoberts:
@sergeyklay I've tried to write tests but as PUT data is taken from `php://input`, I can't think of a feasible way to test it. On the plus side, attempting to write tests made me realise `$_PUT` doesn't exist. :stuck_out_tongue:

sergeyklay:
Another plus of testing

andresgutierrez:
The contents of file_get_contents("php://input") must be cached in a property, because if you call it twice the second will return null

sergeyklay:
Why not use `define('PHP_INPUT', 'php://input')` or something like that at start (bootstrap) time? :)

andresgutierrez:
Could you please update the CHANGELOG?

SidRoberts:
@andresgutierrez Done.

andresgutierrez:
Thanks

sergeyklay:
@SidRoberts 

>  I can't think of a feasible way to test it. 

```php
stream_wrapper_unregister("php");
stream_wrapper_register("php", "your_awesome_stream_mock_class");
// ...
// tests
// ...
stream_wrapper_restore("php");
```

PR #10290, opened 2015-05-07
Fills-in empty docblock comment

Green-Cat:


sergeyklay:
It seems `[ci skip]` works only fro commits, but not pull request

PR #10295, opened 2015-05-08
Update complex.zep and simple.zip

tmihalik:


phalcon/mvc/model/resultset/complex.zep
```diff
@@ -72,7 +72,7 @@ class Complex extends Resultset implements ResultsetInterface
 		 * If the database result is an object, change it to fetch assoc
 		 */
 		if typeof result == "object" {
-			result->setFetchMode(1);
+			result->setFetchMode(\Phalcon\Db::FETCH_ASSOC);
```
Green-Cat: Please add with a use statement instead.

andresgutierrez:
Thanks

Issue #10296, opened 2015-05-08
Phalcon\Mvc\Micro\Collection missing via() method


mahelious:
[doucmentation](http://docs.phalconphp.com/en/latest/reference/micro.html#defining-routes) for Micro-controllers indicates the following to map multiple methods at once:

> //Matches if the HTTP method is GET or POST
> $app->map('/repos/store/refs',"action_product")->via(array('GET', 'POST'));

However via() is absent from [Phalcon\Mvc\Micro\Collection](http://docs.phalconphp.com/en/latest/api/Phalcon_Mvc_Micro_Collection.html) and throws an error when attempting to invoke.

sergeyklay:
Can you please create issue in [docs](https://github.com/phalcon/docs) repo?

mahelious:
I don't believe the issue is with the docs, the Micro Collection ought to have a via() method.

Green-Cat:
It is not, I will be able to fix this when I get home, if no one makes a PR by then.

Green-Cat:
The via() method however, does not even exist in 1.3.5, so not too sure about that one.

mahelious:
Skimming the code I am mystified at how via() was ever working.  The only clue I have is that in 1.3 map() seemed to return a Phalcon\Mvc\Micro\CollectionInterface while in 2.0 it returns a Phalcon\Mvc\Micro\Collection.  It's not much help as there is no via function in ext/mvc/micro/collectioninterface.c (nor in ext/mvc/micro/collection.c for that matter).  I can confirm the code worked in Phalcon 1.3.3.

In any case, map() adds an item to the collection _handlers with an empty method.  Is there a "correct" way to populate the method (or preferably multiple methods) once map() is finished?  My thought is immediately after calling map() to pop the last element, and then foreach over the list of methods to update _handlers with multiple "cloned" items, but I don't know if that matches the expected behavior.

Green-Cat:
It implements the interface, so returning itself does return the interface.

andresgutierrez:
This is fixed in 2.0.x

PR #10298, opened 2015-05-08
[2.0.x] Mvc\Micro\Collection::map() now returns this

SidRoberts:
Previously returned `this` in [1.3.4](https://github.com/phalcon/cphalcon/blob/1.3.4/ext/mvc/micro/collection.c#L258) as do [`get()`](https://github.com/phalcon/cphalcon/blob/2.0.x/phalcon/mvc/micro/collection.zep#L157), [`post()`](https://github.com/phalcon/cphalcon/blob/2.0.x/phalcon/mvc/micro/collection.zep#L171), _et cetera_.

See also https://github.com/phalcon/cphalcon/issues/10296#issuecomment-100293384

### DIFF
--- a/phalcon/mvc/micro/collection.zep
+++ b/phalcon/mvc/micro/collection.zep
@@ -140,7 +140,8 @@ class Collection implements CollectionInterface
 	 */
 	public function map(string! routePattern, var handler, var name = null) -> <Collection>
 	{
-		return this->_addMap(null, routePattern, handler, name);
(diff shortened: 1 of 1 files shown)


Green-Cat:
:+1:

sergeyklay:
Or just 
```zephir
protected function _addMap(string! method, var routePattern, var handler, var name)
{
    ...
    return this;
}
```

andresgutierrez:
Thanks

PR #10299, opened 2015-05-08
[2.0.x] Fix micro collection

Green-Cat:
The return is meant to be an instance of Collection, however since
_addMap does not return anything it was NULL instead.


Green-Cat:
Oh well, I was beaten to it.
Duplicate of #10298.

PR #10303, opened 2015-05-09
Added missing @Delete annotation to full REST support.

baryshev:
Also replace tree of conditions with one hash lookup.

### DIFF
--- a/phalcon/mvc/router/annotations.zep
+++ b/phalcon/mvc/router/annotations.zep
@@ -242,7 +242,7 @@ class Annotations extends Router
 		<Annotation> annotation)
 	{
 		var isRoute, name, actionName, routePrefix, paths, value, uri,
-			route, methods, converts, param, convert, conversorParam, routeName, routerAnnotations;
+			route, methods, converts, param, convert, conversorParam, routeName;
 
 		let isRoute = false, methods = null;
 
@@ -251,11 +251,13 @@ class Annotations extends Router
 		/**
 		 * Find if the route is for adding routes
 		 */
-		let routerAnnotations = ["Route": null, "Get": "GET", "Post": "POST", "Put": "PUT", "Options": "OPTIONS", "Delete": "DELETE"];
-
-		if isset routerAnnotations[name] {
-			let isRoute = true;
-			let methods = routerAnnotations[name];
+		switch name {
+			case "Route": let isRoute = true; break;
+			case "Get": let isRoute = true, methods = "GET"; break;
+			case "Post": let isRoute = true, methods = "POST"; break;
+			case "Put": let isRoute = true, methods = "PUT"; break;
+			case "Delete": let isRoute = true, methods = "DELETE"; break;
+			case "Options": let isRoute = true, methods = "OPTIONS"; break;
 		}
 
 		if isRoute === true {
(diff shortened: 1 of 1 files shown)


andresgutierrez:
Using an array is slower than the set of if

baryshev:
Ok, i replaced array lookup with switch statement. I created a benchmark and got this results on 100000 iterations:

```
if: 0.11329197883606
array: 0.27652096748352
switch: 0.10886406898499
```

`benchmark.zep`

```php
namespace Test;

class Benchmark
{
	public function testConditions(name)
	{
		var isRoute, methods;
		let isRoute = false, methods = null;
		if name == "Route" {
			let isRoute = true;
		} else {
			if name == "Get" {
				let isRoute = true, methods = "GET";
			} else {
				if name == "Post" {
					let isRoute = true, methods = "POST";
				} else {
					if name == "Put" {
						let isRoute = true, methods = "PUT";
					} else {
						if name == "Options" {
							let isRoute = true, methods = "OPTIONS";
						}
					}
				}
			}
		}
		return [isRoute, methods];
	}

	public function testArray(name)
	{
		var isRoute, methods, routerAnnotations;
		let isRoute = false, methods = null;
		let routerAnnotations = ["Route": null, "Get": "GET", "Post": "POST", "Put": "PUT", "Options": "OPTIONS", "Delete": "DELETE"];
		if isset routerAnnotations[name] {
			let methods = routerAnnotations[name];
		}
		return [isRoute, methods];
	}

	public function testSwitch(name)
	{
		var isRoute, methods;
		let isRoute = false, methods = null;

		switch name {
			case "Route": let isRoute = true; break;
			case "Get": let isRoute = true, methods = "GET"; break;
			case "Post": let isRoute = true, methods = "POST"; break;
			case "Put": let isRoute = true, methods = "PUT"; break;
			case "Delete": let isRoute = true, methods = "DELETE"; break;
			case "Options": let isRoute = true, methods = "OPTIONS"; break;
		}
		return [isRoute, methods];
	}
}
```

`test.php`

```php
<?php

$t = new Test\Benchmark();

$count = 100000;

$time1Start = microtime(true);
for ($i = 0; $i < $count; $i++) {
	$t->testConditions('Get');
	$t->testConditions('Post');
	$t->testConditions('Put');
	$t->testConditions('Options');
}
$time1End = microtime(true);

$time2Start = microtime(true);
for ($i = 0; $i < $count; $i++) {
	$t->testArray('Get');
	$t->testArray('Post');
	$t->testArray('Put');
	$t->testArray('Options');
}
$time2End = microtime(true);

$time3Start = microtime(true);
for ($i = 0; $i < $count; $i++) {
	$t->testSwitch('Get');
	$t->testSwitch('Post');
	$t->testSwitch('Put');
	$t->testSwitch('Options');
}
$time3End = microtime(true);

echo 'if: ' . ($time1End - $time1Start) . PHP_EOL;
echo 'array: ' . ($time2End - $time2Start) . PHP_EOL;
echo 'switch: ' . ($time3End - $time3Start) . PHP_EOL;
```

sergeyklay:
:+1:

Green-Cat:
:+1:

andresgutierrez:
Thanks

PR #10305, opened 2015-05-10
[2.0.x] Added `stats()` methods to Beanstalk

SidRoberts:
Also features `Queue\Beanstalk::readYaml()` as a way to read YAML data from a Beanstalk server.


andresgutierrez:
Thanks

PR #10310, opened 2015-05-11
Fix #10266 update class `Phalcon\Db` constant

### LINKED ISSUE #10266: Undefined class constant 'FETCH_KEY_PAIR'
Hi!
Working code started return - Fatal error : Undefined class constant 'FETCH_KEY_PAIR'.
I cannot figure out, why its happens, before it works ok.
If i choose some other constant, like FETCH_ASSOC or FETCH_OBJ - it works, but \Phalcon\Db::FETCH_KEY_PAIR does not work, why it can be, maybe i need any updates ?

Phalcon version 2.0.0

dreamsxin:


andresgutierrez:
We don't add changes directly to master, could you please submit this to the 2.0.x branch?

PR #10315, opened 2015-05-11
Fix "The argument is not initialized or iterable()" in Headers::send()

endeveit:


SidRoberts:
Because this changes the behaviour of `send()` (if `typeof this->_headers != "array"`, `send()` will never return true, regardless of `headers_sent()`), I'd like to propose changing this instead:

NOW:
```zephir
protected _headers;
```

TO:
```zephir
protected _headers = [];
```

endeveit:
@SidRoberts done

andresgutierrez:
Could you please remove this commit from your PR? https://github.com/endeveit/cphalcon/commit/fea27421bb77f3204a0908bf09f621217bf3d1db

endeveit:
@andresgutierrez done

andresgutierrez:
Thanks

Issue #10273, opened 2015-05-06
Implement JsonSerializable in Phalcon\Mvc\Model, Phalcon\Mvc\Model\Resultset and Phalcon\Mvc\Model\Row


stibiumz:
Once dropped PHP < 5.4 support, http://php.net/jsonserializable

patrick-zippenfenig:
+1 usefull for APIs. I'd be happy to contribute

patrick-zippenfenig:
Started implementing right now :)

stibiumz:
@patrick-zippenfenig is then Phalcon 2.0 built for PHP >= 5.4? I think that it's supporting 5.3...

patrick-zippenfenig:
Hm.. I always tend to forget version differences. I could just implement jsonSerialize without actually implementing the interface and hope php >5.3 just looks for jsonSerialize. Otherwise we could define the interface for php 5.3 (http://stackoverflow.com/questions/9683687/json-encode-is-not-working-with-array-of-objects/27809271#27809271) I'll give it a try

patrick-zippenfenig:
PHP >5.3 does check for JsonSerializable interface and have to call jsonSerialize() manually: https://github.com/patrick-zippenfenig/cphalcon/commit/fbe06ad493f00b329d40c20e66d44a9a0ecee6e3
````
$robots = Robots::find();
echo json_encode( $robots->jsonSerialize() );
````
@andresgutierrez would this behaviour comply with phalcon guidelines?

**Comment by andresgutierrez** [comment]
I think the following would be better:

```php
$robots = Robots::find();
echo json_encode( $robots->toJson() );
```

sergeyklay:
I agree with @andresgutierrez `toJson` is more readable than `jsonSerialize`. Also I thing that `toJson` is meets common CS like `toArray`.

virgofx:
:+1: toJson()

KorsaR-ZN:
@andresgutierrez, @sergeyklay, @virgofx
This requirement interface,
please see http://php.net/jsonserializable

virgofx:
Ah well we could always alias toJson().

patrick-zippenfenig:
PHP >5.4 calls ````jsonSerialize()```` automatically for objects implementing ````JsonSerializable```` interface:
````php
class Robots extends Mvc\Model {
  public function jsonSerialize() {
    return array('name' => $this->name);
  }
}
$robots = Robots::find();
echo json_encode( $robots );
````
[...]

PR #10320, opened 2015-05-11
Implement jsonSerialize() in mvc\model+row+result-set

patrick-zippenfenig:
Implements `jsonSerialize()` in mvc\model+row+result-set, BUT does not implement `JsonSerializable` interface as not supported by php 5.3 (otherwise `jsonSerialize()` would be called by `json_encode()` automatically).

See  #10273 

Syntax:

``` php
$robots = Robots::find();
echo json_encode( $robots->jsonSerialize() );
```

### DIFF
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -4204,6 +4204,20 @@ abstract class Model implements ModelInterface, ResultInterface, InjectionAwareI
 		return data;
 	}
 
+    /**
+    * Serializes the object for json_encode
+    *
+	*<code>
+	* echo json_encode($robot->jsonSerialize());
+	*</code>
+    *
+    * @return array
+    */
+	public function jsonSerialize() -> array
+	{
+	    return this->toArray();
+	}
+
 	/**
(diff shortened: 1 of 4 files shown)


Green-Cat:
Can't we use zephir prototypes to allow the interface even on 5.3? (although of cause it won't call the method automatically).

patrick-zippenfenig:
@andresgutierrez Is it possible to define ````JsonSerializable```` interface in zephir if not defined / php version 5.3?

andresgutierrez:
@patrick-zippenfenig Currently not, as the class will be binary linked to a reference to an interface that does not exist and that will make the extension not to load. I suggest wait for 2.1.x and deprecate 5.3

Green-Cat:
@andresgutierrez Major :+1: to dropping support for 5.3.

andresgutierrez:
@patrick-zippenfenig Could you please submit this to the 2.1.x branch?

patrick-zippenfenig:
@andresgutierrez Sure. I'll create a new PR this evening. Closing this one so far

PR #10327, opened 2015-05-12
[2.0.x] Multiple logger (context + critical)

tmihalik:


andresgutierrez:
This will break compatibility with 1.3.x https://github.com/phalcon/cphalcon/blob/1.3.5/ext/logger/multiple.c#L179

tmihalik:
@andresgutierrez no, It will work:
https://github.com/phalcon/cphalcon/blob/2.0.x/phalcon/logger/adapter.zep#L229

andresgutierrez:
Thanks

Issue #10328, opened 2015-05-12
[ORM] Phalcon\Mvc\Model - getSource() blocks generic use of prefix


marianacapelo:
Looking for a workaround to table prefixes that didnt exactly use the name class, I thought that it would be really simple to control it if not for the getSource method. Maybe I'm not thinking straight, but still, here is my post in the forum

http://forum.phalconphp.com/discussion/6711/table-prefix-without-getsource-method

sergeyklay:
Implemented in the `3.2.x` branch.

PR #10329, opened 2015-05-12
Model tables can now have prefixes (#10328)

SidRoberts:
This does introduce a problem: to override a model's source developers can only use:

``` php
public function initialize()
{
    $this->setSource('new_source');
}
```

I personally don't like overriding the `getSource()` method as it bypasses the models manager but I know many people do it.

### DIFF
--- a/unit-tests/ModelsTest.php
+++ b/unit-tests/ModelsTest.php
@@ -812,4 +812,30 @@ class ModelsTest extends PHPUnit_Framework_TestCase
 		$personer->refresh();
 		$this->assertEquals($personerData, $personer->toArray());
 	}
+
+	public function testModelPrefixes()
+	{
+		$di = $this->_getDI(
+			function() {
+				require 'unit-tests/config.db.php';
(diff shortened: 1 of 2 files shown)


sergeyklay:
:+1:

unit-tests/ModelsTest.php
```diff
@@ -794,4 +794,23 @@ protected function _executeTestsRenamed($di)
...
+
+		$di->set('modelsManager', function() {
+			$modelsManager = new Phalcon\Mvc\Model\Manager();
+			$modelsManager->setModelPrefix('test_');
+			return $modelsManager;
+		});
+
+		$persona = new Personas();
+
+		$this->assertEquals($persona->getSource(), 'test_personas');
```
ashpumpkin: According to phpunit doc expected value should go first

sergeyklay:
@SidRoberts Could you please rebase onto 2.1.x? Also update please CHANGELOG.md

sergeyklay:
Implemented in the `3.2.x` branch. Thank you for contributing

PR #10350, opened 2015-05-14
Fixing Phalcon\Dispatcher::getHandlerClass

DasRed:
The generation of the handler class in the dispatcher is wrong. If the handler name does not start with \, the first char of the name will be cutted.

This will result in something like this: Api\V1\Controller\etail_tableController

but it shoud be: Api\V1\Controller\**D**etailTableController


sergeyklay:
Hm.. this issue still not solved? See #10292

DasRed:
currently it is not solved in master or 2.0.1 release but it is merged into branch phalcon:2.0.x... the other solution in #10292 is okay but not my favorite, because handler names with a underscore, will not be camelized. :(

DasRed:
removinf

PR #10359, opened 2015-05-15
2.0.x PDOException: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'username.tablename' doesn't exist 

lieshanhe:
-  I find my table via findFirst

``` javascript
$user = \Model\User::findFirst([                                       
    "(email :email: OR id = :email:) AND pw = :password:'",  
    'bind' => array('email' => $email, 'password' => sha1($password))                     
]);                                                                                       
```

and is indeed get right data.
-   I update the instance 

``` javascript
$user->setLoginTime(date("Y-m-d H:i:s"));                  
$user->setLoginIp($this->request->getClientAddress());     
```
-  I save the instance.

``` javascript
if ($user->update() == false) {                             
    $user->rollback("Cannot Update User Login info");   
    foreach ($user->getMessages() as $message) {            
        echo $this->flashSession->error($message);              
    }                                                           
}                                                               
```

I got the error

> PDOException: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'username.tablename' doesn't exist 

so, I am confused that why I can't save the data, and also why is

> username.tablename' 

not 

> databasename.tablename' 

?


andresgutierrez:
Could you please create your problem as an issue? https://github.com/phalcon/cphalcon/issues/new

zikezhang:
Okay

PR #10373, opened 2015-05-19
Enable 32 row prefetch for mvc\model\resultset\complex

patrick-zippenfenig:
32 row prefetch was only available for simple resultset.
Additionally reduced code redundancies in current() 

Todo: Implement ["chunks of 32 row](http://docs.phalconphp.com/en/latest/reference/models.html#model-resultsets). Unfortunately this requires rewrite of most resultset code....

See: https://github.com/phalcon/cphalcon/pull/10100#issuecomment-103437907

### DIFF
--- a/phalcon/mvc/model/resultset.zep
+++ b/phalcon/mvc/model/resultset.zep
@@ -138,27 +138,33 @@ abstract class Resultset
 			if this->_row === null || this->_pointer != position {
 				let result = this->_result;
 				if result !== false {
-					/**
-					* 1. If row is not set, query is executed in "result->dataSeek()"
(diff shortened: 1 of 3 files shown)


patrick-zippenfenig:
Odd... all tests pass on my local testing environment

Green-Cat:
:+1:

patrick-zippenfenig:
I'm confused. Local tests work fine. Travis seems to crash at graphics functions or is missing test databases.
Last commit uses a parent constructor which moves almost all pdo-statement handling in abstract class and saves a lot of calls to count() etc.

andresgutierrez:
Thanks

PR #10374, opened 2015-05-19
added test to check exact exception message for 2.0.0 DI when service is missing

digitronac:
it was changed from since version 1 and some tests in our app were failing because of it. its not very important, but wont do any harm to include it :)

### DIFF
--- a/unit-tests/DiTest.php
+++ b/unit-tests/DiTest.php
@@ -381,6 +381,18 @@ class DiTest extends PHPUnit_Framework_TestCase
 		$this->assertTrue(true);
 	}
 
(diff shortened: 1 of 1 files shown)


digitronac:
i would, but i do not see where did i make indentation problem :(

sergeyklay:
Use tab instead spaces for indents here

selection_018

digitronac:
wow, was checking in PHPStorm and it didnt look like that ... will fix it now and will be more careful next time ;)

thanks!

andresgutierrez:
Thanks

PR #10377, opened 2015-05-19
[2.0.x] docs fix for docs.phalconphp.com

iGusev:
For example:
http://docs.phalconphp.com/en/latest/api/Phalcon_Mvc_Model_Validator_Inclusionin.html
http://docs.phalconphp.com/en/latest/api/Phalcon_Mvc_Model_Validator_Inclusionin.html
http://docs.phalconphp.com/en/latest/api/Phalcon_Mvc_Model_Validator_Numericality.html
etc.


andresgutierrez:
Thanks

PR #10379, opened 2015-05-20
Fix bug about subqueries in PHQL

dreamsxin:


andresgutierrez:
Thanks

ogarbe:
@andresgutierrez @dreamsxin hello, this commit (or the previous one about subquery) broke my project with the "Corrupted SELECT AST" exception (https://github.com/phalcon/cphalcon/blob/2.0.x/phalcon/mvc/model/query.zep#L1547) on a findFirst request

I try to search the root cause.

dreamsxin:
@ogarbe this commit, you must use zephir generate code and build it.

ogarbe:
yes, i do "../zephir/bin/zephir compile -c" on the 2.0.x cphalcon branch and try this against my project, i've changed all exception "Corrupted SELECT AST" to have more info and this is what i get : 

```zephir
if !fetch tables, select["tables"] {
                        throw new Exception("Corrupted SELECT AST - no table : " . json_encode(select));
}
```

```shell
Phalcon\Mvc\Model\Exception: Corrupted SELECT AST - no table : "port=3306;charset=utf8;persistent=1;dbname=voyapriv_common;host=vp-aix-sqloxyfront01.aix.vpg.lan"
```

dreamsxin:
Can you submit test code?

ogarbe:
first i try to get back to a working code, as i'm not sure if it's really this commit which created originally the pb, my last testing version was May 14th

ogarbe:
Sorry, after a lot of searches i've found the cause, it's due to optimization in dialect and something with var / string in zephir. Perhaps due to the fact i've got xdebug activated, i don't know.

here is a commit which tries to fix some part of the pb https://github.com/vpg/cphalcon/commit/f4d70b1c10ccfe4706f1e0a43371807ffee51314

andresgutierrez:
@ogarbe I fixed the problem, it was related to Zephir

PR #10384, opened 2015-05-20
Added status() to the Session adapter

niden:
- Added conditional on Session adapter start() to check if the session has already been started
- Added status() function in Session adapter to return the status of the session (disabled/none/started)


andresgutierrez:
Thanks

PR #10388, opened 2015-05-21
Phalcon\Validation::getDefaultMessage() should now return null instead of phalcon/validation.zep warning if no message was found

digitronac:
this was default behavior of Phalcon\Validation::getDefaultMessage prior to version 2

### DIFF
--- a/phalcon/validation.zep
+++ b/phalcon/validation.zep
@@ -267,9 +267,14 @@ class Validation extends Injectable
 
 	/**
 	 * Get default message for validator type
+	 *
+	 * @param string type
 	 */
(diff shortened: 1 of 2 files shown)


digitronac:
sorry bout that ... still getting used to zephir and tab indentation

sergeyklay:
@digitronac use the `isset` that is faster also because is a language construct, not a function. For example http://juliusbeckmann.de/blog/php-benchmark-isset-or-array_key_exists.html

digitronac:
yep, i got it, pushed new commit to this branch yesterday. should be ok now i think

andresgutierrez:
Thanks

Issue #3033, opened 2014-11-14
Private class variables on models appear to be handled improperly.


tyler-shaw:
When creating a model extending from `\Phalcon\Mvc\Model`, class variables set as private are accessible from outside of the class without error. This applies only when changing the value. Attempting to use the value results in an error, as expected.

For example, let's say we have a plain php class, called MyClass:

	class MyClass {
		private $myVariable;
	}

If we were to create an instance of this class and then attempt to use that variable from outside the class, we would receive the following errors:

	$myClass = new MyClass();

	$myClass->myVariable = 20;
	// Result: Fatal error: Cannot access private property MyClass::myVariable in \File\Path\Here.php on line 3

	echo $myClass->myVariable;
	// Result: Fatal error: Cannot access private property MyClass::myVariable in \File\Path\Here.php on line 6
	
Now let's examine a similar scenario with a class that extends from `\Phalcon\Mvc\Model`.

	class MyModel extends \Phalcon\Mvc\Model {
		private $myDatabaseField;
	}
	
In this scenario, the private variable represents a field in the database that the model interacts with. Performing similar actions to the ones above:

	$myModel = new MyModel();
	
	$myModel->myDatabaseField = 20;
	// Result: Field is updated without notice of any type.
	
	echo $myModel->myDatabaseField;
	// Result: Access to undefined property MyModel::myDatabaseField in \File\Path\Here.php on line 6

Here is my issue with this:

- Allowing a private variable to be updated as if it is public completely defeats the purpose of having a private variable.
- Trying to access a private variable is treated as a notice when PHP itself considers such an action a fatal error, while less important to me, seems rather inconsistent.

Hopefully my example demonstrates the issue I am talking about. I am currently using 1.3.3 I believe, on PHP 5.5

daveamol:
The has not been done on phalcon 2.0 right. So temporarily to prevent using the magic setters and getters can one overwrite the __set function of the model with a dummy function ?

bmoore:
This is possibly being handled in https://github.com/phalcon/cphalcon/pull/11320

PR #10391, opened 2015-05-22
[2.0.x] Mvc/Model: Setter fallback

brainformatik:
Add setter fallback in "__set" magic.
This is useful if you use protected (private) properties and assign a value directly to a property.
If this property has setter/getter functions they will be called preferably.

### DIFF
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -3928,7 +3928,7 @@ abstract class Model implements ModelInterface, ResultInterface, InjectionAwareI
 	public function __set(string property, value)
 	{
 		var lowerProperty, related, modelName, manager, lowerKey, relation, referencedModel,
-			key, item;
+			key, item, method;
 
 		/**
 		 * Values are probably relationships if they are objects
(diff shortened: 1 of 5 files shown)


sergeyklay:
@brainformatik  Can you please provide a small test?

In any case :+1:

Green-Cat:
:+1:

brainformatik:
Added a few tests

sergeyklay:
Fine! :+1:

Surt:
I would like to see this one in collection too. By the way I would like to see some "entity" behaviours interfaces in both, collections and models. As well a ResultSet kind of data for both, Models resultsets and ODM multiple results.

noobiwan:
What's the status on this?

noobiwan:
#10391

noobiwan:
Similar issue: #3033

sergeyklay:
@brainformatik Can you please rebase onto `2.0.x` branch and amend CHANGELOG.md?

bmoore:
This will conflict with #11320 if we try to bring #11320 to 2.0.x

bmoore:
https://github.com/phalcon/cphalcon/pull/11417 Is the ported code from #11320 for 2.0.x

PR #10412, opened 2015-05-26
Removed `underscore` variable from Mvc\Model\Resultset\Complex::current()

SidRoberts:


sergeyklay:
@SidRoberts  As part of optimization?

SidRoberts:
I assumed it was from when the code was written in C.

Issue #10239, opened 2015-05-04
not set dynamically cache value time!...this is a bug?


hardlick:
in volt template:
login.volt
i have:

```
{% cache "loginPage" 28800 %}
    <body>...</body>
{% endcache %}
```

How to set the value 28800 that is in second, change for a variable ? like:
for example:

```
{% set variable = 54545 %}
{% cache "loginPage" variable %}
```

I try with this code:
    {% set variable = 54545 %}
    {% cache "loginPage" variable %}
but, Not!, not working, this is the error message:

Fatal error: Uncaught exception 'Phalcon\Mvc\View\Exception' with message 'Syntax error, unexpected token IDENTIFIER(variable)

this is my volt page:

```
{% set variable = 54545 %}
{% cache "loginPage" variable %}
```

PR #10418, opened 2015-05-27
Fix bug #10239

dreamsxin:


andresgutierrez:
I think it could be more powerful to use an expression for the cache lifetime instead of identifier or integer

dreamsxin:
@andresgutierrez The best way is to modify the parameters of the cache, like this:
```php
{% cache "sidebar", 3600 %}
    <!-- generate this content is slow so we are going to cache it -->
{% endcache %}
```

andresgutierrez:
Is the parser having conflicts if you use an expression?

dreamsxin:
@andresgutierrez Yes, will conflict, error: `Not enough precedence: error`
```lemon
cache_statement(R) ::= OPEN_DELIMITER CACHE expr(E) expr(N) CLOSE_DELIMITER statement_list(T) OPEN_DELIMITER ENDCACHE CLOSE_DELIMITER . {
```
There is something wrong with the priority.

andresgutierrez:
Ok, so I think we're fine with this

PR #10433, opened 2015-05-28
Implements the Blameable behavior

Green-Cat:
Adds the blameable behavior, ported from incubator with a few fixes and tests added.


Green-Cat:
Will resubmit.

PR #10434, opened 2015-05-28
Implements the Blameable Behavior for Models

Green-Cat:
Adds the blameable behavior, ported from incubator with a few fixes and tests added.

### DIFF
--- a/phalcon/mvc/model/behavior/blameable.zep
+++ b/phalcon/mvc/model/behavior/blameable.zep
@@ -48,18 +48,15 @@ class Blameable extends Behavior implements BehaviorInterface
     /**
      * Creates an Audit instance based on the current enviroment
      */
-    internal function createAudit(string! type, <ModelInterface> model) -> <Audit>
+    private function createAudit(string! type, <ModelInterface> model) -> <Audit>
     {
         var session, request, audit;
-
-        //Get the session service
+        
         session = model->getDI()->getSession();
-
-        //Get the request service
         request = model->getDI()->getRequest();
-
+        
         audit = new Audit();
-
+        
         //Get the username from session
         audit->user_name = session->get("userName");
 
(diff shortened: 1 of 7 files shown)


andresgutierrez:
Could you please rebase?

andresgutierrez:
@Green-Cat ping

Green-Cat:
@andresgutierrez Will do.

phalcon/mvc/model/behavior/blameable.zep
```diff
@@ -0,0 +1,138 @@
...
+     * Creates an Audit instance based on the current enviroment
+     */
+    private function createAudit(string! type, <ModelInterface> model) -> <Audit>
+    {
+        var session, request, audit;
+        
+        session = model->getDI()->getSession();
+        request = model->getDI()->getRequest();
+        
+        audit = new Audit();
```
ogarbe: i think this class should be returned by DI because if not it's mandatory to create an Audit object.

phalcon/mvc/model/behavior/blameable.zep
```diff
@@ -0,0 +1,138 @@
...
+        var audit, metaData, fields, field, auditDetail;
+        array details;
+
+        //Create a new audit
+        audit    = this->createAudit("C", model);
+        metaData = model->getModelsMetaData();
+        fields   = $metaData->getAttributes($model);
+
+        foreach (fields as field) {
+            auditDetail = new AuditDetail();
```
ogarbe: same here, i think this class should be returned by DI because if not it's mandatory to create an Audit object.

phalcon/mvc/model/behavior/blameable.zep
ogarbe: perhaps also get the class name in the behavior constructor.

Green-Cat:
I will rework this and resubmit for 2.1.x soon-ish.

sergeyklay:
@Green-Cat

sergeyklay:
@Green-Cat ping

PR #10443, opened 2015-05-30
New system for column types

wamxx:
I changed the system of column types to be more extensible , now we can add more column type with php : 

Create column type class : 

``` php
namespace MyNamespace\ColumnType;

use Phalcon\Db\Column\Type as ColumnType;
use Phalcon\Db\Column;

class Smallint extends ColumnType
{
    public function setup()
    {
        $this->dialects = [
                "mysql":"SMALLINT(#size#)",
                "sqlite":"SMALLINT",
                "postgresql":"SMALLINT"
            ];
        $this->_autoIncrement = true;
        $this->_scale = true;
        $this->_isNumeric = true;
        $this->_bindType = Column::BIND_PARAM_INT;
    }
    public function castValue(value) {
        return (int)value;
    }
}
```

And add your type in the system : 

```
Phalcon\Db\Column::addType( (new MyNamespace\ColumnType\Smallint()) );
```

Now you can use this new type in your model annotations for example.

I'm open for all advices and contributions for optimise this, i'm a beginner with Zephir.

### DIFF
--- a/phalcon/db/column.zep
+++ b/phalcon/db/column.zep
@@ -257,7 +257,7 @@ class Column implements ColumnInterface
 	
 	public static function initialize() {
 		
-		var type,className,dialects,dialect,dialectType;
+		var type,className,dialects,dialect,dialectType,dialectTypes,dtypes;
 	
 		if empty self::columnTypes {
 			let self::columnTypes  = [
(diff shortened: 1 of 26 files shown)


phalcon/db/column.zep
```diff
@@ -339,11 +502,25 @@ class Column implements ColumnInterface
...
-		}
+		let this->_bindType = this->_type->getBindType();
 
 	}
+	
+	/**
+	 *
+	 *
+	 *
+	 *
```
Green-Cat: Why the large empty docblock?

Green-Cat:
Liking the functionality, not sure about the performance. Would this not add 10s of extra function calls for every query parsed?

wamxx:
@Green-Cat not so much, it's more the side of the memory consumption will increase a little bit. I would have to do a benchmark.

sjinks:

PR #10444, opened 2015-05-30
Convert format CHAGELOG to MD

KorsaR-ZN:
[see](https://github.com/KorsaR-ZN/phalcon-core/blob/bb180d1a062f9fd34cd4d43cc4d1a521207a572b/CHANGELOG.md)


niden:
Thank you

PR #10459, opened 2015-06-02
Added SPEC file for RPM package

GioMac:


sergeyklay:
1. `master` is freezed
2. Why just not create a separated repo for rpm (like ppa for ubuntu)?

GioMac:
1. where to move?
2. creating repo is not recommended and not in best practice in modern days: requires more decentralized manpower for maintenance, which probably won't know about changes in upstream release that must be reflected in spec. and this must be done for every single distribution or delegated to other services like OBS. Also, this is the cause why we have terrible situation and support. Package management becomes headache then. Official SPEC means more improvements, centrally, for everyone.
screen shot 2015-06-02 at 14 56 34

andresgutierrez:
[phalcon-ppa]() uses a submodule to have the latest changes in cphalcon https://github.com/phalcon/phalcon-ppa

sergeyklay:
@GioMac We can create official repo for RPMs. Tell me if you'll able to maintain it

GioMac:
@sergeyklay That's good! I can maintain it, when/where can we discuss it?

sergeyklay:
@GioMac please contact me (serghei@phalconphp.com) at your convenience

PR #10464, opened 2015-06-03
Change criteria to get model name from get method on execute

renatomjr:
This change implements the same behaviour of the Phalcon\Mvc\Model::getSource() method to the Phalcon\Mvc\Model\Criteria::getModelName() method.
Phalcon\Mvc\Model\Criteria::execute() now gets the model name from the method instead of the property.


KorsaR-ZN:
@renatomjr Please move your PR to branch 2.0.x

PR #10473, opened 2015-06-05
Fixes router not-found

blindshadow:
When use short route path on not-found, dispatcher throw a exception.
$router->notFound('Index::notFound');


andresgutierrez:
What exception did you get before this change?

blindshadow:
@andresgutierrez 
Phalcon\Mvc\Dispatcher\Exception from Phalcon\Dispatcher setParams method, And get message：Parameters must be an Array
Short path will not trans to array in router.

Issue #10461, opened 2015-06-02
Db Profiler docs fix


michanismus:
It is not possible to assign an instance of Phalcon\Db\Profiler to an adapter.

http://docs.phalconphp.com/en/latest/api/Phalcon_Db_Profiler.html

andresgutierrez:
I think the documentation must be changed as the correct functionality is: https://docs.phalconphp.com/en/latest/reference/db.html#profiling-sql-statements

PR #10481, opened 2015-06-06
Support set the connection profiler

dreamsxin:


SidRoberts:
References #10461

andresgutierrez:
This will add more overhead to the database component, maybe update the documentation to this example? https://docs.phalconphp.com/en/latest/reference/models.html#profiling-sql-statements

dreamsxin:
@andresgutierrez Use the event, the `executePrepared` and `execute` can't captured, and not enough easy to use

dreamsxin:
Can change `Profiler` to this:
```php
class Profiler
{
    public function beforeQuery($event, $db, $bindParams = NULL)
    {
        // TODO
    }

    public function afterQuery($event, $db, $bindParams = NULL)
    {
        // TODO
    }
}
$profiler = new Profiler;
$eventsManager->attach('my-component', $profiler);
```

PR #10491, opened 2015-06-09
fixed bug

admpub:


andresgutierrez:
We don't add changes directly to master, could you please submit this to the branch 2.0.x?

PR #10494, opened 2015-06-09
Removes unnecessary nested if

Green-Cat:


andresgutierrez:
It it was my decision we well never use elseif in phalcon :)

Green-Cat:
@andresgutierrez There are several instances of it in the codebase already though.

Green-Cat:
Even in the exact same file, https://github.com/phalcon/cphalcon/blob/c96782766cfe5852acb821af944140379125c1cb/phalcon/db/dialect.zep#L188

andresgutierrez:
I'm not against it, just personal tastes :+1:

PR #10511, opened 2015-06-15
autoBegin not transmited to transaction's constructor when get() called

xueron:
small fix: autoBegin not transmited to transaction's constructor when get() called


andresgutierrez:
Thanks

PR #10515, opened 2015-06-16
[2.0.x]  "static::" instead of "self::" in \Phalcon\Version class

racztiborzoltan:
Reason: simpler class extending

Example: 

``` php
namespace MyFooBar {

    class Version extends \Er\Version
    {
        protected static function _getVersion()
        {
            return [6, 5, 4, 3, 2];
        }
    }
}
namespace {
    echo \MyFooBar\Version::get();
}
```

### DIFF
--- a/phalcon/version.zep
+++ b/phalcon/version.zep
@@ -121,11 +121,11 @@ class Version
 
 		let version       = static::_getVersion();
 
-		let major         = version[static::VERSION_MAJOR],
-			medium        = version[static::VERSION_MEDIUM],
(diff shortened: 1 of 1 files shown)


andresgutierrez:
```
Zephir\ParseException: Syntax error in /home/travis/build/phalcon/cphalcon/phalcon/version.zep on line 124
	  let major         = version[static::VERSION_MAJOR],
	--------------------------------------------------^
```

racztiborzoltan:
Fixed! (I hope)

PR #10525, opened 2015-06-19
Improved regular expression in Mvc\Url to detect external resources

sergeyklay:
- Cleanup local regexp
- Digits for matching need for example for `e2dk` protocol


andresgutierrez:
It needs two regex (or less/more) to match these:

-  //localhost/something
- http://localhost/something
- mailto:something@something.com

sergeyklay:
@andresgutierrez 
- Added support for `mailto:something@something.com`
- `http://localhost/something` works fine

Can you explain please `//localhost/something` ?

andresgutierrez:
It takes the current scheme according to the served page: http://www.paulirish.com/2010/the-protocol-relative-url/

sergeyklay:
Ok, added support for CDN

Issue #10536, opened 2015-06-23
Implement distinct/setDistinct on Phalcon\Mvc\Model\Criteria like on query builder


williamokano:
You guys could add the Distinct from the QueryBuilder into the Phalcon\Mvc\Model\Criteria

I think that it's more practical to do MyModel::query() instead of creating the whole query from the query builder.

As I could see, there's no Distinct method on the Criteria class.

Example: https://gist.github.com/williamokano/a17cb07d791da65bc997

I could use just ->distinct(true) and my problem was solved.

Hi guys, I was wondering if is possible to get distinct values with a query built with the model criteria.

Follow an example (PS.: don't bother with the lack of the bind parameter, it was just a test):

https://gist.github.com/williamokano/a17cb07d791da65bc997

In this case, when I have 1 user that already lived on a city A (id=1) and on a city B (id=2) I will get the user twice (or more, that depends on how many cities he lived).

I know that it's caused because of the intermediate table called lived, and I get the same problem creating "raw queries". But in raw queries I can add the DISTINCT clause and I couldn't find how to do that with the Criteria class.

PS.: I know that if I use the query builder I can set distinct(true) and my problem is solved, but my question is exactly on how to use distinct with Criteria and not with the QueryBuilder.

Regards and apologies about the bad english.

https://docs.phalconphp.com/en/latest/api/Phalcon_Mvc_Model_Criteria.html

KorsaR-ZN:
@williamokano see #10538

williamokano:
Thanks, I didn't found this issue. Closed and waiting for the merge :)

Thanks @KorsaR-ZN :+1:

williamokano:
Feature merged!

Thanks @KorsaR-ZN!

PR #10538, opened 2015-06-23
[2.0.x] Added method distinct() in Mvc\Model\Criteria

KorsaR-ZN:
Fix for #10536

### DIFF
--- a/phalcon/mvc/model/criteria.zep
+++ b/phalcon/mvc/model/criteria.zep
@@ -114,6 +114,15 @@ class Criteria implements CriteriaInterface, InjectionAwareInterface
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


williamokano:
@andresgutierrez :+1:

phalcon/mvc/model/criteria.zep
```diff
@@ -114,6 +114,15 @@ class Criteria implements CriteriaInterface, InjectionAwareInterface
 	}
 
 	/**
+	 * Sets SELECT DISTINCT / SELECT ALL flag
+	 */
+	 public function distinct(var distinct) -> <Criteria>
```
Green-Cat: Shouldn't it be a boolean?

PR #10550, opened 2015-06-26
fix gettext for Linux

duythien:


andresgutierrez:
Hello, we don't add changes directly to master, could you please submit this to the 2.0.x branch?
Also, please don't submit changes to ext/ folder

PR #10575, opened 2015-06-30
fix #10573 - empty value replaced by null when cast on hydrate

### LINKED ISSUE #10573: Problem with nullable columns and cast on hydrate
The cast_on_hydrate function creates a bug when a column is nullable

Table document : 
id int
name string
parent_document int (nullable) which is a FK to document.id

my model Document can have a parent document which is Document->Parent but a document can have "no parent"

if i activate cast on hydrate => all documents with no parents become document with parent document 0 (which does not exists..)

in my opinion a test should be made here https://github.com/phalcon/cphalcon/blob/2.0.x/phalcon/mvc/model.zep#L541-L555 to validate value is not empty and replace castValue by NULL in this case.

@andresgutierrez what do you think of that ?

ogarbe:
cf #10573

### DIFF
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -538,25 +538,42 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 						continue;
 					}
 
-					switch attribute[1] {
+					if value != "" {
+						switch attribute[1] {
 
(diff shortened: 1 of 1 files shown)


andresgutierrez:
Thanks

PR #10576, opened 2015-06-30
[2.0.x] [NFR] Added fastest alias in Phalcon\Mvc\Controller for redirect and forward

KorsaR-ZN:


andresgutierrez:
This violates the [SRP](https://en.wikipedia.org/wiki/Single_responsibility_principle)

KorsaR-ZN:
@andresgutierrez This does not violate the SRP, because the class does not appear the new responsibilities he did the appropriate classes to delegate a proxy methods...

andresgutierrez:
Yeah, but you depend on Mvc\Controller to make redirections or forwards

KorsaR-ZN:
@andresgutierrez Yes you are right.
I make another PR which adds only the interface to the controller.

PR #10603, opened 2015-07-07
The filter now no longer removes the ' character, fixes #10595

### LINKED ISSUE #10595: Filters - email removes the ' character
According to Wikipedia the following special characters are allowed in an email address:
```
"These special characters: # - _ ~ ! $ & ' ( ) * + , ; = : and percentile encoding i.e. %20"
```

https://en.wikipedia.org/wiki/Email_address#Local_part

Also when you do a 'native' php filter_var call, the `'` character is not removed:

```
$email = "some'email@example.com";
echo filter_var($email, FILTER_SANITIZE_EMAIL); // some'email@example.com
```

Why does Phalcon remove the `'` character?

https://github.com/phalcon/cphalcon/blob/master/phalcon/filter.zep#L150

AmazingDreams:
Regarding issue #10595

### DIFF
--- a/phalcon/filter.zep
+++ b/phalcon/filter.zep
@@ -147,7 +147,7 @@ class Filter implements FilterInterface
 				/**
 				 * The 'email' filter uses the filter extension
 				 */
-				return filter_var(str_replace("'", "", value), constant("FILTER_SANITIZE_EMAIL"));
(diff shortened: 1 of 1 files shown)


andresgutierrez:
Thanks

PR #10614, opened 2015-07-08
Bug fix in GD adapter _pixelate function

pauliuspetronis:
https://forum.phalconphp.com/discussion/7365/image-adapter-pixelate-bug


andresgutierrez:
Thanks

Issue #10631, opened 2015-07-09
Phalcon\Logger\Formatter\Line: no more newline at end of message like in 1.3.x


olivier-monaco:
In 1.3.x, Phalcon\Logger\Formatter\Line::format added a newline a the end. In 2.0.x, no more. Instead, Phalcon\Logger\Adapter\File add a newline before calling logInternal. But we use Phalcon\Logger\Adapter\Stream that don't add a newline.

Use case:
<pre>
use Phalcon\Logger\Adapter\Stream;
use Phalcon\Logger;

$logger = new Stream('php://output');
$logger->setLogLevel(Logger::INFO);

$logger->log(Logger::ERROR, 'Error');
$logger->log(Logger::INFO, 'Info');
</pre>

Is this a feature or a bug?

andresgutierrez:
It looks like something that can be fixed

olivier-monaco:
Ok. I will in some hours. Thanks.

andresgutierrez:
Fixed in the 2.0.x branch

PR #10634, opened 2015-07-10
2.0.x logger newline

olivier-monaco:
Phalcon\Logger\Formatter\Line: no more newline at end of message like in 1.3.x (#10631)

Note: I didn't understand where to put unit test.


andresgutierrez:
```
1) LoggerTest::testIssues2262
Failed asserting that 3 matches expected 1.
/home/travis/build/phalcon/cphalcon/unit-tests/LoggerTest.php:54
2) LoggerTest::testIssues2798
Failed asserting that 3 matches expected 1.
/home/travis/build/phalcon/cphalcon/unit-tests/LoggerTest.php:96
FAILURES!
Tests: 421, Assertions: 11571, Failures: 2, Skipped: 26.
```

olivier-monaco:
testIssues2262 does not test only the 2262 issue but also the 2798...

olivier-monaco:
In fact the problem is elsewhere... testIssues2262 and testIssues2798 include a test for something not related to these issues...

And the test is, in my opinion, wrong...

Test:
<pre>
                $logger = new \Phalcon\Logger\Adapter\File($logfile);
                $logger->setFormatter(new \Phalcon\Logger\Formatter\Json());
                $logger->log('This is a message');
                $logger->log("This is an error", \Phalcon\Logger::ERROR);
                $logger->error("This is another error");
</pre>

Expected result:
<pre>
{"type":"DEBUG","message":"This is a message","timestamp":1436773428}
{"type":"ERROR","message":"This is an error","timestamp":1436773428}
{"type":"ERROR","message":"This is another error","timestamp":1436773428}
</pre>

Actual result with my PR:
<pre>
{"type":"DEBUG","message":"This is a message","timestamp":1436773428}{"type":"ERROR","message":"This is an error","timestamp":1436773428}{"type":"ERROR","message":"This is another error","timestamp":1436773428}
</pre>

Problem occurs because I no more add a PHP_EOL at the end of the log line in Phalcon\Logger\Adapter\File::logInternal. The PHP_EOL was added is commit 0e7dbec301aae019d9300c47591b395d85caf747: "Logger fix, replace emergence with emergency #2638, enable LoggerTest". The commit has nothing to do with adding à PHP_EOL and related issue.

See. https://github.com/phalcon/cphalcon/commit/0e7dbec301aae019d9300c47591b395d85caf747#diff-ed06fe9c7eeb1c2f708ab625ea64b340R123

My opinion is to change the unit test but this can break compatibility with 2.0.x branch. However, the current implementation breaks compatibility with 1.3.x...

@andresgutierrez What do you think?

andresgutierrez:
@olivier-monaco Please fix the test so it will work with 1.3.x

olivier-monaco:
In the Json formatter, I added a newline at the end of encoded data. Otherwise, the resulting data is hard to parse. I also added a unit test for this issue.

andresgutierrez:
Thanks

PR #10644, opened 2015-07-12
Fix #1933

### LINKED ISSUE #1933: Odd view behavior with layouts + pick
I noticed some odd behavior with calling `setLayout()` and `pick()`.

```
1. $view->setLayout('fancy');
2. $view->pick('someview');
```

1. Turns out if I call `setLayout()` then `pick()` then the `setLayout()` gets ignored. If there is a default layout, that gets picked which is worse since the framework ignores user intent (line 1).

2. The work around right now is to call `pick(['view', 'layout'])`. This behavior is undocumented and I gleaned it by reading the C source. Is this undocumented because it's unsupported and liable to change or is this okay to use?

3. It would be nice to be able to fix this bug and leave out the undocumented array syntax since the code legibility is better and it allows full control to the user to compose views throughout the application lifecycle (set a main, a layout and a view at different stages of the app lifecycle).

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1896481-odd-view-behavior-with-layouts-pick?utm_campaign=plugin&utm_content=tracker%2F50707&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F50707&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ryomo:
Fix an odd behavior with `$view->setLayout()` and `$view->pick()`.
This may not be a bug :confused: :question:

### DIFF
--- a/phalcon/mvc/view.zep
+++ b/phalcon/mvc/view.zep
@@ -803,8 +803,10 @@ class View extends Injectable implements ViewInterface
 			 * The 'picked' view is an array, where the first element is controller and the second the action
 			 */
 			let renderView = pickView[0];
-			if fetch pickViewAction, pickView[1] {
-				let layoutName = pickViewAction;
+			if layoutName === null {
+				if fetch pickViewAction, pickView[1] {
+					let layoutName = pickViewAction;
+				}
 			}
 		}
 
--- a/unit-tests/ViewTest.php
+++ b/unit-tests/ViewTest.php
@@ -223,6 +223,28 @@ class ViewTest extends PHPUnit_Framework_TestCase
 		$this->assertEquals($view->getContent(), '<html>Well, this is the view content: here.</html>' . PHP_EOL);
 	}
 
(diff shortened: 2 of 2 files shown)


ryomo:
More information.

When I call both `setLayout('layoutName')` and `pick('path/to/view')`, `setLayout()` gets ignored, because `layoutName` is always overwritten by `pick()`.
I fixed this behavior.

But the behavior is intended, this pull request should not be merged.

benfavre:
This is anoying.
+1

sergeyklay:
+1

sergeyklay:
@ryomo

ryomo:
@sergeyklay 
I rebased this.

sergeyklay:
Can you please send this PR to the `2.1.x` branch?

PR #10659, opened 2015-07-14
Fixed Postgres DB dialect from adding two pairs of quotes in dropTable()

SidRoberts:


andresgutierrez:
But if the name of table is a reserved word?

SidRoberts:
The current code adds quotes twice:

image

This PR will only add them once:

```
DROP TABLE IF EXISTS "articles"
```

andresgutierrez:
I see, Thanks

PR #10671, opened 2015-07-15
Add error code in Memcached

dugwood:
I'm having some trouble with Memcached->set() and I can't access the «getResultCode()» directly (this->_memcache is private). So adding the error code in the Exception message would help.


andresgutierrez:
Thanks

PR #10672, opened 2015-07-15
Reverted query builder to fix too many bindings error from LIMIT clause

cadencelabs-master:
I removed the new code for organizing the LIMIT clause with a binding - in getPhql() it's incrementing the hiddenParam variable _each_ time getPhql() is called - meaning when you execute a query with a limit, it can throw a "too many bindings" error (for bindings than tokens)


andresgutierrez:
We don't add changes to master.

Also, sorry but this will not be merged this reverts the use of bound parameters in LIMIT/OFFSET clauses.

alanbarber111:
The bound parameters are causing a problem and I don't know how to fix, this should be merged until that can be done - please advise

andresgutierrez:
Yeah but reverts the old behavior introduce a new problem in exchange of another

alanbarber111:
Can you recommend some way to track whether the limit clause has been bound so that calling getPhql doesn't add a new binding for the limit clause each time? I'm happy to implement

Issue #10648, opened 2015-07-14
[2.0.4 bug] Cannot use a scalar value as an array


galki:
I'm using PHP 5.6.8 and getting the below notices when I have error reporting turned on.
It looks like the way an array is written as another array's key is the cause.

// Notice: Cannot use a scalar value as an array in phalcon/mvc/model/resultset/simple.zep on line 209
```let renamed[renamedKey[0]] = value;```

// Notice: Cannot use a scalar value as an array in phalcon/mvc/model.zep on line 3600
```let snapshot[attribute[0]] = value;```

**Comment by olivier-monaco** [comment]
@andresgutierrez look here: https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/model/resultset/simple.zep#L205

<pre>
		                    if !fetch renamedKey, renamedKey[0] {
                	            throw new Exception("Column '" . key . "' is not part of the column map");
                        	}

							let renamed[renamedKey[0]] = value;
</pre>

We replace renamedKey by its first item but them ask for the first item of renamedKey...

It must be
<pre>
let renamed[renamedKey] = value;
</pre>

Same for the other notice.

**Comment by olivier-monaco** [comment]
Usecase:
```php
<?php
$di = new Phalcon\DI();

$di['db'] = new Phalcon\Db\Adapter\Pdo\MySQL([
    "host"     => "localhost",
    "username" => "phalcon_test",
    "password" => "phalcon_test",
    "dbname"   => "phalcon_test",
]);

\Phalcon\Mvc\Model::setup(['castOnHydrate' => true]);

$di['modelsManager'] = new Phalcon\Mvc\Model\Manager();
$di['modelsManager']->setDi($di);
$di['modelsMetadata'] = new Phalcon\Mvc\Model\MetaData\Memory();

class Robot extends Phalcon\Mvc\Model
{
    public $id;
}

$q = new Phalcon\Mvc\Model\Query('select * from Robot limit 1', $di);
$r = $q->execute();
var_dump($r->toArray());
```

Result:
```
PHP Notice:  Cannot use a scalar value as an array in phalcon/mvc/model/resultset/simple.zep on line 209 in /volumes/data1/vhosts-dev/libraries/cphalcon-olivier/issues/i10673a.php on line 24
PHP Stack trace:
PHP   1. {main}() /volumes/data1/vhosts-dev/libraries/cphalcon-olivier/issues/i10673a.php:0
[...]

PR #10673, opened 2015-07-15
Notice "Cannot use a scalar value as an array"

olivier-monaco:
#10648

### DIFF
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -3597,10 +3597,9 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 							continue;
 						}
 					}
-					let snapshot[attribute[0]] = value;
-				} else {
-					let snapshot[attribute] = value;
 				}
+
+				let snapshot[attribute] = value;
 			}
 
 			let this->_snapshot = snapshot;
--- a/phalcon/mvc/model/resultset/simple.zep
+++ b/phalcon/mvc/model/resultset/simple.zep
@@ -205,11 +205,9 @@ class Simple extends Resultset
 		                    if !fetch renamedKey, renamedKey[0] {
                 	            throw new Exception("Column '" . key . "' is not part of the column map");
                         	}
-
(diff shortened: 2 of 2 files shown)


andresgutierrez:
Could you please provide the script or unit-test you used to test this?

olivier-monaco:
For simple.zep, I don't have any test. It's by code review. It revert the following commit: https://github.com/olivier-monaco/cphalcon/commit/4f5e28abfbe586f43d5ee6c41d3901b91a85d02a#diff-0a0014cdd026fc43956d243d1ecebbcdL202

The current code is doing something like:
<pre>
let renamed[this->_columnMap[key][0][0]] = value
</pre>

But a I think column map entry can't be array of array.

Same for model.zep.

andresgutierrez:
They can be arrays, it you're using orm.castOnHydrate = true

olivier-monaco:
Interesting... I'm working on a use case...

olivier-monaco:
Using use case from https://github.com/phalcon/cphalcon/issues/10648#issuecomment-121749438.

I added a var_dump in simple.zep just before using renamedKey:
```php
						/**
						 * Check if the key is part of the column map
						 */
						if !fetch renamedKey, columnMap[key] {
							throw new Exception("Column '" . key . "' is not part of the column map");
						}

var_dump("SIMPLE", renamedKey);
                        if typeof renamedKey == "array" {
		                    if !fetch renamedKey, renamedKey[0] {
                	            throw new Exception("Column '" . key . "' is not part of the column map");
                        	}

							let renamed[renamedKey[0]] = value;
		                } else {
							let renamed[renamedKey] = value;
						}
```

The result:
```
string(6) "SIMPLE"
array(2) {
  [0]=>
  string(2) "id"
  [1]=>
  int(0)
}
PHP Notice:  Cannot use a scalar value as an array in phalcon/mvc/model/resultset/simple.zep on line 209 in .../i10673a.php on line 24
```

It's not an array of array.

andresgutierrez:
Thanks

Issue #10689, opened 2015-07-18
[NFR] Response events


iotch:
Hello. Is it a bad idea to support events in `Phalcon\Http\Response` component? For example, `'response:beforeSendHeaders'` to modify headers depending on response contents. Thanks.

PR #10690, opened 2015-07-18
[2.0.x] [NFR] Added events response:beforeSendHeaders, response:afterSendHeaders

KorsaR-ZN:
see #10689

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# [2.0.8](https://github.com/phalcon/cphalcon/releases/tag/phalcon-v2.0.8) (2015-XX-XX)
+- Added events `response:beforeSendHeaders` and `response:afterSendHeaders` ([#10689](https://github.com/phalcon/cphalcon/issues/10689)
+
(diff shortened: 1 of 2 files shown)


andresgutierrez:
I think this event should be in Mvc\Application instead of here

iotch:
@andresgutierrez Response events is useful when using the component stand-alone. For example if I don't need a full-stack Mvc\Applcation, and only a request, router, dispatcher and response. Router and dispatcher are sending events, will be great to have the response to.

KorsaR-ZN:
@iotch :+1:

KorsaR-ZN:
@andresgutierrez 

Quote from the docs:
>  What is Phalcon?
 ....
Phalcon also is loosely coupled, allowing you to use its objects as glue components based on the needs of your application.


If the event is moved `Mvc\Application`, it will not be able to use `Http\Response` as not dependent component.

iotch:
@andresgutierrez Please, accept the request, no one else has commented against this change and this will not break anything

PR #10700, opened 2015-07-22
Fixed annotation errors

greew:
Added the backslash in front of Phalcon in annotation tags, where it was missing.
Also added a backslash in front of a Closure type hint, where it was missing.

### DIFF
--- a/phalcon/mvc/modelinterface.zep
+++ b/phalcon/mvc/modelinterface.zep
@@ -101,7 +101,7 @@ interface ModelInterface
 	/**
 	 * Assigns values to a model from an array
 	 *
-	 * @param Phalcon\Mvc\Model object
+	 * @param \Phalcon\Mvc\Model object
 	 * @param array data
 	 * @param array columnMap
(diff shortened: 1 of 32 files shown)


phalcon/mvc/modelinterface.zep
```diff
@@ -158,7 +158,7 @@ interface ModelInterface
 	/**
 	 * Create a criteria for a especific model
 	 *
-	 * @param Phalcon\DiInterface dependencyInjector
+	 * @param \Phalcon\DiInterface dependencyInjector
 	 * @return Phalcon\Mvc\Model\CriteriaInterface
```
greew: Actually, what about return types as here - didn't they ought to be prefixed with a backslash as well?

andresgutierrez:
Thanks

phalcon/mvc/modelinterface.zep
andresgutierrez: I think they must be also prefixed

PR #10703, opened 2015-07-22
Fixes to docblocks and parameter types

ovr:


andresgutierrez:
Thanks

wumouse:
Remove these docblocks will cause docs of phalcon api contains some unknown type

PR #10706, opened 2015-07-23
make transactionManager service name be concordant

xueron:
make transactionManager service name be concordant.

``` php
foreach ($di->getServices() as $service) {
    $realService = $di->get($service->getName());
} 
```

will throw:
Service 'transactions' wasn't found in the dependency injection container


andresgutierrez:
Thanks

PR #10713, opened 2015-07-24
Image\Adapter\Gd::save() no longer fails if the method or the instance is created with a filename without an extension

SidRoberts:
See #10709

Previously, something like these examples would have failed:

``` php
$image = new \Phalcon\Image\Adapter\Gd('image.jpg');
// ...
$image->save();
```

or...

``` php
$image = new \Phalcon\Image\Adapter\Gd('image.jpg');
// ...
$image->save('filename-without-extension');
```


andresgutierrez:
Thanks

Issue #10526, opened 2015-06-20
[2.0.x] INNER JOIN statement is added after an explicit JOIN on the intermediate table


dschissler:
`Persons` and `Groups` are related with hasManyToMany.

The following causes the resultant SQL for PersonsGroups JOIN statement to be added twice because it is being implicitly added and its not taking into account that it was explicitly added.

This is broken:

``` php
    $Persons = "Webird\\Models\\Persons";
    $Groups = "Webird\\Models\\Groups";
    $PersonsGroups = "Webird\\Models\\PersonsGroups";
    // This is broken
    $phql = "SELECT $Persons.* FROM $Persons JOIN $PersonsGroups JOIN $Groups WHERE $Groups.id = :id: AND $PersonsGroups.type = :type: LIMIT 1";
```

So it needs to be removed:

``` php
    $phql = "SELECT $Persons.* FROM $Persons JOIN $Groups WHERE $Groups.id = :id: AND $PersonsGroups.type = :type: LIMIT 1";
```

but then the new v2.0.3 ModelsManager alias form can't be used because there is no way to reference the intermediate join table:

``` php
    $phql = "SELECT P.* FROM wb:Persons P JOIN wb:PersonsGroups PG JOIN wb:Groups G WHERE G.id = :id: AND PG.type = :type: LIMIT 1";

```

Issue #10705, opened 2015-07-22
[2.0.4] PHQL DISTINCT does not work for my case due to inappropriate implicit join


dschissler:
Between bugs #10526 and #10507 I'm tempted to remove all of the assigned relationships to severals core models that make up all of the interesting parts of my user data and to only use the assigned relationships for the mundane admin related models.

There is no way to get the phql below to work with the joining as it is now.  I need to be able to specify that Placement is joining against `SuperRecord` and not `Record`.

```php
        $phql = <<<PHQL
SELECT DISTINCT Record.*
FROM $Record Record
LEFT JOIN $Type Type ON Type.recordsId = Record.id
LEFT JOIN $Attribute Attribute ON Attribute.recordsId = Record.id
... conditions
AND
EXISTS (
  SELECT SuperRecord.*
  FROM $Record SuperRecord
  LEFT JOIN $Placement Placement ON Placement.setId = SuperRecord.id
... conditions
)
PHQL;
```

In my case I actually have two assigned relationships between `Placement` and `Record` (so another edge case) and I get this error message: `There is more than one relation between models 'Webird\Data\Record' and 'placement', the join must be done using an alias, when preparing: SELECT DISTINCT Record.*`.  Ok, but aliases are very broken when things start getting complicated (mostly I guess due to #10526) and the forced implicit joins.

I think a more comprehensive rework of the joining and aliases is necessary but for now I've created a more simple bug #10704 to let Phalcon get out of the way as people push the ORM into strange database specific use cases.

dschissler:
When I explicitly declared the join conditions inside the EXISTS it ended up listing it out **3** times.  Wow!  Something needs to get cleared out.

```php
...
EXISTS (
  SELECT `SuperRecord`.* FROM `record` AS `SuperRecord`
  LEFT JOIN `placement` AS `Placement` ON `Placement`.`setId` = `SuperRecord`.`id`
  LEFT JOIN `type` AS `Type` ON `Type`.`recordsId` = `SuperRecord`.`id`
  LEFT JOIN `attribute` AS `Attribute` ON `Attribute`.`recordsId` = `SuperRecord`.`id`
  LEFT JOIN `placement` AS `Placement` ON `Placement`.`setId` = `SuperRecord`.`id`
  LEFT JOIN `type` AS `Type` ON `Type`.`recordsId` = `SuperRecord`.`id`
[...]

Issue #10704, opened 2015-07-22
Feature Request: Allow implicit joins to be turned off because they block more complicated queries from being possible


dschissler:
Simply:

```php
  'no_implicit_joins_just_right_now_please' => YES
```

would be nice.  It would help to alleviate the immediate pains of bug #10526.

dschissler:
Due to the nature of the other bug this won't immediately help me but it might someday help some adapter pioneers out a lot.

dschissler:
This is my solution so far:

```
		if this->_explicit_joins_only == true {
			for joinItem in selectJoins {
				let joinAlias = joinItem["alias"]["name"];

				let joinType = joinTypes[joinAlias];
				let joinSource = joinSources[joinAlias];
				let sqlJoins[] = [
					"type": joinType,
					"source": joinSource,
					"conditions": [joinPreCondition[joinAlias]]
				];
			}
			return sqlJoins;
		}
```

andresgutierrez:
@dschissler where is that fragment of code located?

dschissler:
[Line 430](https://github.com/phalcon/cphalcon/blob/4651c14873fb903d495431835d3a2af196e35035/phalcon/mvc/model/query.zep#L1430).  I just short circuited the loop to avoid so much bracing and tab madness.

dschissler:
@andresgutierrez Here is my modified constructor:

```
	/**
	 * Phalcon\Mvc\Model\Query constructor
	 *
	 * @param string phql
	 * @param \Phalcon\DiInterface dependencyInjector
	 */
	public function __construct(phql = null, <DiInterface> dependencyInjector = null, options = null)
	{
		var explicit_joins_only;

		if typeof phql != "null" {
			let this->_phql = phql;
		}

		if typeof dependencyInjector == "object" {
			this->setDI(dependencyInjector);
		}

		let this->_explicit_joins_only = false;
		if typeof options == "array" {
			if fetch explicit_joins_only, options["explicit_joins_only"] {
				// let this->_explicit_joins_only = (explicit_joins_only == true);
				let this->_explicit_joins_only = true;
			}
		}
	}
```

andresgutierrez:
You can better take the global configuration variables approach:

* https://github.com/phalcon/cphalcon/blob/master/config.json#L84-L87
* https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/model/resultset/simple.zep#L111

dschissler:
[...]

Issue #10507, opened 2015-06-13
PHQL uses join ON 1 condition when joining directly against the hasMany linking table 


dschissler:
I have 3 table:
- `table1`: `id`
- `link_table`: `id`, `table1Id`, `table2Id`, `type`
- `table2`: `id`

The SQL reads `INNER JOIN table1 ON 1` instead of inferring the relationship from the hasManyToMany relationship.  I must include an additional hasMany relationship from table1 the linking table but this information is already available from the hasManyToMany relationship.

This stems from needing to include a flag field in the link table.  Phalcon seems to assume that the link hasManyToMany link table will only have a primary id and two foreign keys and just falls apart and offers no facilities for dealing with it any other way.

If there are no edge cases that would become ambiguous then I propose that in effect that a hasManyToMany relationship will automatically add an extra hasMany relationship to the linking table or some other other equivalent method that allows the `ON` condition to be filled out.

I did enjoy the new `$query->getSql()` method though and it made it real easy to catch once I knew what I was looking for.

dschissler:
@phalcon 

I created some scripts on [phalcon_bug_10507](https://github.com/dschissler/phalcon_bug_10507) that demonstrate the various errors.  Forgive the layout of the project because I needed to test different permutations of actual model classes and also how it interacts with settings on the ModelsManager.

One of the scripts works but it appears to be impossible to use the new ModelsManager aliases for the query that I want to use.  Using aliases appears to break the hasManyToMany searches when searching for values on the intermediate linking table.  I had to add an additional hasMany relationship to link them together.

So I think that maybe there are two hard bugs and another would-be-nice-to-have.

Its difficult to keep track of the bugs since sometimes there is a hard fail from it including the table JOIN condition twice when it is explicitly declared and other times the relationship table join `ON` condition is set to 1.  Its difficult to succinctly explain the variety of bugs and weird behaviour and so hopefully you can see that the queries shouldn't be anything too weird and that they just don't work unless setup in a very particular way.

dschissler:
[65199a71660ddc00aae8ee1b670c8ec22350ebaa](https://github.com/phalcon/cphalcon/commit/65199a71660ddc00aae8ee1b670c8ec22350ebaa) does not seem to fix the `sad_1.php`, `sad_2.php` or `sad_3.php` script.  New bug #10526 should resolve some of the tests/errors but then there is still the matter that if you add a relationship alias in `Persons` that it will fail.

[...]

PR #10721, opened 2015-07-25
 Implemented feature #10704, added ability to disable implicit joins for query.

dschissler:
I think that the impact on the code for stability concerns is very low because this feature is skipped over with the default settings.  The options name can be changed easily once this is accepted.

I also have a another fix for bug #10705 but I think that I can refactor the solution to better take advantage of calculations performed earlier in the routine.

So I see this as a two part fix with an iterative strategy;
1. To be able to bypass the problematic implicit join code as new issues come up.
2. To iterate on fixing bugs in the join code as they surface and thus being able to re-enable the implicit joining.

Its a terrible situation to hit a blocker bug when doing complicated PHQL queries (such as nested queries) and this will allow the developer to continue building their application in PHP land by performing explicit-only joins until such a time that he has time to fix the Phalcon bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # [2.0.7](https://github.com/phalcon/cphalcon/releases/tag/phalcon-v2.0.7) (2015-XX-XX)
-
+- Added orm.enable_implicit_joins option for allowing implicit joins to be disabled [#10704](https://github.com/phalcon/cphalcon/issues/10704)
+- Fixed bug causing SELECT subqueries to join against all of the tables from the previous query [#10705](https://github.com/phalcon/cphalcon/issues/10705)
 
 # [2.0.6](https://github.com/phalcon/cphalcon/releases/tag/phalcon-v2.0.6) (2015-07-21)
 - Builds in TravisCI now uses Docker to perform faster builds
--- a/config.json
+++ b/config.json
@@ -85,6 +85,10 @@
             "type": "bool",
             "default": false
         },
+        "orm.enable_implicit_joins": {
+            "type": "bool",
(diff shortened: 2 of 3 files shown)


dschissler:
@andresgutierrez I guess that I messed up with this.  I had intended for my two commits to be completely separate pull requests based upon the commit.  Let me know what you want to do with this?

andresgutierrez:
Could you please add unit-tests and update the CHANGELOG?

dschissler:
Honestly I don't know what the hell to test for and honestly I'm tired of learning new systems.  Now granted is Saturday night and I'm drunk so its possible that my opinion could change but its my understanding now that I'm already overextended my efforts to fix this issues and my project is long and hard.  I can do a changelog before the release but otherwise I'm tapped out.

I can probably manage to fit learning the test case stuff but I just don't see a win for me in the immediate days.  I'm leaning towards adding a test case by next minor version.  I don't see an issue with this because people apparently aren't using this feature ATM anyways.

dschissler:
As I'm considering it I'm definitely not interesting in spending significant more time on this issue in the immediate time frame.  I'm heavily invested in Phalcon but in the immediate frame I don't have much more to offer.

dschissler:
I think that I might need to extend some test models to be able to test for this.  I would like to fix bugs #10526 and #10507 in the future when I have time.

dschissler:
@andresgutierrez 

Should I create another pull request with just the fix?  Would you accept it without additional tests?

andresgutierrez:
I'll make the tests. Could you please rebase?

dschissler:
I'll do that later today.

PR #10722, opened 2015-07-25
Fix for bug #10705

dschissler:
I perceive this bug to be an artifact of Phalcon's humble beginnings when it did not support nested queries.

The select routine performs correctly for the outermost `SELECT` but then fails on join conditions within an `EXISTS` subquery.  This is due to the join conditions loop first iterating on a list of all of the tables so far on a global context (which is correct for the first query).  This patch hopes to fix this issue without causing regression by accessing the AST table data from that query only (down into the AST) instead of from the global list (of everything so far merged together).


dschissler:
I think that I pushed this against master.  I still learning the github interface.

PR #10723, opened 2015-07-25
Fixed wrong return types

greew:
As agreed on beforehand, when returning a namespaced PHP object, we need the backslash in front of the object.


andresgutierrez:
Thanks

PR #10751, opened 2015-08-02
[2.0.x] [NFR] \Mvc\Model\Criteria - fromImput #10749

### LINKED ISSUE #10749: New Feature - \Mvc\Model\Criteria - fromImput
https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/model/criteria.zep
#line 739

Some way to set it to OR, i know I can build one but would be nice being from core.

Thanks for building phalcon !

KorsaR-ZN:


ovr:
BC, it's need 2.1.x i think

andresgutierrez:
Thanks

PR #10757, opened 2015-08-04
__callStatic arguments should not be optional

Surt:
as php will send an empty array for arguments
It produces errors when using mock objects, etc..
(This is https://github.com/phalcon/cphalcon/pull/10565 but with __callStatic)


andresgutierrez:
Thanks

PR #10763, opened 2015-08-05
Fix nested SELECT bug #10705, implemented feature #10704 ability to disable implicit joins for query.

dschissler:
I was short on time and having trouble rebasing with two branches and so I started over with the latest 2.0.x code.

### DIFF
--- a/config.json
+++ b/config.json
@@ -85,6 +85,10 @@
             "type": "bool",
             "default": false
         },
+        "orm.enable_implicit_joins": {
+            "type": "bool",
+            "default": true
+        },
         "orm.cast_on_hydrate": {
             "type": "bool",
             "default": false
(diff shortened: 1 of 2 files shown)


dschissler:
I recommendation after working with this bit of code is to get it into 2.0.x towards the beginning of a minor cycle.  I'm not intimately familiar with the AST but it does pass current tests and the approach is much improved because it pulls table names from the nested query AST branch and not from the recorded global context.

My impression of the risk is that there might be some scenarios where the tables are coming from more places than the FROM section (as JOIN tables don't appear to be relevant for the outside loop).  Someone expertly versed in the AST and query should be able to figure out if that ever happens

In perhaps two months I'll set aside perhaps two weeks to learn a whole bunch of the latest testing systems in Javascript, PHP and Phalcon land.  Then I'll flesh out some of those other bugs.  I also plan on amending the testing schema a small amount to do this.  I figure that worst case there will be a few more (possibly disabled) tests for edge case that are not yet fixed.

andresgutierrez:
Thanks

dschissler:
More efforts to come later.

PR #10765, opened 2015-08-05
Update Model::getMessages description

sergeyklay:
Refs https://github.com/phalcon/zephir/issues/1044


SidRoberts:
It's also in the [interface](https://github.com/sergeyklay/cphalcon/blob/patch-4/phalcon/mvc/modelinterface.zep#L236). :smile:

PR #10786, opened 2015-08-13
reset readconnection in transaction

xueron:
if read and write connection are set separately and a transaction detected, use transaction's connection instead of predefined readConnection, for 'for update' statements


andresgutierrez:
Could you please submit only the commits that belong to this PR?

Remove this: https://github.com/xueron/cphalcon/commit/d52c4a5cd6cea82d87178d8e2024738eb40e8cd1
And this: https://github.com/xueron/cphalcon/commit/c03b2e6e75fcd8208a9e40d1ccdfadc741c0a3e6

PR #10790, opened 2015-08-13
Correctly assing keys and values to select field options

theDisco:
`Phalcon\Forms\Element\Select::addOption()` incorrectly assigned the keys and values creating optgroups instead of options.

### DIFF
--- a/phalcon/forms/element/select.zep
+++ b/phalcon/forms/element/select.zep
@@ -74,9 +74,13 @@ class Select extends Element implements ElementInterface {
 	 */
 	public function addOption(var option) -> <Element>
 	{
-		var key, value;
-		for key, value in option {
-			let this->_optionsValues[key] = value;
+		if is_array(option) {
+			var key, value;
+			for key, value in option {
+				let this->_optionsValues[key] = value;
+			}
+		} else {
+			let this->_optionsValues[] = option;
 		}
 		return this;
 	}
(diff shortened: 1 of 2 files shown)


phalcon/forms/element/select.zep
```diff
@@ -74,7 +74,10 @@ class Select extends Element implements ElementInterface {
 	 */
 	public function addOption(var option) -> <Element>
 	{
-		let this->_optionsValues[] = option;
```
sergeyklay: In this case

```zephir
addOption(array! option)
```

 ?

unit-tests/FormsTest.php
```diff
@@ -517,4 +517,12 @@ public function testIssue2045()
 
 		$this->assertEquals('<input type="text" name="name" class="big-input" />', $element->render());
 	}
+
+	public function testCorrectlyAddOptionToSelectElement()
+	{
+		$element = new Select('test-select');
+		$element->addOption(array('key' => 'value'));
```
sergeyklay:
```php
$element->addOption('foo-bar');
```

broke test

phalcon/forms/element/select.zep
theDisco: The question is what is the intended behavior? Doc block states an array and method signature doesn't enforce it. The behavior has changed between branch 1.x and 2.x although the signature didn't. I would personally go for an array cast, but since there are no test for it and if we want to maintain backwards compatibility, I need to change the implementation to support strings as well.

andresgutierrez:
Thanks

PR #10792, opened 2015-08-13
Cookie fix #10789

### LINKED ISSUE #10789: Cookies implementation
\Phalcon\Http\Response\Cookies implementation is wrong : 

- [x] it's coupled with \Phalcon\Cookie object without possibility to use DI
- [x] get return a Cookie Object (would be better to have a CookieInterface)
- [x] \Phalcon\Cookie is coupled with Session without reason, if session is not started it will break

ogarbe:
Update in Cookie and Cookies classes to fix #10789


sergeyklay:
Thanks

PR #10802, opened 2015-08-15
Add 'setParameter' to Phalcon\Di\ServiceInterface

ph55:
Since \Phalcon\DiInterface::getService returns \Phalcon\Di\ServiceInterface:

```
    /**
     * Returns the corresponding Phalcon\Di\Service instance for a service
     *
     * @param string $name 
     * @return \Phalcon\Di\ServiceInterface
     */
    public function getService($name);
```

\Phalcon\Di\ServiceInterface::setParameter should be added.


andresgutierrez:
Thanks

PR #10812, opened 2015-08-18
use transaction's connection as readconnection.

xueron:
if read and write connection are set separately and a transaction detected, use transaction's connection instead of predefined readConnection, for 'for update' statements

### DIFF
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -377,6 +377,13 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 	 */
 	public function getReadConnection() -> <AdapterInterface>
 	{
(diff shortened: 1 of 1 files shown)


xueron:
Now，I'm using this way, implemented in extended Model class 

```php
   public function selectReadConnection()
    {
        if ($this->_transaction != null) {
            return $this->_transaction->getConnection();
        }
        return $this->getReadConnection();
    }
```

xueron:
I think here _transaction may should be maintained in modelsManager, because if I start a transaction and want lock a row use Model::findFirst()  with for_update option, I cannot use setTransaction before call findFirst().

andresgutierrez:
Thanks

PR #10815, opened 2015-08-19
Added \Phalcon\Secutity\Random::base58

sergeyklay:
Added `\Phalcon\Secutity\Random::base58`.

From [Wikipedia](https://en.wikipedia.org/wiki/Base58):

> Base58 is a group of binary-to-text encoding schemes used to represent large integers as alphanumeric text. It is similar to Base64 but has been modified to avoid both non-alphanumeric characters and letters which might look ambiguous when printed. It is therefore designed for human users who manually enter the data, copying from some visual source, but also allows easy copy and paste because a double-click will usually select the whole string.

The result may contain alphanumeric characters except `0`, `O`, `I` and `l`. 

``` php
$random = new \Phalcon\Security\Random();

// Random base58 string
echo $random->base58();   // 4kUgL2pdQMSCQtjE
echo $random->base58();   // Umjxqf7ZPwh765yR
echo $random->base58(24); // qoXcgmw4A9dys26HaNEdCRj9
echo $random->base58(7);  // 774SJD3vgP
```

If `$len` is not specified, 16 is assumed. It may be larger in future. Throws `\Phalcon\Security\Exception` if secure random number generator is not available or unexpected partial read.

Refs #10804

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# [2.0.8](https://github.com/phalcon/cphalcon/releases/tag/phalcon-v2.0.8) (2015-XX-XX)
+- Added `Phalcon\Security\Random::base58` - to generate a random base58 string
+
 # [2.0.7](https://github.com/phalcon/cphalcon/releases/tag/phalcon-v2.0.7) (2015-08-17)
 - `Image\Adapter\Gd::save()` no longer fails if the method or the instance is created with a filename without an extension
 - Fixed segfault in `Image\Adapter\Imagick::text()`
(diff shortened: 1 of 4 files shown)


sergeyklay:
@andresgutierrez Improved performance.

```sh
php > $random = new \Phalcon\Security\Random();
php > $start = microtime(true); for ($i = 0; $i <= 1000000; $i++) { $random->base58(); if ($i % 100000 === 0) { echo microtime(true) - $start, PHP_EOL; } }; echo microtime(true) - $start, PHP_EOL;

0.00037980079650879
1.4900188446045
2.9223399162292
4.3424589633942
5.7653248310089
7.1917288303375
8.6174719333649
10.03345990181
11.453278779984
12.878019809723
14.295183897018
14.295220851898
```

sergeyklay:
```php
0.00033712387084961
1.2719330787659
2.5017211437225
3.751944065094
4.9992401599884
6.2343351840973
7.4692080020905
8.7010400295258
9.9340560436249
11.207904100418
12.466425180435
12.466461181641
```

now 100 000 ~ per 1.2sec

andresgutierrez:
Much better now!

PR #10826, opened 2015-08-24
Removes tests for deprecated registerPrefixes method

Green-Cat:
This is currently making the unit tests fail on the 2.1.x branch


Green-Cat:
Looks like there are some mvc\micro route issues too.

Issue #10819, opened 2015-08-21
[BUG][NFR] Phalcon2: Phalcon\Validation\Message is missing $code parameter


temuri416:
Hi,

Before:
https://docs.phalconphp.com/en/master/api/Phalcon_Validation_Message.html

Now:
https://docs.phalconphp.com/en/latest/api/Phalcon_Validation_Message.html

Phalcon 1.3 `Phalcon\Validation\Message` constructor had the fourth argument: `[int $code]`.

Can we have it back in Phalcon 2 along with getter and setter?

Thanks!

P.S. The same goes for `Phalcon\Mvc\Model\Message`.

temuri416:
Hi @andresgutierrez 

Was that omission a bug or a feature?

Cheers!

andresgutierrez:
It looks that no one is using it so it's probably just noticed

temuri416:
Can we have it back, please? I think it's incredibly useful to be able to assign codes to validation error messages.

Green-Cat:
@temuri416 I will implement it later today.
Also 'Phalcon\Mvc\Model\Validation' is deprecated in 2.1 so not much point fixing that one.

temuri416:
Glad to hear about the demise of `Phalcon\Mvc\Model\Validation` :+1: 

Thanks!

Green-Cat:
Done, #10827.

temuri416:
Much appreciated!

andresgutierrez:
Implemented in the 2.0.x branch

PR #10827, opened 2015-08-24
Adds the missing code property of validation\message

Green-Cat:
Adds the code property which was present in 1.3, but is currently missing in 2.0.x.
ref #10819


andresgutierrez:
Thanks

Issue #10448, opened 2015-05-31
migrations types problems


stefanchiriac:
Hello,

- when generating migrations is not finding the type 'TIMESTAMP', 'BIGINT', and other types
- there is no Column::TYPE_TIMESTAMP, but I got over it by just writing 'TIMESTAMP',
- default for TYPE_INTEGER does nothing if I set an int value
- when I add 'CURRENT_TIMESTAMP' as default value I get an mysql error. 

     `SQLSTATE[42000]: Syntax error or access violation: 1067 Invalid default value for 'created_at'`

Edit: Seems like if I don't set anything on default, is setting by itself CURRENT_TIMESTAMP and an attribute 'on update CURRENT_TIMESTAMP' which is not what I want at all.

It is a way to get over this, or to be implemented in 2.0.3?
Thanks

andresgutierrez:
Can you please post this in the devtools repo? github.com/phalcon/phalcon-devtools

stefanchiriac:
but this is part of cphalcon. Is not about phalcon-devtools, but that cphalcon doesn't have support integrated for these at all.

andresgutierrez:
This is fixed in the 2.0.4 branch

PR #10832, opened 2015-08-24
Added Column::TYPE_TIMESTAMP

### LINKED ISSUE #2850: migration problem with TIMESTAMP columns
The command `phalcon migration generate` mapped TIMESTAMP (MySQL) columns to `TYPE_DATE`.
`phalcon migration run` creates a new table with those columns as DATE columns.
http://docs.phalconphp.com/en/latest/api/Phalcon_Db_Column.html
I think `TYPE_TIMESTAMP` may be needed

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/4894281-migration-problem-with-timestamp-columns?utm_campaign=plugin&utm_content=tracker%2F50707&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F50707&utm_medium=issues&utm_source=github).
</bountysource-plugin>

sergeyklay:
Fixed #2850
Refs #10448

### DIFF
--- a/phalcon/db/adapter/pdo/mysql.zep
+++ b/phalcon/db/adapter/pdo/mysql.zep
@@ -173,6 +173,14 @@ class Mysql extends PdoAdapter implements AdapterInterface
 				}
 
 				/**
+				 * Timestamp are dates
+				 */
+				if memstr(columnType, "timestamp") {
+					let definition["type"] = Column::TYPE_TIMESTAMP;
+					break;
+				}
+
+				/**
 				 * Text are varchars
 				 */
 				if memstr(columnType, "text") {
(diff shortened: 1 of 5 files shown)


andresgutierrez:
And SQLite?

sergeyklay:
> SQLite does not have a storage class set aside for storing dates and/or times. Instead, the built-in Date And Time Functions of SQLite are capable of storing dates and times as TEXT, REAL, or INTEGER values:

> TEXT as ISO8601 strings ("YYYY-MM-DD HH:MM:SS.SSS").

For example we can't use Column::TYPE_DATETIME and SQLite, nevertheless, we have this column type

sergeyklay:
@andresgutierrez Added missed timestamp type to SQLite adapter

Green-Cat:
:+1:

andresgutierrez:
Thanks

PR #10836, opened 2015-08-25
Fix notice in _exists method of model

onshorechet:
In some cases fetchOne return false.  The _exists method should handle that condition and should be more explicit in what the return type from fetchOne.


sergeyklay:
We don't add changes directly to stable branches or master. Can you please do PR to 2.0.x branch?

PR #10837, opened 2015-08-25
Fix notice in _exists method of model

onshorechet:
In the _exists method of model the fetchOne call can return false.  That condition should not throw notices.  It should also be explicit to fetchOne for the return type.

### DIFF
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -1054,7 +1054,7 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 		 */
 		let num = connection->fetchOne(
 			"SELECT COUNT(*) \"rowcount\" FROM " . connection->escapeIdentifier(table) . " WHERE " . uniqueKey,
-			null,
+			\Phalcon\Db::FETCH_ASSOC,
 			uniqueParams,
 			uniqueTypes
 		);


andresgutierrez:
Could you please explain in which situation a COUNT(*) can return false?

chet0xhenry:
I'm not really sure, but is likely not the count however something going wrong in fetchOne.  I'm sorry i don't really have time to delve into fetchOne or fetchColumn.

We could short circuit this and call fetchColumn.
```zephir
let num = connection->fetchColumn(
    "SELECT COUNT(*) FROM " . connection->escapeIdentifier(table) . " WHERE " . uniqueKey,
    \Phalcon\Db::FETCH_ASSOC,
    uniqueParams,
    uniqueTypes
);
if num {
    let this->_dirtyState = self::DIRTY_STATE_PERSISTENT;
    return true;
} else {
    let this->_dirtyState = self::DIRTY_STATE_TRANSIENT;
}
```

chet0xhenry:
Yeah, nevermind about the fetchColumn it looks like it calls fetchOne. I seems like it should call fetchColumn on pdo directly but that is neither here nore there.

chet0xhenry:
I finally ran across the cause of this notice and the main problem  of the model re-inserting existing rows.

In applications that access the pdo connection directly it is possible to set the default return type from fetch.  If that is set to object this method will receive the count as an object.  Since php just throws a notice for using an object as an array and then uses null for the value the model will not get the correct count from the db and try to insert instead of update.

Passing the argument for return as associate array solves this problem.

It maybe better to make fetchColumn explicitly set the return type.  However if the application code uses these low level pdo abstration methods it is probably correct to use pdo's default and the framework should always pass the return type it is expecting.

Let me know if i need to rebase this on another branch and clean up the history.

chet0xhenry:
I just rebased this branch to clean up the git history.  Because I have found the root problem I was also able to minimize the diff to a single line.

chet0xhenry:
This is how one might setup a db connection that causes problems in the phalcon models:
```php
$connection = new \Phalcon\Db\Adapter\Pdo\Postgresql(
    array(
        'dsn'     => 'service=app;',
        'options' => [
            PDO::ATTR_DEFAULT_FETCH_MODE => PDO::FETCH_OBJ,
            PDO::ATTR_ERRMODE            => PDO::ERRMODE_EXCEPTION
        ]
    )
);
```

apriebe:
Can't wait for this to get pushed through!

andresgutierrez:
Thanks

PR #10844, opened 2015-08-27
fix transaction::isManagerd()

xueron:


andresgutierrez:
Could you please provide a use case where this patch is needed?

PR #10841, opened 2015-08-26
[2.1.x] Disable storing of cache keys per default. Performance regression

patrick-zippenfenig:
Disable options["statsKey"] in cache backends per default. It is required for flush() or queryKeys() which are probably rarely used in production.
It was also not possible to disable statKey in redis and xcache. Memcache issued some notices.

Mentioned previously in #10343, [forum 3310](https://forum.phalconphp.com/discussion/3310/how-to-disable-phcm-store-all-the-keys-when-using-libmemcached-)

Unit-tests may fail as I do not run all cache backend on my vagrant box

PR #10850, opened 2015-08-28
Restore setTrackingKey and getTrackingKey methods on Memcache and Libmemcached backends

jimjim2a:
In 1.3.x, setTrackingKey and getTrackingKey methods were available on Phalcon\Cache\Backend\Memcache and Phalcon\Cache\Backend\Libmemcached.

Restore also the 1.3.x behaviour of queryKeys() method : 
If there was no statsKey defined, queryKeys() method returns void instead of throwing an exception as in 2.0.x

### DIFF
--- a/phalcon/cache/backend/libmemcached.zep
+++ b/phalcon/cache/backend/libmemcached.zep
@@ -479,9 +479,11 @@ class Libmemcached extends Backend implements BackendInterface
 	 * Stores special memcached key use internally to store all memcache keys
 	 * @param string|null key
 	 */
-	public function setTrackingKey(var key) -> void
+	public function setTrackingKey(var key) -> <Libmemcached>
 	{
 		let this->_options["statsKey"] = key;
+
+		return this;
 	}
 
 	/**
--- a/phalcon/cache/backend/libmemcached.zep
+++ b/phalcon/cache/backend/libmemcached.zep
@@ -229,7 +229,7 @@ class Libmemcached extends Backend implements BackendInterface
 			throw new Exception("Unexpected inconsistency in options");
 		}
 
-		if typeof specialKey != "null" {
+		if specialKey != "" {
 			/**
 			 * Update the stats key
 			 */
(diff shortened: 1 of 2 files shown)


phalcon/cache/backend/libmemcached.zep
```diff
@@ -299,27 +299,27 @@ class Libmemcached extends Backend implements BackendInterface
 	 * Query the existing cached keys
 	 *
 	 * @param string prefix
-	 * @return array
+	 * @return array|void
 	 */
 	public function queryKeys(prefix = null)
```
sergeyklay: why not
```zephir
public function queryKeys(string prefix = null) -> array|void
```
without phpdoc

jimjim2a: Because Zephir throws a "Zephir\ParseException: Syntax" error when buiding Phalcon. (I use latest Zephir)

andresgutierrez:
@patrick-zippenfenig Is this compatible with your last contribution?

patrick-zippenfenig:
@andresgutierrez Almost. #10841 uses an empty string to disable stats-key features instead of NULL. Otherwise an empty string could be used as stats-key and raise an issue in memcache (invalid key). Otherwise setTrackingKey should be fine.

There is quite a lot a code duplication. I would create a base class with all stats-key features and inherit memcache, libmemcache, redis and xcache

jimjim2a:
Actually in 1.2.x statsKey tests were based on empty strings, and in 1.3.x were based on NULL. That's why I re-introduced the NULL version, to match the last behaviour.

I totally agree with @patrick-zippenfenig about the empty string key issue in memcache, if nobody else complains about it, I can update this PR and change the behaviour.

Also, in that case, setTrackingKey method could accept only string parameters.

I can do it if everyone is ok.

jimjim2a:
Any news about this ?

Is everyone ok if I update this PR with the behaviour we described ?

patrick-zippenfenig:
+1 for empty string to prevent memcached issues

jimjim2a:
PR updated

sergeyklay:
@andresgutierrez

jimjim2a:
ping

sergeyklay:
@jimjim2a Could you please rebase it onto 2.1.x? Also update please `CHANGELOG.md`

sergeyklay:
@jimjim2a Could you please rebase onto 3.2.x

sergeyklay:
I'm closing this PR due to the lack of any reaction. I'll open it again if the need arises

PR #10858, opened 2015-08-30
Fix bug with uppercase controller name

ovr:
If you register Route like this

``` php
$router->add('/reset-password/{code}/{email}/{id}/', array(
    'controller' => 'Auth',
    'action' => 'reset'
))
```

Application is working but it cant render View via it's Auth not auth

Not errors, no excepions...


ovr:
ping @andresgutierrez

PR #10859, opened 2015-08-30
Added UUID validator.

SidRoberts:


sergeyklay:
Hello @SidRoberts 

There are [5 versions](https://en.wikipedia.org/wiki/Universally_unique_identifier) of UUID. I've [introduced](https://github.com/phalcon/cphalcon/pull/10804) version 4 (random) of the UUID. I think you need to specify clearly which version (or all) of the UUID this validator covers.

Also please see UUID v4 validator: https://github.com/phalcon/cphalcon/pull/10814/files

sergeyklay:
@SidRoberts Could you please deal with this?

sergeyklay:
@SidRoberts ping

Issue #10866, opened 2015-08-31
[NFR] Dispatcher::getParam return defaultValue if sanitized length is 0


lajosbencz:
The current implementation of `Dispatcher::getParam` will only return the specified `defaultValue` if there is no such key in the parameters array.

If `defaultValue` argument is `null`, it should still return the sanitized value, even when empty.
If `defaultValue` is anything but `null` and the length of the sanitized parameter is 0, it should return the `defaultValue`.

PR #10867, opened 2015-08-31
[NFR] #10866

lajosbencz:
NFR https://github.com/phalcon/cphalcon/issues/10866


lajosbencz:
Disregard... sorry, branched from previous NFR commit

PR #10868, opened 2015-08-31
Implement and Test NFR #10866

lajosbencz:
NFR https://github.com/phalcon/cphalcon/issues/10866

### DIFF
--- a/phalcon/dispatcher.zep
+++ b/phalcon/dispatcher.zep
@@ -240,6 +240,17 @@ abstract class Dispatcher implements DispatcherInterface, InjectionAwareInterfac
 	}
 
 	/**
+	 * Determine if param exists by its name or numeric index
+	 *
+	 * @param  mixed param
+	 * @return bool
+	 */
+	public function hasParam(param) -> bool
+	{
+		return isset this->_params[param];
+	}
+
+	/**
 	 * Gets a param by its name or numeric index
 	 *
 	 * @param  mixed param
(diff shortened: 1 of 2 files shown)


andresgutierrez:
This looks too arbitrary, not sure if everyone wants to ignore empty strings

phalcon/dispatcher.zep
```diff
@@ -265,7 +265,11 @@ abstract class Dispatcher implements DispatcherInterface, InjectionAwareInterfac
 			this->{"_throwDispatchException"}("A dependency injection object is required to access the 'filter' service", self::EXCEPTION_NO_DI);
 		}
 		let filter = <FilterInterface> dependencyInjector->getShared("filter");
-		return filter->sanitize(paramValue, filters);
+		let paramValue = filter->sanitize(paramValue, filters);
+		if strlen(paramValue) < 1 && defaultValue !== null {
+			return defaultValue;
```
sergeyklay: Mb 
```zephir
return filter->sanitize(defaultValue, filters);
```
?

lajosbencz: Makes sense!

lajosbencz:
Imho, I can't really think of a useful case when a 0 length dispatcher param is useful after sanitation. One will most likely handle it somehow, or just skip a prog branch. I think it's counter-intuitive if one still gets an empty string if a defaultValue is provided.
If only sanitation is the objective, defaultValue can be left empty.
I'm not keen on this implementation, because I don't need it in any project atm, it just makes sense this way :)

lajosbencz:
On second thought, adding a ```hasParam``` method would also allow for more flexibility without any impairment

PR #10872, opened 2015-09-01
Add "onConstruct" hook on \Phalcon\Cli\Task

jimjim2a:
Same hook as Phalcon\Mvc\Controller. Useful as __construct() method is final (it was not the case in 1.3.x)


andresgutierrez:
Thanks

sergeyklay:
:+1:

PR #10877, opened 2015-09-02
This adapter store sessions in Mongo

danclive:


andresgutierrez:
Could you please add docblocks to each method and class?

Example: https://github.com/phalcon/cphalcon/blob/2.0.x/phalcon/session/adapter/memcache.zep

Could you please add unit-tests? https://github.com/phalcon/cphalcon/tree/2.0.x/tests/PhalconTest/Session/Adapter

PR #10880, opened 2015-09-02
NPE in Phalcon\Form\Element::appendMessage - missing statement

olivier-monaco:
There was a missing statement in appendMessage of Phalcon\Form\Element. Before this patch, the unit test I added:

```
There was 1 error:

1) FormsTest::testElementAppendMessage
RuntimeException: Trying to call method appendmessage on a non-object

.../cphalcon-olivier/unit-tests/FormsTest.php:541
```

With this patch, the unit test is working.


sergeyklay:
:+1:

andresgutierrez:
Thanks

PR #10894, opened 2015-09-05
Simple credit card validation using luhn algorithm

caiofralmeida:


caiofralmeida:
correct array syntax to support PHP 5.3

PR #10900, opened 2015-09-06
[2.1.x] Added internal check before calling a validator

KorsaR-ZN:


phalcon/validation.zep
```diff
@@ -459,4 +466,20 @@ class Validation extends Injectable implements ValidationInterface
 
 		return value;
 	}
+
+	/**
+	 * Internal validations, if it returns true, then skip the current validator
+	 */
+	protected function preChecking(string field, <ValidatorInterface> validator) -> boolean
+	{
+		if validator->getOption("allowEmpty", false) === true {
```
sergeyklay: This means that the developer will not be able to use
```php
'allowEmpty' => 1
```

KorsaR-ZN: @sergeyklay fixed it

andresgutierrez:
Could you please rebase?

KorsaR-ZN:
@andresgutierrez rebased

andresgutierrez:
Thanks

Issue #2411, opened 2014-05-09
[bug] Phalcon\Paginator\Adapter\QueryBuilder count is not accurate


zhouyl:
(Sorry, i am chinese! Too bad English.)

When QueryBuider presence group by, count is not accurate.

My Code

```php
class MyQueryBuilder extends \Phalcon\Paginator\Adapter\QueryBuilder
{

    public function getPaginate()
    {
        $paginate = parent::getPaginate();

        $builder = $this->getQueryBuilder();

        if ($group = $builder->getGroupBy()) {
            $result = $builder->columns('count(distinct ' . $group . ') as total')
                ->groupBy(null)
                ->getQuery()
                ->execute()
                ->getFirst();
            $paginate->total_items = $result->total;
            $paginate->total_pages = ceil($paginate->total_items / $this->getLimit());
        }

        return $paginate;
    }

}
```

braf:
Very sorry about referencing this issue a billion times, I was committing from a vagrant box and the author kept coming out wrong and I was fixing it with some rebasing and didn't realise it would do this from my own fork. :cry:

andresgutierrez:
@braf :+1: no problem

Green-Cat:
@sergeyklay Fixed in #11195

PR #10903, opened 2015-09-07
Fix for incorrect page counts when using group by with the query builder pagination

braf:
This is a temporary fix for issue #2411. This is where GROUP BY clauses cause an incorrect total item count and thus page count for the QueryBuilder

I appreciate that this is not the perfect fix as it doesn't use a count and a FROM subquery like it has in previous versions. However this is not a function that is supported by PHQL at this time so this fix was not possible.

Another fix I came up with involved moving the grouping into a COUNT(DISTINCT col) but it doesn't work with multiple group by clauses and is notoriously slow in older versions of MySQL.

I've opted for this solution as it doesn't really change the function that much but does correct the issue.

### DIFF
--- a/phalcon/paginator/adapter/querybuilder.zep
+++ b/phalcon/paginator/adapter/querybuilder.zep
@@ -175,10 +175,21 @@ class QueryBuilder extends Adapter implements AdapterInterface
 		/**
 		 * Obtain the result of the total query
 		 */
-		let result = totalQuery->execute(),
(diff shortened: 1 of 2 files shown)


sergeyklay:
@braf Could you please rebase onto 3.2.x branch

niden:
Closing this because it has been addressed in the code with tests and changes. Thank you @braf  and apologies for taking this long to check this out. Really appreciate the work.

PR #10913, opened 2015-09-09
Adds mixed return type to session adapter get method

Green-Cat:
This is so that php IDEs do not see the get method as returning void.
screenshot_2

### DIFF
--- a/phalcon/session/adapter.zep
+++ b/phalcon/session/adapter.zep
@@ -126,7 +126,7 @@ abstract class Adapter
 	 *	$session->get('auth', 'yes');
 	 *</code>
 	 */
-	public function get(string index, var defaultValue = null, boolean remove = false)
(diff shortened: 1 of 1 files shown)


sergeyklay:
This will be 
```php
 * @return var
```

Unfortunately currently, Zephyr does not covert `var` to `mixed`

Green-Cat:
I have tested and the stub generator makes it into 
```php
     * @return variable 
```
which is wrong. I will see if I can make a fix PR for that.

Green-Cat:
This is actually already fixed in the master branch of zephir https://github.com/phalcon/zephir/blob/master/Library/Stubs/MethodDocBlock.php#L71
The stubs just need to be regenerated.

sergeyklay:
Great! :+1:

Green-Cat:
@andresgutierrez

andresgutierrez:
Thanks

Issue #10912, opened 2015-09-09
Reusable Micro Middleware (before) does not stop operation?


sszdh:
Hi there,

I have an Access class as a Middleware for my Micro App:

``` php
namespace Api\Middleware;

use Phalcon\Mvc\Micro as PhMicro,
    Phalcon\Mvc\Micro\MiddlewareInterface;

/**
 * Class Access
 * @package Api\Middleware
 */
class Access implements MiddlewareInterface
{
    public function call(PhMicro $application)
    {
        $session  = $application['session'];

        if(!$application->session->has('identity')) {
            $application->response->unAuthorized(); // Response json error blah blah
            return false;
        }
        return true;
    }
}
```

but when it attached to application as follow:

``` php
$application = new \Phalcon\Mvc\Micro();
$application->before(new Api\Middleware\Access());
$application->handle();
```

The operation does not stop and route execute after middleware returns FALSE!

I confused cause using middleware as a closure already works:

``` php
$application = new \Phalcon\Mvc\Micro();
$application->before(function () use($application) {
    if(!$application->session->has('identity')) {
        $application->response->unAuthorized();
        return false;
    }
    return true;
});
$application->handle();
```

Is this a bug or something else?

Thanks.

pauliuspetronis:
I think it's a bug.

By [documentation](https://docs.phalconphp.com/en/latest/reference/micro.html#middleware-events): 

Event Name | Triggered | Can stop operation?
------------ | ------------- | -------------
before | Before executing the handler. It can be used to control the access to the application | *Yes*
after | Executed after the handler is executed. It can be used to prepare the response | *No*
finish | Executed after sending the response. It can be used to perform clean-up | *No*

By [source code](https://github.com/phalcon/cphalcon/blob/2.0.x/phalcon/mvc/micro.zep#L640-L687), last *if statement* will never be executed.

I created Pull request with fix(in 2.0.x)

pauliuspetronis:
@sszdh, you must call `$application->stop()` before `return false` (in your Middleware class)

sszdh:
@pauliuspetronis, thanks
Yes I think it's a bug!

The '$status' tied with '$_stopped' property and I do not understand that...

steambao:
Yep, can confirm bug. Been waiting for this to be fixed for a while actually :)

hinablue:
Every `handle` called will reset the `this->_stopped` to `false`, so middleware will never stop. If use the `this->_stopped` to check.

[...]

PR #10923, opened 2015-09-10
Bugfix #10912 (Micro Middleware (before) does not stop operation)

pauliuspetronis:
In this pull request is one important change.

before this change:
1. if `event` (before) is closure and return false -  [cancels the route execution](https://github.com/phalcon/cphalcon/blob/2.0.x/phalcon/mvc/micro.zep#L676-L678)
2. if `event` (before) is `MiddlewareInterface` object and return false/true/... - not cancels route execution

In this change:
1. if `event` (before) is closure and return false/true/... - **not** cancels route execution
2. if `event` (before) is `MiddlewareInterface` and return false/true/... - not cancels route execution

If you need to cancel execution on `event` (before) - in both cases **must be** called `$microApplication->stop();`

For example:

``` php
<?php
$app->before(function () use ($app) {
    if ($app['session']->get('auth') == false) {
        $app->stop();
        return false;
    }
    return true;
});
$app->handle();
```

or

``` php
<?php
class Auth implements Phalcon\Mvc\Micro\MiddlewareInterface
{
    public function call(Phalcon\Mvc\Micro $app)
    {
         if ($app['session']->get('auth') == false) {
            $app->stop();
            return false;
        }
        return true;
    }
}
$app->before(new Access());
$app->handle();
```

### DIFF
--- a/phalcon/mvc/micro.zep
+++ b/phalcon/mvc/micro.zep
@@ -671,18 +671,22 @@ class Micro extends Injectable implements \ArrayAccess
 						}
 
 						/**
-						 * Call the before handler, if it returns false exit
+						 * Call the before handler
 						 */
-						if call_user_func(before) === false {
-							return false;
-						}
+						let status = call_user_func(before);
 
 						/**
-						 * Reload the 'stopped' status
+						 * break the execution if the middleware was stopped
 						 */
-						if this->_stopped {
-							return status;
+						if  this->_stopped {
+							break;
 						}
+					}
+					/**
+					 * Reload the 'stopped' status
+					 */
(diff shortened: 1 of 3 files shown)


andresgutierrez:
I think this must be submitted to 2.1.x

sszdh:
So what is `return false/true` for, anymore?

Why we should use `$app->stop()`

Thanks

pauliuspetronis:
@andresgutierrez - submitted (#10931)

JuliusKoronci:
This is still there even with version 3+ ..no matter if your return true, false or call $app->stop() it will not top with the MiddlewareInterface

PR #10931, opened 2015-09-11
Bugfix #10912 (Micro Middleware (before) does not stop operation)

pauliuspetronis:
In this pull request is one important change.

before this change:
1. if `event` (before) is closure and return false -  [cancels the route execution](https://github.com/phalcon/cphalcon/blob/2.0.x/phalcon/mvc/micro.zep#L676-L678)
2. if `event` (before) is `MiddlewareInterface` object and return false/true/... - not cancels route execution

In this change:
1. if `event` (before) is closure and return false/true/... - **not** cancels route execution
2. if `event` (before) is `MiddlewareInterface` and return false/true/... - not cancels route execution

If you need to cancel execution on `event` (before) - in both cases **must be** called `$microApplication->stop();`

For example:

``` php
<?php
$app->before(function () use ($app) {
    if ($app['session']->get('auth') == false) {
        $app->stop();
        return false;
    }
    return true;
});
$app->handle();
```

or

``` php
<?php
class Auth implements Phalcon\Mvc\Micro\MiddlewareInterface
{
    public function call(Phalcon\Mvc\Micro $app)
    {
         if ($app['session']->get('auth') == false) {
            $app->stop();
            return false;
        }
        return true;
    }
}
$app->before(new Access());
$app->handle();
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 - Fixed `Phalcon\Mvc\Micro::handle` to prevent attemps to send response twice [#12668](https://github.com/phalcon/cphalcon/pull/12668)
 - Fixed `Di::set`, `Di::setShared` to allow pass more than 10 arguments [#12299](https://github.com/phalcon/cphalcon/issues/12299)
 - Fixed `Phalcon\Mvc\Model\MetaData\Strategy\Annotations::getColumnMaps` where only renamed columns where returned if there was one
+- Fixed `Phalcon\Mvc\Micro:handle` to correctly handle `before` handlers [#10931](https://github.com/phalcon/cphalcon/pull/10931)
 
 # [3.1.2](https://github.com/phalcon/cphalcon/releases/tag/v3.1.2) (2017-04-05)
 - Fixed PHP 7.1 issues [#12055](https://github.com/phalcon/cphalcon/issues/12055)
--- a/phalcon/mvc/micro.zep
+++ b/phalcon/mvc/micro.zep
@@ -677,18 +677,22 @@ class Micro extends Injectable implements \ArrayAccess
 						}
 
 						/**
-						 * Call the before handler, if it returns false exit
+						 * Call the before handler
 						 */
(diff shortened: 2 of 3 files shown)


andresgutierrez:
Just to make sure, return false does not stop the events?

pauliuspetronis:
Yes.
In `before` event -  will stop only if you call `stop()` function
In `after` and `finish` events - will never cancel process, but if `event` call `stop()` then skip next same type events (`after`, or `finish`)

sszdh:
:+1:

sergeyklay:
:+1:

sergeyklay:
@pauliuspetronis Could you please rebase onto 3.2.x branch

pauliuspetronis:
@sergeyklay - done

sergeyklay (approved):
Looks good to me

sergeyklay:
@pauliuspetronis Could you please update CHANGELOG.md

pauliuspetronis:
@sergeyklay - updated. But I think the same problem is with [afterBindingHandlers](https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/micro.zep#L791). Or I'm wrong?

sergeyklay:
@pauliuspetronis Looks like yes. Anyway it should be in separated PR. Thank you for contributing !

necipallef:
First of all, I would like to say thanks for this awesome framework and awesome work. Really appreciate it. We love using PHP and Phalcon in our projects, we've already written several separate libraries to make the best use of Phalcon [(here is an example)](https://github.com/Fabsolute/Phalcon-REST-API).

However, by changing a crucial functionality like this (calling return false does not stop the execution on before) broke most of our applications that are live today, causing a huge security issues. It is unpleasant to be aware of this after it is been a while since this change was released. 

Maybe I have to check here more often, but we are already dependent on lots of other libraries too, and checking them all is not always possible. Please be more careful not to break people's code who loves using Phalcon in the future.

PR #10937, opened 2015-09-11
Fix #10871 Updated method Phalcon\Http\Cookie::delete

### LINKED ISSUE #10871: [BUG] Cookie delete, subdomain
If I have `test.example.com` and writing cookies to `.example.com`. 

``` php
/**
 * Set the cookie from test.example.com to .example.com
 */
$this->cookies->set('remember-me', 'some value', 0, '/', null, '.example.com');
```

Later I can't delete the cookie:

``` php
/**
 * Get the cookie
 */
$this->cookies->get('remember-me'); // it's ok

/**
 * Delete the cookie
 */
$this->cookies->get('remember-me')->delete(); // does nothing
// or
$this->cookies->delete('remember-me'); // does nothing

```

dreamsxin:


andresgutierrez:
Thanks

Issue #10343, opened 2015-05-13
Libmemcached cache\backend performance regression


patrick-zippenfenig:
Hi!

tl;dr: please disable "statKey" (storing ALL keys) per default

I just spend 6 hours debugging a performance regression in memcache. One of three memcached servers updated keys at to up 130Mbit/s. In the last couple of days totals >5TB. Network bandwidth showed an odd pattern:
![unknown-3](https://cloud.githubusercontent.com/assets/11670456/7615862/abea6010-f9a0-11e4-9ec1-7ed83f21a8f6.jpeg)

I suspected the issue anywhere but in phalcon as it is a quite complex webapplication with a lot of objects in memcached. I started logging cache set/get's, changed cache keys to prevent possible uneven distribution, binary protocol, compression, implemented persistent memcached connections, analysed access-logs, etc.

After tcp_dump'ing memcache packets and analysing traffic flow I recognised "_PHCM" was updated very frequently. A quick look at [libmemcached.zep#L87](https://github.com/phalcon/cphalcon/blob/master/phalcon/cache/backend/libmemcached.zep#L87) reveals: "statKey" feature is enabled per default. Every updated key is stored as an entry in memcached to enable ````$cache->flush()```` capabilities. ````"statKey" => ""```` solved the issue, dropped load by half, freed a few GB of memory (memcache+php) and reduced network bandwidth to a regular level.
Phalcon queried "_PHCM" for each ````$cache->save()````, added the modified key and stored it again in "_PHCM". After some time time an internal limit was hit, crashed and it started all over again (see bandwidth graph).

I doubt that any application use ````flush()```` in production. I would like to request to disable this behaviour per default in all cache backends. An exception is already thrown in ````flush()```` if "statKey" is disabled.

Thanks!

andresgutierrez:
+1 for disabling statKey by default

kimthangatm:
Hi, I use Phalcon v2.0, Code `$cache->flush()` not working.

dugwood:
@patrick-zippenfenig : I fixed it a while ago (https://github.com/phalcon/cphalcon/pull/3183) but nobody seemed to be ashamed of such an «issue»

@kimthangatm : flush() won't work, as it just empty all the known cached entries, but keeps the list. My suggestion: add a parameter in the prefix of Memcached parameters (Memcached::OPT_PREFIX_KEY), and update the prefix when you want all your keys to be disabled. That won't remove them, but after the timeout these could be removed automatically.

[...]

PR #10939, opened 2015-09-12
[2.1.x] Memcached persistent connection pool for cache and session backends.

patrick-zippenfenig:
Memcached persistent connection pool for cache and session backends. Enabled per default. Cache and Session can share the same connection pool as long as `$server` and `$client` are equal.

``` php
$cache = new \Phalcon\Cache\Backend\Libmemcached($frontCache, [ 
  "server" => $servers,
  "persistent_id" => "my_pool",
  "client" => $client_settings
]);
$session = new Phalcon\Session\Adapter\Libmemcached([ 
  "lifetime" => 3600,
  "server" => $servers,
  "persistent_id" => "my_pool",
  "client" => $client_settings
]);
```

2x Unit-Tests updated accordingly to 'purge' client settings.
I've been using persistent memcached pool since my statsKey performance regression #10343.

### DIFF
--- a/phalcon/cache/backend/libmemcached.zep
+++ b/phalcon/cache/backend/libmemcached.zep
@@ -57,36 +57,6 @@ use Phalcon\Cache\Exception;
  * $data = $cache->get('my-data');
  *
  *</code>
- *
- * For multiple memcached servers please make sure to enable proper fail-over and timeouts.
- * The following settings could be used in a fast local network environment.
- * Shared environments with higher latencies require different settings.
- *
- *<code>
- *$cache = new \Phalcon\Cache\Backend\Libmemcached($frontCache, [
- *  "server" => $servers,
- *  "persistent_id" => "phalcon_cache",
- *  "client" => [
- *    // Ensure correct mapping of keys for failed servers
- *    Memcached::OPT_DISTRIBUTION => Memcached::DISTRIBUTION_CONSISTENT,
- *    // Remove failed servers from connection pool
- *    Memcached::OPT_REMOVE_FAILED_SERVERS => true,
- *    // Number of connection failures after a server is considered down
- *    Memcached::OPT_SERVER_FAILURE_LIMIT => 2,
- *    // Retry failed servers every 1 second
- *    Memcached::OPT_RETRY_TIMEOUT => 1,
- *    // Can not initially connect within 20ms -> consider offline
- *    Memcached::OPT_POLL_TIMEOUT => 20,
- *    // An open connection does not respond to polling within 20ms -> consider offline
- *    Memcached::OPT_CONNECT_TIMEOUT => 20,
- *    // Binary transfer is marginally faster than ascii transfer
- *    Memcached::OPT_BINARY_PROTOCOL => true,
- *    // Asynchronous I/O is supposedly the fastest transport
- *    Memcached::OPT_NO_BLOCK => true,
- *    // Don't queue network packages. Send as fast as possible
- *    Memcached::OPT_TCP_NODELAY => true
- *]]);
- *</code>
  */
 class Libmemcached extends Backend implements BackendInterface
 {
--- a/phalcon/cache/backend/libmemcached.zep
+++ b/phalcon/cache/backend/libmemcached.zep
@@ -27,9 +27,8 @@ use Phalcon\Cache\Exception;
 /**
  * Phalcon\Cache\Backend\Libmemcached
  *
- * Allows to cache output fragments, PHP data or raw data to a libmemcached backend
- *
- * This adapter uses the special memcached key "_PHCM" to store all the keys internally used by the adapter
+ * Allows to cache output fragments, PHP data or raw data to a libmemcached backend.
+ * Per default persistent memcached connection pools are used.
  *
  *<code>
  *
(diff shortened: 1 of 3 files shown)


patrick-zippenfenig:
````Zephir\CompilerException: Class 'Memcached' does not implement method: 'getServerList'```` works on my vagrant box...
@andresgutierrez [prototypes/memcached.php](https://github.com/phalcon/zephir/blob/master/prototypes/memcached.php) is missing multiple funtions. Are there any generator scripts to create prototypes automatically?

andresgutierrez:
No, those files are maintained manually

andresgutierrez:
Load the extension can solve the problem

dugwood:
@patrick-zippenfenig seems good to me, except the distribution setting. Hopefully, you stated it as a comment, but DISTRIBUTION_CONSISTENT doesn't handle correctly failed servers: instead of distributing again the keys of the failed server, all get/set against this server will fail. This implies a very bad behaviour, especially with few Memcached servers, as many get/set will fail, adding a lot of stress on the database (I think you use Memcached to remove useless database requests).
I don't have anything else to add, as I don't use the latter two parameters (maybe I should, but I don't have time to test these for now).

Thanks for you work :-)

patrick-zippenfenig:
@dugwood Would it be better to recommend OPT_LIBKETAMA_COMPATIBLE? I though DISTRIBUTION_CONSISTENT handles failed servers correctly, but never really investigated. Memcached is a weird creature ;-)

@andresgutierrez What do you mean by loading the extension? It working fine on my box, but fails in travis despite the extension is loaded. Otherwise i'd open a PR and add the missing prototype function in zephir.

dugwood:
@patrick-zippenfenig : my bad :-) The issue was with OPT_LIBKETAMA_COMPATIBLE, not with DISTRIBUTION_CONSISTENT. So stick with your current settings. I've tried LIBKETAMA, with no real advantage (I don't mix libmemcached versions and/or clients).
If a node is down, the consistent distribution will spread the «dead» keys on the remaning servers, avoiding the risk to overload a node (say if servers A/B/C, and B is down, keys won't go only on C as in a circle implementation).

andresgutierrez:
@patrick-zippenfenig The zephir prototype is only loaded in case the extension is not available. Maybe the version provided by Travis is outdated and doesn't have that method.

patrick-zippenfenig:
````getServerList```` is now in zephir memcached prototype along other missing function.

I removed the memcached options doc block. I experienced some issues with ````Memcached::OPT_REMOVE_FAILED_SERVERS => true````. Servers were not re-added. Maybe I'll setup some testing VMs in the next couple of days

dugwood:
Thanks @patrick-zippenfenig. For the `OPT_REMOVE_FAILED_SERVERS`, was it during the same test (you run the script only once), or was it the next time you run the script? The goal of this parameter is to remove the server during the current PHP run, so that the faulty server isn't hit anymore. But on the next run, you should have it back. It is removed after `OPT_SERVER_FAILURE_LIMIT` failed calls (defaut is 5 IIRC).

patrick-zippenfenig:
@dugwood In production ````OPT_REMOVE_FAILED_SERVERS```` began to remove memcached servers randomly and did not restore them. I tried various settings (even with ````OPT_LIBKETAMA_COMPATIBLE````,) but had to disable ````OPT_REMOVE_FAILED_SERVERS ````. I did not collect additional information. I'd have to setup a few VMs to run further tests.
As far as I know ````OPT_REMOVE_FAILED_SERVERS```` removes a connection after ````OPT_SERVER_FAILURE_LIMIT```` failures from the current persistent pool and is supposed to re-enable the failed server after ````OPT_RETRY_TIMEOUT````.

dugwood:
Yes, that's how I see it too. I thought I add tested it this way: shutdown a server, try to set data on it too many times, see the fail status, then enable the server again and wait (still in the same script) at least `OPT_RETRY_TIMEOUT` and tried again to set a variable: seemed to work.
How do you know that the servers were not restored? As it depends from the PHP pool you've set up (the info is on the Memcached client, not on the Memcached server). For example, in FastCGI PHP, I don't think the info lasts longer than the script run. On PHP-FPM, it may be kept by the FPM thread.

patrick-zippenfenig:
@dugwood I only use PHP-FPM (FastCGI Process Manager) with persistent instances. ````Memcached::OPT_SERVER_FAILURE_LIMIT => 1```` solved my issue. A server is marked immediately down and I don't receive an exception ````Failed storing data in memcached```` and failed servers were successfully reinstated. No idea why a value of ````2```` prevented this behaviour.

@andresgutierrez The PR should be fine and only enables persistent libmemcached connections and applies configuration before adding servers. Zephir now includes all /Memcached functions.

As reference my memcached client settings: (no included in the docs anymore)
````php
// Ensure correct mapping of keys for failed servers
Memcached::OPT_DISTRIBUTION => Memcached::DISTRIBUTION_CONSISTENT,
Memcached::OPT_LIBKETAMA_COMPATIBLE => true,
// Remove failed servers from connection pool
Memcached::OPT_REMOVE_FAILED_SERVERS => true,
// Number of connection failures after a server is considered down. Only a value of 1 seems to work correctly
Memcached::OPT_SERVER_FAILURE_LIMIT => 1,
// Retry failed servers every 1 second
Memcached::OPT_RETRY_TIMEOUT => 1,
// Can not initially connect within 20ms -> consider offline
Memcached::OPT_POLL_TIMEOUT => 20,
// An open connection does not respond to polling within 20ms -> consider offline
Memcached::OPT_CONNECT_TIMEOUT => 20,
// Binary transfer is marginally faster than ascii transfer
Memcached::OPT_BINARY_PROTOCOL => true,
// Asynchronous I/O. Supposedly fastest transport
Memcached::OPT_NO_BLOCK => true,
// Don't queue network packages. Send as fast as possible
Memcached::OPT_TCP_NODELAY => true
````

andresgutierrez:
@dugwood @patrick-zippenfenig Thanks

patrick-zippenfenig:
Follow up:
After some high availability issues with memcached I run some further tests:
- ````REMOVE_FAILED_SERVERS_LIMIT = 1 or 0```` immediately removes a memcache server from the pool and it won't be checked again. PHP restart is required to add it again. Undesired behaviour -> inconsistent data!
- ````REMOVE_FAILED_SERVERS / (_LIMIT)```` should be very big to continuously check a failed server
- On server failure, memcache will return false or raise an error on each function call to store, get, delete, exists etc.
- It is necessary to set ````Memcached::OPT_LIBKETAMA_COMPATIBLE```` again to [temporarily change key distribution and disable the failed server](http://hoborglabs.com/en/blog/2013/memcached-php)
- A failed server will be added to the pool after ````OPT_RETRY_TIMEOUT```` (even if still offline) and the next memcache function call will raise an error again
- BINARY_PROTOCOL prevents ````OPT_RETRY_TIMEOUT```` from [working on load](https://bugs.launchpad.net/libmemcached/+bug/1135781)

A silent fail-over would require: (+ return code checking and retry loops)
````php
$client_options = [
    Memcached::OPT_DISTRIBUTION => Memcached::DISTRIBUTION_CONSISTENT,
    Memcached::OPT_LIBKETAMA_COMPATIBLE => true,
    Memcached::OPT_REMOVE_FAILED_SERVERS => PHP_INT_MAX,
    Memcached::OPT_BINARY_PROTOCOL => false
];

$success = $memcache->set($lastKey, $cachedContent, $tt1);
if (!$success) {
    $memcache->setOption(\Memcached::OPT_LIBKETAMA_COMPATIBLE, true);
    $memcache->set($lastKey, $cachedContent, $tt1);
}
````
I continue testing and might create a new PR

PR #10944, opened 2015-09-14
Add 'index' option for Redis

sukobuto:
Add 'index' option for select Redis database.
Metadata and Session.


sergeyklay:
1. We don't add changes directly to stable branches or master
2. The `2.1.x` is best place for this
3. As you can see tests fails

Issue #10953, opened 2015-09-17
Issue in Phalcon\Forms\Element::appendMessage()


joao-pinheiro:
Hi,
It seems there is an issue in the appendMessage() method. When this->_messages is not initialized, the new Group is created in this->_messages but the local messages variable is not updated. 

code near https://github.com/phalcon/cphalcon/blob/master/phalcon/forms/element.zep#L537 :
        let messages = this->_messages;
        if typeof messages != "object" {
            let this->_messages = new Group();
        }
        messages->appendMessage(message);

in PHP:
$this->get('elementName')->appendMessage(...) throws exception, but doing $this->get('elementName')->setMessages(new Group()); $this->get('elementName')->appendMessage(...) seems to work.

Can anyone please confirm if this is actually an issue?

PR #10954, opened 2015-09-17
Messages in Forms\Element are now always available (#10953)

SidRoberts:
Form Messages are now initialized in the constructor. #10953

### DIFF
--- a/phalcon/forms/element.zep
+++ b/phalcon/forms/element.zep
@@ -64,6 +64,7 @@ abstract class Element implements ElementInterface
 		if typeof attributes == "array" {
 			let this->_attributes = attributes;
 		}
+		let this->_messages = new Group();
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


andresgutierrez:
Thanks

phalcon/forms/element.zep
```diff
@@ -484,16 +485,7 @@ abstract class Element implements ElementInterface
...
-
-		let messages = this->_messages;
-		if typeof messages == "object" {
-			return messages;
-		}
-
-		let messages = new Group(),
-			this->_messages = messages;
-		return messages;
+		return this->_messages;
```
sergeyklay: why not
```zephir
protected _messages {get};
```

SidRoberts: Because I forgot. :stuck_out_tongue:

PR #10962, opened 2015-09-21
according to PSR, return camelCase style property

r3jack:


sergeyklay:

r3jack:
sorry, i will re-request to 2.0.x

PR #10964, opened 2015-09-22
Adding beforeException on Application. implement _handleException on getModule

Surt:
Actually it's is not possible to handle the module not defined exception except for a catch on application launch. 
The new _handleException is a copy from dispatcher. If an exception is needed the application will launch an event: application:beforeException where it will be possible to override the default behaviour (throw the exception)

### DIFF
--- a/phalcon/mvc/application.zep
+++ b/phalcon/mvc/application.zep
@@ -145,6 +145,22 @@ class Application extends Injectable
 	{
 		return this->_modules;
 	}
+	
+	/**
+	 * Handles a user exception
+	 */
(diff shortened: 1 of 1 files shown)


sergeyklay:
@Surt can you please rebase onto 2.1.0?

sjinks:

Surt:
Yep, sorry, I created a new as @sjinks already saw. This one can be closed.
The problem is that I don't have the time to mount a cphalcon dev environment and finish the developing, so I just let it there hoping that someone will take the chance.

https://github.com/phalcon/cphalcon/pull/12746#discussion_r107943816

PR #10976, opened 2015-09-25
Correct return types for http request methods

Green-Cat:


sergeyklay:
#10966 :)

Green-Cat:
@sergeyklay oops.

PR #10983, opened 2015-09-27
Fixed return for Phalcon\Di

sergeyklay:


andresgutierrez:
```
Zephir\CompilerException: Returning type: null but this type is not compatible with return-type hints declared in the method in /home/travis/build/phalcon/cphalcon/phalcon/di.zep on line 416
```

sergeyklay:
@andresgutierrez fixed

andresgutierrez:
thanks

PR #11012, opened 2015-10-08
When http method is POST,  getPut() should get an empty array

AlloVince:
Please confirm this behavior change

### DIFF
--- a/phalcon/http/request.zep
+++ b/phalcon/http/request.zep
@@ -117,6 +117,10 @@ class Request implements RequestInterface, InjectionAwareInterface
 	 */
 	public function getPut(string! name = null, var filters = null, var defaultValue = null, boolean notAllowEmpty = false, boolean noRecursive = false)
 	{
+		if !this->isMethod("PUT") {
+			return [];
+		}
+
 		var put;
 
 		let put = this->_putCache;
@@ -317,6 +321,15 @@ class Request implements RequestInterface, InjectionAwareInterface
 	public function isSecureRequest() -> boolean
 	{
 		return this->getScheme() === "https";
+	}
+
+	/**
+	 * Set request raw body for unit test
+	 */
+	public function setRawBody(string rawBody) -> <Request>
+	{
+	    let this->_rawBody = rawBody;
+	    return this;
 	}
 
 	/**
--- a/phalcon/http/request.zep
+++ b/phalcon/http/request.zep
@@ -118,7 +118,7 @@ class Request implements RequestInterface, InjectionAwareInterface
 	public function getPut(string! name = null, var filters = null, var defaultValue = null, boolean notAllowEmpty = false, boolean noRecursive = false)
 	{
 		if !this->isMethod("PUT") {
-			return [];
+			return name ? null : [];
 		}
 
 		var put;
@@ -321,15 +321,6 @@ class Request implements RequestInterface, InjectionAwareInterface
 	public function isSecureRequest() -> boolean
 	{
 		return this->getScheme() === "https";
-	}
-
-	/**
-	 * Set request raw body for unit test
-	 */
-	public function setRawBody(string rawBody) -> <Request>
-	{
-	    let this->_rawBody = rawBody;
-	    return this;
 	}
 
 	/**
--- a/unit-tests/RequestTest.php
+++ b/unit-tests/RequestTest.php
@@ -390,7 +390,7 @@ class RequestTest extends PHPUnit_Framework_TestCase
 		stream_wrapper_unregister('php');
 		stream_wrapper_register('php', 'PhpStream');
 
-		file_put_contents('php://input', [json_encode(['bar' => 'baz'])]);
+		file_put_contents('php://input', array(json_encode(array('bar' => 'baz'))));
 
 
 		$data = file_get_contents('php://input');
@@ -405,7 +405,7 @@ class RequestTest extends PHPUnit_Framework_TestCase
 	{
 		stream_wrapper_unregister('php');
 		stream_wrapper_register('php', 'PhpStream');
-		file_put_contents('php://input', ["foo" => "bar"]);
+		file_put_contents('php://input', array("foo" => "bar"));
 		$_SERVER['REQUEST_METHOD'] = 'POST';
 		$request = new \Phalcon\Http\Request();
 		$this->assertEquals([], $request->getPut());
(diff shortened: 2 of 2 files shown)


phalcon/http/request.zep
```diff
@@ -320,6 +324,15 @@ class Request implements RequestInterface, InjectionAwareInterface
 	}
 
 	/**
+	 * Set request raw body for unit test
+	 */
+	public function setRawBody(string rawBody) -> <Request>
+	{
+	    let this->_rawBody = rawBody;
+	    return this;
+	}
```
sergeyklay: Tell me at least one reason on which the Request body shouldn't be read-only?

AlloVince: Please see my unit test case, because Phalcon get request body by `file_get_contents("php://input")`, but we are not able to change `php://input`, which means we are not able to mock a request and make unit test for a request with rawbody. I think add `setRawBody`  is necessary

sergeyklay:
```php
<?php

class PhpStream
{
    protected $index = 0;
    protected $length = 0;
    protected $data = '';

    public $context;

    public function __construct()
    {
        if (file_exists($this->buffer_filename())) {
            $this->data = file_get_contents($this->buffer_filename());
        } else {
            $this->data = '';
        }

        $this->index = 0;
        $this->length = strlen($this->data);
    }

    protected function buffer_filename()
    {
        return sys_get_temp_dir() . DIRECTORY_SEPARATOR . 'php_input.txt';
    }

    public function stream_open($path, $mode, $options, &$opened_path)
    {
        return true;
    }

    public function stream_close()
    {
    }

    public function stream_stat()
    {
        return [];
    }

    public function stream_flush()
    {
        return true;
    }

    public function stream_read($count)
    {
        if (null === $this->length) {
            $this->length = strlen($this->data);
        }

        $length = min($count, $this->length - $this->index);
        $data = substr($this->data, $this->index);
        $this->index = $this->index + $length;

        return $data;
    }

    public function stream_eof()
    {
        return ($this->index >= $this->length ? true : false);
    }

    public function stream_write($data)
    {
        return file_put_contents($this->buffer_filename(), $data);
    }

    public function unlink()
    {
        if (file_exists($this->buffer_filename())) {
            unlink($this->buffer_filename());
        }

        $this->data = '';
        $this->index = 0;
        $this->length = 0;
    }
}
```

**Test php://input**
```php
stream_wrapper_unregister('php');
stream_wrapper_register('php', PhpStream::class);

file_put_contents('php://input', [json_encode(['bar' => 'baz'])]);


$data = file_get_contents('php://input');

var_dump($data);
assert($data === file_get_contents('php://input'));
```

AlloVince: Thanks for the solution, I removed setRawBody method and make unit test as your way

unit-tests/RequestTest.php
```diff
@@ -379,4 +384,109 @@ public function testIssues2294()
...
+		$this->assertEquals($data, json_encode(['bar' => 'baz']));
+
+		$_POST['foo'] = 'bar';
+		$this->assertNotEquals($_POST, $data);
+	}
+
+	public function testPutEmptyWhenMethodIsPost()
+	{
+		stream_wrapper_unregister('php');
+		stream_wrapper_register('php', PhpStream::class);
```
sergeyklay: This syntax for php >= 5.5, use `'PhpStream'` instead

unit-tests/RequestTest.php
```diff
@@ -379,4 +384,109 @@ public function testIssues2294()
 
 		$this->assertEquals($request->getHeaders(), $headers);
 	}
+
+	public function testPhpStream()
+	{
+		stream_wrapper_unregister('php');
```
sergeyklay: Need rollback this after test:
```php
stream_wrapper_restore('php');
```

unit-tests/RequestTest.php
```diff
@@ -379,4 +384,109 @@ public function testIssues2294()
...
+	}
+
+	public function testPutEmptyWhenMethodIsPost()
+	{
+		stream_wrapper_unregister('php');
+		stream_wrapper_register('php', PhpStream::class);
+		file_put_contents('php://input', ["foo" => "bar"]);
+		$_SERVER['REQUEST_METHOD'] = 'POST';
+		$request = new \Phalcon\Http\Request();
+		$this->assertEquals([], $request->getPut());
```
sergeyklay:
```php
$this->assertEquals([], $request->getPut());
```

Okay, but below you use
```php
file_put_contents('php://input', ["foo" => "bar"]);
```

unit-tests/RequestTest.php
AlloVince: thx for the review, stream restore already added

unit-tests/RequestTest.php
```diff
@@ -379,4 +384,111 @@ public function testIssues2294()
 
 		$this->assertEquals($request->getHeaders(), $headers);
 	}
+
+	public function testPhpStream()
+	{
+		stream_wrapper_unregister('php');
+		stream_wrapper_register('php', 'PhpStream');
+
+		file_put_contents('php://input', [json_encode(['bar' => 'baz'])]);
```
sergeyklay: We still use PHP 5.3. Please use `array( )` instead

unit-tests/RequestTest.php
```diff
@@ -379,4 +384,111 @@ public function testIssues2294()
...
+	public function testPhpStream()
+	{
+		stream_wrapper_unregister('php');
+		stream_wrapper_register('php', 'PhpStream');
+
+		file_put_contents('php://input', array(json_encode(array('bar' => 'baz'))));
+
+
+		$data = file_get_contents('php://input');
+		$this->assertEquals($data, json_encode(['bar' => 'baz']));
```
sergeyklay: `[ ]` =>  `array( )`

unit-tests/RequestTest.php
```diff
@@ -379,4 +384,111 @@ public function testIssues2294()
...
+		return true;
+	}
+
+	public function stream_close()
+	{
+	}
+
+	public function stream_stat()
+	{
+		return [];
```
sergeyklay: `return array();`

sergeyklay:
@AlloVince Could you please deal with this?

sergeyklay:
I'm closing this issue due to the lack of any reaction. I'll open it again if the need arises

Issue #11024, opened 2015-10-16
[BUG] Phalcon 2.1.b: Libmemcached::queryKeys($keyPrefix) - "$keyPrefix" is ignored, all keys are returned.


temuri416:
Hi,

The code:

``` php
class Memc extends Phalcon\Cache\Backend\Libmemcached
{
    function getConn() {
        return $this->_memcache;
    }
}

 $frontCache = new \Phalcon\Cache\Frontend\Data(array(
    "lifetime" => 172800
 ));
 $cache = new Memc($frontCache, array(
     "servers" => array(
         array('host' => 'localhost',
               'port' => 11211,
               'weight' => 1),
     ),
     'statsKey' => '_PHCM',
));

$cache->save('prefix1-myKey', ['a', 'b']);
$cache->save('prefix2-myKey', ['x', 'z']);
$prefix1Keys = $cache->queryKeys('prefix1');
$_PHCM = $cache->getConn()->get('_PHCM');
```

I expect `$prefix1Keys` to only contain **prefix1-myKey**, however it also includes **prefix2-myKey**.

**$_PHCM** correctly lists both keys as array keys and TTL as values.

Zephir source of **queryKeys()** is dead simple and my **$_PHCM** is equivalent of **keys** below:

``` php
let keys = memcache->get(specialKey);
if typeof keys == "array" {
    let keys = array_keys(keys);
    for key in keys {
        if prefix && !starts_with(key, prefix) {
            unset keys[key];
        }
    }
}
```

No idea what may be failing there...

Thanks!

temuri416:
@andresgutierrez 

I think there's an obvious bug in ZEP code around here:

https://github.com/phalcon/cphalcon/blob/2.1.x/phalcon/cache/backend/libmemcached.zep#L347

Once we flipped contents of ```$_PHCM``` array, cache keys are values of ```keys``` array. So, the ```for key in keys``` yields cached key. Thus, ```unset keys[key];``` is wrong.

So, the ZEP code should be changed to:

```zep
let keys = memcache->get(specialKey);
if typeof keys == "array" {
    let keys = array_keys(keys);
    for key, idx in keys {
        if prefix && !starts_with(key, prefix) {
            unset keys[idx];
        }
    }
}
```

If you confirm it, I'll create a pull request.

Thanks!

temuri416:
@sergeyklay any chance you fix this? thanks!

sergeyklay:
I'll try to sort out

[...]

PR #11028, opened 2015-10-17
2.0.x. Fix for queryKeys() ignoring prefix.

temuri416:
Issue #11024


andresgutierrez:
@patrick-zippenfenig Can you check please?

patrick-zippenfenig:
@andresgutierrez Sorry for the late reply (traveling in asia). Yes, this is an obvious error and the solution looks good. 

@temuri416 Could you check the implementation of all cache adapters? There are some inconsistencies.

sergeyklay:
@temuri416

Issue #10405, opened 2015-05-26
[2.0.1]RuntimeException: Trying to call method appendmessage on a non-object


wojewsky:
Hi,

with phalon 1.3.x we've added error messages to form elements by ->appendMessage.
With phalcon 2.0.1 i've got an runtime exception:
RuntimeException: Trying to call method appendmessage on a non-object

For example:

```php
    $form = new \Phalcon\Forms\Form();
    $element = new \Phalcon\Forms\Element\Text('field');
    $form->add($element);
    $element->appendMessage(new \Phalcon\Validation\Message('error message'));
    die(var_dump($form->getMessages()));
```

noobiwan:
Any update on this? I'm using Phalcon 2.1.X and I'm getting the same error!

noobiwan:
I was checking the source code and the problem seems to occur when `appendMessage()` is called before a call to `validate()`.  The method `appendMessage` depends on `this->_messages` being set ([line 346](https://github.com/phalcon/cphalcon/blob/2.1.x/phalcon/validation.zep#L346)) but it is only set in the `validate()` method at [line 111](https://github.com/phalcon/cphalcon/blob/2.1.x/phalcon/validation.zep#L111)

It seems to me that the logic is that `appendMessage` should be called in a *beforeValidation* event, that is triggered from the `validate()` method, but maybe there is a validity to allow messages to be added before that, or to throw an exception so the error is clear!

Zaszczyk:
Up, I also have this problem.

Knallcharge:
Same here. If I call "method_exists($validator, 'appendMessage')" it returns true, calling appendMessage on the next line will say that validator is not an object ... !?

Jurigag:
@Knallcharge what phalcon version you are using ?

Knallcharge:
@Jurigag 2.0.8

Jurigag:
I have 2.0.13 and can't reproduce problem with code from OP. Can you post your code @Knallcharge to reproduce this ?

Knallcharge:
Using the validator in a unit-test will result in the described behavior.

`$myValidator = new MyValidator();
$validation = new Phalcon\Validation();
$validation->bind(new stdClass(),
    array(
        'day'   => date('d'),
        'month' => date('m'),
        'year'  => date('Y')+1
    ));

$myValidator->validate($validation, '');`

[...]

PR #11038, opened 2015-10-21
Throw exception when calling appendMessage outside of validators

luiz-brandao:
This is a quick fix for issue #10405. Please check if we should use `empty` or `isset`.

### DIFF
--- a/phalcon/validation.zep
+++ b/phalcon/validation.zep
@@ -343,6 +343,10 @@ class Validation extends Injectable implements ValidationInterface
 	 */
 	public function appendMessage(<MessageInterface> message) -> <Validation>
 	{
+		if empty this->_messages {
+			throw new Exception("You can only append messages from validators");
+		}
+		
 		this->_messages->appendMessage(message);
 		return this;
 	}


phalcon/validation.zep
```diff
@@ -343,6 +343,10 @@ class Validation extends Injectable implements ValidationInterface
 	 */
 	public function appendMessage(<MessageInterface> message) -> <Validation>
 	{
+		if empty this->_messages {
+			throw new Exception("You can only append messages from validators");
+		}
+
```
sergeyklay: Why not?
```zephir
var messages;

let messages = this->_messages;
if typeof messages != "object" {
    let messages = new Group();
}

messages->appendMessage(message);

let this->_messages = messages;

return this;
```

sergeyklay:
Could you please cover this feature by test?

noobiwan:
Yeah, that would be ideal but the problem is that on [line 111](https://github.com/phalcon/cphalcon/blob/2.1.x/phalcon/validation.zep#L111) the validate method is overwriting `this->_messages` and I didn't want to touch that. So I guess we should implement this code there too on [line 95](https://github.com/phalcon/cphalcon/blob/2.1.x/phalcon/validation.zep#L95)?

About the example you provided, why not simply:

```zep
if typeof this->_messages != "object" {
    let this->_messages = new Group();
}

this->_messages->appendMessage(message);

return this;
```
PS: I just started studying Zephir.

phalcon/validation.zep
noobiwan: What about simply:

```zep
if typeof this->_messages != "object" {
    let this->_messages = new Group();
}

this->_messages->appendMessage(message);

return this;
```

sergeyklay: operations with global variables more expensive. more expensive than php

sergeyklay:
@noobiwan

noobiwan:
Hi @sergeyklay, is help still needed? I already implemented the changes, now I need to write the unit test but I'm not sure where to write it. Should I write it in the FormsTest.php or ValidationTest.php? Please advice.

sergeyklay:
Hello @noobiwan You should use ValidationTest

noobiwan:
@sergeyklay Could you tell me the steps to get the tests running? It's my first time doing this with Phalcon. I updated the .zep files and wrote a unit test. Should I compile Phalcon and then run the tests? Right now I'm getting "No tests executed!" output. Please advice.

Jurigag:
There are commands to run tests:

vendor/bin/phpunit --bootstrap tests/_ci/phpunit.php unit-tests/
vendor/bin/codecept build
vendor/bin/codecept run

noobiwan:
@Jurigag there is no _ci diretory under tests/

Jurigag:
Beacause you first need to composer install --dev --prefer-source

sergeyklay:
We started to refactor tests (https://github.com/phalcon/cphalcon/pull/11479, https://github.com/phalcon/cphalcon/pull/11432).

All new tests should  be placed in https://github.com/phalcon/cphalcon/tree/2.1.x/tests/unit

Neither `FormsTest` nor `ValidationTest` not refactored yet. You should get one test method 
that covers your changes and implement new test with one method in new place.

Also please see our test guide: https://github.com/phalcon/cphalcon/blob/2.1.x/tests/README.md

sergeyklay:
Fixed in the `2.1.x` branch.

Issue #11021, opened 2015-10-14
Loader doesn't follow PSR-4 specification


luiz-brandao-jr:
I ran into a problem and I realized that the Loader doesn't allow a namespace to be mapped to multiple directories. Take this [PSR-4 example](http://www.php-fig.org/psr/psr-4/examples/) for instance:

``` php
$loader->addNamespace('Foo\Bar', '/path/to/packages/foo-bar/src');
$loader->addNamespace('Foo\Bar', '/path/to/packages/foo-bar/tests');
```

This is currently not possible with Phalcon\Loader because it threats the namespaces array as unidimensional. Is this a know issue? Any plans to fix that?

sergeyklay:
Yes, we plan to fix it. Something like this:
```
$loader->addNamespace('Foo\Bar', [ '/path/to/packages/foo-bar/src']);

$loader->addNamespace(
    'Buz\Bar',
    [
         '/path/to/packages/buz-bar/src1',
         '/path/to/packages/buz-bar/src2',
          '/path/to/packages/buz-bar/src3'
    ]
);
```

sergeyklay:
@noobiwan I've introduced `Phalcon\Loader\Extended` in the [Phalcon Incubator](https://github.com/phalcon/incubator/pull/457). This component extends `Phalcon\Loader` and adds ability to set multiple directories a namespace.

Also I have [tried to implement](https://github.com/phalcon/cphalcon/issues/11044) this directly in the Phalcon. But we have some difficulties with the latest version of Zephir and Phalcon 2.1 . This will be fixed later.

However, I would like to draw your attention that even if we fix this PR or/and some Zephir issues, the ability that you described will be implemented in Phalcon 2.1, not in 2.0.x

butamuh4o:
Thank you @sergeyklay

Issue #692, opened 2013-06-06
[NFR] Namespace-related views paths


hesalx:
What about to implement namespace-related views paths?
I mean the implementation in the core level, because it could be done by dynamic call of view->pick() but it would be really slower and not pretty at all =)

It would be great to have such structure:
/views/name/space/my_controller/my_action.phtml
/views/layouts/name/space/my_controller.phtml
/views/partials/part.phtml
/views/index.phtml
1. Single general index.phtml (main layer)
2. Shared partials.
3. Separated controllers/actions views.
4. Separated layouts.
##

nsossonko:
This functionality would definitely be appreciated in our environment. We namespace our controllers for organization and proper hierarchical structure but the view module doesn't recognize namespaces, so our views folder has to be "un-namespaced".

tugrul:
I have done using @arantir 's way.

Plugin:
```php
namespace App\Website\Plugins;

class CustomRenderer extends \Phalcon\Mvc\User\Plugin
{
    public function beforeExecuteRoute($event, $dispatcher)
    {
        $defaultNamespace = $dispatcher->getDefaultNamespace();
        $namespace = $dispatcher->getNamespaceName();
        
        if (strpos($namespace, $defaultNamespace) === 0) {
            $namespace = substr($namespace, strlen($defaultNamespace));
        }

        $parts   = array_filter(explode('\\', $namespace));
        $parts[] = $dispatcher->getControllerName();
        $parts[] = $dispatcher->getActionName();

        $this->view->pick(implode(DIRECTORY_SEPARATOR, $parts));
    }
}
```

Binding:
```php
$di->set('dispatcher', function() use ($di) {
            $eventsManager = $di->getShared('eventsManager');
            $eventsManager->attach('dispatch', new CustomRenderer()); 

            $dispatcher = new Dispatcher();
            $dispatcher->setEventsManager($eventsManager);
            $dispatcher->setDefaultNamespace('App\Website\Controllers'); 
            return $dispatcher;
        }); 
```

challet:
@trugrul There is an (it seems undocumented) [application:viewRender](https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/application.zep#L336) event that might be closer to what you want. In case many dispatches happend before one render, as when  _$dispatcher->forward_ is called.
The plugin is great, with one tiny downside : the "default namespace" is also becoming the "root namespace".

PR #11044, opened 2015-10-21
Introducing ability of multiple directories for namespace

sergeyklay:
Fixed issue #11021

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -199,4 +199,4 @@ ext/phalcon/annotations/lemon.dSYM/
 ext/phalcon/annotations/parser.out
 ext/phalcon/mvc/model/query/lemon
 ext/phalcon/mvc/model/query/parser.out
-vendor/
+/vendor/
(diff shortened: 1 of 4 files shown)


andresgutierrez:
Thanks!

virgofx:
@sergeyklay Does this incorporate namespaced related view/partial support as outlined in #692?  If not, would it be possible to incorporate this into the PR as well?

Thanks!

sergeyklay:
@virgofx  This PR introduce class loading through searching for the namespace in multiple directories. 
Only classes, not views. In particular, it solves [this issue](https://github.com/phalcon/cphalcon/issues/11021) only.

virgofx:
@sergeyklay Since you just worked on this any chance you could see about implementing `https://github.com/phalcon/cphalcon/issues/692#issuecomment-50416113` which would just be directory support for views based on controller namespaces?  I have a small mockup in my code base extending View to handle this.

Pseudo Code:

View:
* Add a flag to not break backward compatibility.
* Update render() to check if flag is defined from [ useNamespacedViews() ] and if so, quickly parse the namespace and render the hierarchy starting at the bottom.  (Essentially each parent just gets rendered as a template before/after would)

Issue #11049, opened 2015-10-22
[2.0.8] Phalcon\Paginator\Adapter\QueryBuilder notice when query has group by (mysql)


ashpumpkin:
```PHP
class Robots extends Model
{
	public $id;
	public $group_by_field;
}

$di->get('db')->execute('DROP TABLE IF EXISTS robots;');
$di->get('db')->execute('CREATE TABLE `robots` (
  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `group_by_field` int(10) unsigned,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8;');

// no robots, empty table

$queryBuilder = new \Phalcon\Mvc\Model\Query\Builder();
$queryBuilder->addFrom('Robots')->groupBy('group_by_field');

$robotsPaginator = new \Phalcon\Paginator\Adapter\QueryBuilder([
	'builder' => $queryBuilder,
	'limit'   => 10,
	'page'    => 1
]);

$robotsPaginator->getPaginate()->items;

```
this code produces *Notice: Phalcon\Paginator\Adapter\QueryBuilder::getPaginate(): Trying to get property &quot;rowcount&quot; of non-object*

because query with rowcount returns empty set. I think getPaginate() should check this set on emptyness

PR #11050, opened 2015-10-22
Fixed notice in query builder paginator when totalQuery result is empty

ashpumpkin:
issue #11049


andresgutierrez:
`intval` is not returning zero?

sergeyklay:
```
php > # it is permissible scenario for 'row = result->getFirst()'
php > $row = null; 
php > 
php > var_dump(intval($row->rowcount));
PHP Notice:  Trying to get property of non-object in php shell code on line 1
PHP Stack trace:
PHP   1. {main}() php shell code:0

int(0)  # <--  Yeah :)
```

ashpumpkin:
Please read the issue, in some cases row can be false, and row->rowcount produces notice

andresgutierrez:
Thanks

PR #11054, opened 2015-10-23
#11042 - Fix setting relation should not always cause dirty state change

### LINKED ISSUE #11042: Accessing relation in model causes dirty state to change to transient
Code:

```php
  class Parent extends \Phalcon\Mvc\Model {
    public $id;
    public function initialize() {
      $this->hasMany('id', 'Child', 'parent', array(
        'foreignKey' => array('action' => Relation::ACTION_RESTRICT),
        'alias'  => 'relatedchilds'));
    }
  }
  class Child extends \Phalcon\Mvc\Model {
    public $parent;
    public $id;
    public function initialize() {
      $this->belongsTo('parent', 'Parent', 'id', array(
          'alias' => 'relatedparent',
          'foreignKey' => array('message' => 'Must have parent')
      ));
    }
  }

  $children = Child::find(); // get all child records
  foreach ($children as $child) {
    echo "Initial state: {$child ->getDirtyState()} <br />";
    $parent = $child ->relatedparent;
    echo "After accessing parent - state: {$child ->getDirtyState()} <br />";
  }
```

Is this by design? If so, why? This is causing us big headaches in our application, so any help would be appreciated.

nsossonko:
This should fix #11042 although I have no real way of testing it. Hope my style follows conventions, this is my first attempt at hacking Phalcon...

### DIFF
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -4090,17 +4090,21 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 	public function __set(string property, value)
 	{
 		var lowerProperty, related, modelName, manager, lowerKey,
-			relation, referencedModel, key, item;
+			relation, referencedModel, key, item, dirtyState;
 
 		/**
 		 * Values are probably relationships if they are objects
 		 */
 		if typeof value == "object" {
 			if value instanceof ModelInterface {
+				let dirtyState = this->_dirtyState;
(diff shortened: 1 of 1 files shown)


nsossonko:
Third time's the charm...this seems to be working now

sergeyklay:
We don't add changes directly to stable branches or `master`. Can you please do PR to 2.0.x branch?

PR #11059, opened 2015-10-24
Flash\Session::getMessages with type and removal fix

zeopix:


sergeyklay:
1. We don't add changes directly to stable branches or `master`. Can you please do PR to 2.0.x branch?
2. Also can you please explain a bit more what exactly did you fix.
3. When changes of this scope, it is desirable to provide a small test

zeopix:
Sure, will open a new pull request.
A test is provided for uncovered flash session adapter, also for the fix.
Thanks in advance.

PR #11060, opened 2015-10-25
Flash\Session::getMessages with type and removal fix

zeopix:
1. Fixed getMessages in Session Flash Adapter. When retriving messages for a specific type, all messages get removed, so method usage is misleading, see: FlashSessionDefaultTest::testGetMessagesTypeRemoveMessages.
2. Added test cases for Session Adapter, which slightly differs from Direct implementaiton.
3. Added a InMemory mock helper for session adapter.


sergeyklay:
Now the tests are falling

zeopix:
@sergeyklay tests fixed

sergeyklay:
Thanks

Issue #11031, opened 2015-10-18
Fix ODM logic


dzubchik:
Currently ODM working in some interesting way. When do query with set of fields - it returns document,  without - returns collection. My proposition is to correct this - always return document or set of documents, and give ability to change document class to be returned. How about that?

sergeyklay:
:+1:

ashpumpkin:
@sergeyklay 
Maybe my hydration in ODM pr (https://github.com/phalcon/cphalcon/pull/11013) will be alright for that issue?

sergeyklay:
@dzubchik ?

sergeyklay:
@ashpumpkin This PR doesn't solves the problem completely.

ashpumpkin:
@sergeyklay why not?
```PHP
Robots::find(); //returns Robots
Robots::find(['hydration' => Collection::HYDRATE_DOCUMENTS]); // you get Documents
Robots::find(['hydration' => Collection::HYDRATE_COLLECTIONS]); // you get Robots
Robots::find(['hydration' => Collection::HYDRATE_ARRAYS]); // you get raw arrays (the fastest one)
```
and the same works with findFirst(). I think the problem is completely solved :smiley:

aleksandrzen:
:+1:

dzubchik:
@ashpumpkin , it does not solve the issue because you can not declare the document class for collection

ashpumpkin:
@dzubchik @sergeyklay 
How about that:
```PHP
Robots::find(); //returns Robots
Robots::find(['class' => 'Books']); // you get Books
Robots::find(['class' => 'Phalcon\Mvc\Collection\Document']); // you get Documents
Robots::find(['class' => 'UnknowClass']); // you get php error Class 'UnknownClass' not found
Robots::find(['class' => 'KnownClassNotImplementingInterface']); // you get phalcon exception "Class must be an implementation of Phalcon\Mvc\CollectionInterface or an instance of Phalcon\Mvc\Collection\Document"
```
I almost implemented it on my local machine and can make a pr if this realization is good and solving the problem completely.

dzubchik:
@ashpumpkin , I thought it would be something like this:
```php
<?php

use Phalcon\Mvc\Collection;

class Robots extends Collection
{
    public static $documentClass = 'My\Namespace\Robot\Document';
}
```

And by default it would search document class in same namespace as collection. And if there are no document class - uses default.

ashpumpkin:
[...]

PR #11063, opened 2015-10-25
Added ability to change document class to be returned in ODM

ashpumpkin:
issue #11031
Usage:

``` PHP
Robots::find(); //returns Robots
Robots::find(['class' => 'Books']); // you get Books
Robots::find(['class' => 'Phalcon\Mvc\Collection\Document']); // you get Documents
Robots::find(['class' => 'UnknowClass']); // you get php error "Class 'UnknownClass' not found"
Robots::find(['class' => 'KnownClassNotImplementingInterface']); // you get phalcon exception "Object of class 'KnownClassNotImplementingInterface' must be an implementation of Phalcon\Mvc\CollectionInterface or an instance of Phalcon\Mvc\Collection\Document"
```

In case when exception must be thrown all other check become useless waste of time and that is why class check is made so early.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Added `strict` option for ExclusionIn validator
 - Added `Phalcon\Text::underscore` - to make a phrase underscored instead of spaced
 - Added `Phalcon\Text::humanize` - to make an underscored or dashed phrase human-readable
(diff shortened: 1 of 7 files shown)


andresgutierrez:
Could you please rebase?

sergeyklay:
@andresgutierrez

andresgutierrez:
Thanks

PR #11067, opened 2015-10-26
Fixed fallback assign in _set method not assigning when related type is Array

HitoriSensei:


sergeyklay:
We don't add changes to `master`, could you please submit this to the 2.0.x branch?

bmoore:
@sergeyklay I'm not sure if this is applicable anymore to 2.1.x or 2.0.x due to #11320 and #11417

PR #11085, opened 2015-11-03
Add 'strict' option in ExclusionIn validator

altapo:


sergeyklay:
We don't add changes directly to stable branches or `master`. Can you please do PR to `2.0.x` branch?

Issue #10905, opened 2015-09-08
why the backend cache redis add the special key(statsKey) _PHCR


lixiphp:
when the cache is expired, the key still exists in statsKey _PHCR, why design this sepcial key store all the index keys?

Green-Cat:
It exists to be able to flush the Redis cache from phalcon. It is disabled by default in the 2.0.x branch now though.

lixiphp:
@Green-Cat it is a good idea. So flushing cache can realize via PHP code not impacted by KV Redis.

PR #11125, opened 2015-11-12
Modify Redis Backend

pantaovay:
## Changes
1. Modify cache redis backend, delete _PHCR
2. Cache redis backend supports "redis" client option
## Why
1. The redis cache backend use a special key `_PHCR` to store all keys used by cache. There is a issue: #10905 .According to @Green-Cat ,it is disabled in 2.0.x branch which is not true. If use `_PHCR`, every time we set a cache, we have to run `sAdd` and `set` commands. Store all keys in `_PHCR` will cause it expands quickly and this will influence performance.
2. The `libmemecached` backend has supplied the option to disable this feature.
3. The cache backend doesn't support redis option which means that we can not pass a redis connection. So if we use modelsCache and modelsMetadata, there will be two redis connections. So I modify this part and let redis option be supported, and through the lifetime of a request we can use one redis connection.


pantaovay:
@sergeyklay

sergeyklay:

Issue #11136, opened 2015-11-14
Mysql vs MySQL


beliarh:
https://github.com/phalcon/cphalcon/blob/master/phalcon/db/adapter/pdo/mysql.zep#L45
```zephir
class Mysql extends PdoAdapter implements AdapterInterface
```
https://github.com/phalcon/cphalcon/blob/master/phalcon/db/dialect/mysql.zep#L35
```zephir
class MySQL extends Dialect
```

sergeyklay:
Can you please explain a bit more

beliarh:
It would be nice if these classes had the same name

sergeyklay:
Fixed in `2.1.x`

PR #11138, opened 2015-11-15
Fixed issue #11136

sergeyklay:


Green-Cat:
That's a BC break, should probably add it to the read me file or something.

SidRoberts:
It's not a BC break as Zephir classes are not case-sensitive. As an example, [this test should have failed before](https://github.com/phalcon/cphalcon/blob/master/unit-tests/DbDialectTest.php#L1003).

Green-Cat:
Oh my bad.

PR #11145, opened 2015-11-18
add comment.Volt syntax.

yuji0602:
Hi.

I have added the use of check_field.


sergeyklay:
We don't add changes directly to stable branches or `master`. Can you please create PR to `2.0.x` branch?

yuji0602:
ok, thx.

PR #11146, opened 2015-11-18
fix cookie construct

mich-grodno:
if set value must be:
this->_readed = true;


sergeyklay:
We don't add changes directly to stable branches or `master`. Can you please create PR to `2.0.x`branch?

PR #11178, opened 2015-12-01
Gets Query instance from DI container

tmihalik:
as in: https://github.com/phalcon/cphalcon/blob/c8d4916bbfd0cb18395fe25d8c5a663fc5077d8f/phalcon/mvc/model/manager.zep#L1671


sergeyklay:
Looks good to me

sergeyklay:
Thanks

PR #11180, opened 2015-12-01
Added ability to spoof HTTP method

sergeyklay:


sergeyklay:
@andresgutierrez 

Also I made some changes with return value.
In my opinion `Request::getMethod` newer must return empty string. Empty string is invalid HTTP method. In case of an error by recognition a method, any browser will return `GET`. 

In my opinion `Request::getMethod` newer must return empty string. Empty string is invalid HTTP method. Any browser failing to recognize method will return get.

PR #11182, opened 2015-12-01
hard time limit to Libmemcached session adapter

challet:
Memcache has an internal lifetime limit of 30 days, see http://stackoverflow.com/questions/1418324/memcache-maximum-key-expiration-time
Setting something upper to this limit leads to having blank writes with no error, breaking the purpose of the session.

### DIFF
--- a/phalcon/session/adapter/libmemcached.zep
+++ b/phalcon/session/adapter/libmemcached.zep
@@ -78,7 +78,7 @@ class Libmemcached extends Adapter implements AdapterInterface
 			let lifetime = 8600;
 		}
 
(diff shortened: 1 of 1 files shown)


challet:
I don't know why it failed, it seems there is nothing related to libmemcache or to session here https://github.com/phalcon/cphalcon/blob/master/unit-tests/CacheTest.php#L685

sergeyklay:
Firstly remove constant. It is not necessary here, the comment line in the code will be enough. Then restart build by close/open PR.

sergeyklay:
Thanks

Issue #11174, opened 2015-11-30
[Bug] \Phalcon\Mvc\Dispatcher does not set Last Handler on Exception


virgofx:
When an exception occurs dispatching, the last handler will not get set.  The setting of the last handler is only performed on success but needs to happen in **all** cases as subsequent error handling actions further up the stack that may dispatch again would require this information (as in the case in one of my error handling applications) This could easily be achieved with a finally block for the try/catch and moving the setting of the last handler into that section.

See
https://github.com/phalcon/cphalcon/blob/master/phalcon/dispatcher.zep#L510

rianorie:
Can you produce a script that reproduces the behavior?

virgofx:
`IndexController.php`
```php
class IndexController extends \Phalcon\Mvc\Controller
{
    public function indexAction() {
        throw new \MyCustomException('custom exception');
    }
}
```

`bootstrap.php`
```php
$di = new \Phalcon\Di\FactoryDefault();
try {
		$application = new \Phalcon\Application();
		$application->setEventsManager(new \Phalcon\Events\Manager()));
		$application->setDI($di);

		echo $application->handle()->getContent();

	} catch (\MyCustomException $e) {
		// 'IndexController'  [OK]
		var_dump($di->getDispatcher()->getActiveController());
		
		// [empty]  [ WRONG ]
		// [Expected: "IndexController"]
		var_dump($di->getDispatcher()->getLastController());

		// Custom logic here to send errors or do something custom that requires
		// knowledge of the last controller.
	}
}
```

**The Fix**

_Change_
```php
try {
	// We update the latest value produced by the latest handler
	let this->_returnedValue = call_user_func_array([handler, actionMethod], params),
		this->_lastHandler = handler;

} catch \Exception, e {
	if this->{"_handleException"}(e) === false {
		if this->_finished === false {
			continue;
		}
	} else {
		throw e;
	}
}
```


_To_
```php
try {
	// We update the latest value produced by the latest handler
	let this->_returnedValue = call_user_func_array([handler, actionMethod], params);

} catch \Exception, e {
	if this->{"_handleException"}(e) === false {
		if this->_finished === false {
			continue;
		}
	} else {
		throw e;
	}
} finally {
	this->_lastHandler = handler;
}
```

rianorie:
.. you've adjusted the zephir code as such? Then why not make it a pull request? :)

I'm happy to do it, though

virgofx:
[...]

PR #11183, opened 2015-12-01
Fixed dispatcher not setting last handler per #11174

rianorie:


sergeyklay:
PHP 5.5 ? :)

rianorie:
yeah, I thought to get that fix in real quick before the end of day for 24pullrequests, but no such luck. Eventually redid it including an unit test in a way that did work

PR #11184, opened 2015-12-01
Dispatcher does not set last handler

rianorie:
#11174


virgofx:
@rianorie  Almost ... The setting of the last handler needs to be performed **after** the execution of the callback.   Otherwise, once inside the actual controller (for example the test error controller) ... Calling $this->getDispatcher()->getLastController() would be the current controller (which is incorrect). 

Simply moving it into a finally() block will fix the issue.

rianorie:
There is no `finally` in zephir as of yet, so that's not as simple. I
figured this would do the trick but I didn't realize the callback
intervened with the handler, dangit. I'll rework it tomorrow.

virgofx:
Ah no worries -- The finally was in PHP-land ... Just drop that same line just before the error handler is thrown and everything will work fine.

sergeyklay:
See #11421

Issue #1923, opened 2014-01-27
[BUG]  i'm try set 'shared_lock' option true, but not include statement.


iwai:
In document http://docs.phalconphp.com/en/latest/reference/models.html#finding-records

i used 'shared_lock' option set true, but 'LOCK IN SHARE MODE' not include statement.

ohmcoe:
+1 It same problem in 2.0.9

sergeyklay:
Fixed in 2.0.x branch

PR #11187, opened 2015-12-03
Fix bug #1923

dreamsxin:


phalcon/mvc/model/query/builder.zep
```diff
@@ -1370,6 +1370,12 @@ class Builder implements BuilderInterface, InjectionAwareInterface
 			query->setBindTypes(bindTypes);
 		}
 
+		if typeof this->_sharedLock === "boolean" {
+			if this->_sharedLock {
+				query->setSharedLock(true);
+			}
```
sergeyklay: Why not:
```zephir
if typeof this->_sharedLock === "boolean" {
    query->setSharedLock(this->_sharedLock);
}
```

dreamsxin: Because in the query default is false,  But you are right.

sergeyklay:
:+1:

andresgutierrez:
Thanks

PR #11192, opened 2015-12-04
Fixes #11190. Use stream_get_line() instead of fread().

### LINKED ISSUE #11190: Phalcon Beanstalk Queue errors with Reserve when job data > 16KB
When using job data that serializes to amounts greater than 16KB (currently testing around 64KB), the Beanstalk queue will error when performing a reserve() command as it attempts to read the data, and parse it into a job.

`$queue->reserve()`

This will result in unserialization errors as such:
```php
Notice: unserialize(): Error at offset 16365 of 16365 bytes in [file] on line [fileLine]
```

The problem is that the `Beanstalk::read()` does not work correctly when using `fread` when coming through `reserve()`. Although the exact bytes returned in the response are correct (actually +2 makes it correct as noted in the source).  The `fread()` will return data that is truncated around 16KB. 

Switching this to simply do a standard while loop or using `fgets()` works.

https://github.com/phalcon/cphalcon/blob/master/phalcon/queue/beanstalk.zep#L353

virgofx:
stream_get_line() works with beanstalkd streams and large buffers.

### DIFF
--- a/phalcon/queue/beanstalk.zep
+++ b/phalcon/queue/beanstalk.zep
@@ -350,7 +350,7 @@ class Beanstalk
 				return false;
 			}
 
-			let data = fread(connection, length + 2);
+			let data = stream_get_line(connection, length + 2);
 			if stream_get_meta_data(connection)["timed_out"] {
 				throw new \Phalcon\Exception("Connection timed out");
 			}


virgofx:
@andresgutierrez This one is good to go. Can we have this merged in both 2.0.x and 2.1.x ?

andresgutierrez:
I'll merge it to 2.1.x

andresgutierrez:
Thanks

virgofx:
Thank you

Issue #10955, opened 2015-09-17
a problem in paginator/querybuilder 


deathkel:
when I use groupBy in my builder,I got page.total_page always been 1,and page.total_item  =2

vikilaboy:
I have this problem on HAVING

```
SELECT a.id, a.idSubscription, a.fromEmail, a.fromName, CONCAT(a.fromName, ' [', a.fromEmail, ']') AS sender, a.subject, a.body, a.dateAdd, a.status, s.name AS subscription FROM [Dummy\Models\ContentReceived] AS [a] INNER JOIN [Dummy\Models\Subscription] AS [s] ON a.idSubscription = s.id HAVING 1 AND sender LIKE '%yoga%' ORDER BY a.id DESC
```
and 
```
Column 'sender' doesn't belong to any of the selected models (1), when preparing: SELECT COUNT(*) [rowcount] FROM [Dummy\Models\ContentReceived] AS [a] INNER JOIN [Dummy\Models\Subscription] AS [s] ON a.idSubscription = s.id HAVING 1 AND sender LIKE '%yoga%'
```

PR #11195, opened 2015-12-05
Fix total_items and total_pages in Paginator when it has `GROUP BY` #2411

ryomo:
#2411
#10955

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # [2.0.10](https://github.com/phalcon/cphalcon/releases/tag/phalcon-v2.0.10) (2015-11-24)
 - ORM: Added support for DATE columns in Oracle
+- Fixed wrong total_items and total_pages in Paginator when the query builder has set groupBy()
 
 # [2.0.9](https://github.com/phalcon/cphalcon/releases/tag/phalcon-v2.0.9) (2015-11-24)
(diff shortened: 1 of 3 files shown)


sergeyklay:
@andresgutierrez Test for PHP 5.3 fails because https://github.com/phalcon/zephir/pull/1139

sergeyklay:
@andresgutierrez Now tests fails because https://github.com/phalcon/zephir/pull/1141

andresgutierrez:
Thanks

pedoc:
@andresgutierrez
hi,i was use version 2.0.10,but except 'group',i had 'distinct(true) '  is wrong。sorry,my english is bad.
code example：
$builder=$this->modelsManager->createBuilder()->from("Robots")->leftJoin('Parts',"Parts.id=Robots.part_id")->distinct(true);
$paginator = new Phalcon\Paginator\Adapter\QueryBuilder(
            array(
                "builder" => $builder,
                "limit"   => 10,
                "page"    => 1
            )
        );
You can get to the two data，but，‘total_items’ property is 3。
I think the problem is the same ‘group’ and the，It is a bug?
Looking forward to your reply。

Jurigag:
As i see query builder paginator ignores distinct method at all. There is only check for groupBy but not for distinct. https://github.com/phalcon/cphalcon/blob/master/phalcon/paginator/adapter/querybuilder.zep

PR #11196, opened 2015-12-05
Do not test on php 5.3 anymore

Green-Cat:
It's time.
Ref phalcon/zephir#1139


sergeyklay:
:+1: 
@andresgutierrez ?

PR #11204, opened 2015-12-07
Get messages from getter method on ValidationFailed exception launch

renatomjr:
This change allow get translated messages in Exception, overriding the model's getMessages method, as suggested in the documentation (https://docs.phalconphp.com/pt/latest/reference/models.html#validation-messages) .


sergeyklay:
Thanks

PR #11214, opened 2015-12-10
Fix `BadMethodCallException: Wrong number of parameters` 

mr5:
Fix `BadMethodCallException: Wrong number of parameters` when php error occurred.

### DIFF
--- a/phalcon/debug.zep
+++ b/phalcon/debug.zep
@@ -560,7 +560,7 @@ class Debug
 	/**
 	 * Throws an exception when a notice or warning is raised
 	 */
-	public function onUncaughtLowSeverity(severity, message, file, line)
+	public function onUncaughtLowSeverity(severity, message, file, line, context)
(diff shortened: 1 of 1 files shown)


phalcon/debug.zep
```diff
@@ -560,7 +560,7 @@ class Debug
 	/**
 	 * Throws an exception when a notice or warning is raised
 	 */
-	public function onUncaughtLowSeverity(severity, message, file, line)
+	public function onUncaughtLowSeverity(severity, message, file, line, context)
```
sergeyklay: The `file`, `line` and `context` args are optional. See [set_error_handler](http://php.net/manual/en/function.set-error-handler.php).

Mb something like? : 
```zephir
public function onUncaughtLowSeverity(severity, message, string file = __FILE__, int line = __LINE__, array context = null)
{
    ...
}
```

PR #11271, opened 2016-01-04
Fixes #11269

### LINKED ISSUE #11269: Default namespace does not work in the annotation router
As per title, the annotation router does not take into account the namespace set with the setDefaultNamespace method.
So this doesn't work:
```php
use Phalcon\Mvc\Router\Annotations as AnnotationsRouter;

$di->set('router', function() use ($config) {

    //Use the annotations router
    $router = new AnnotationsRouter(false);

    $router->setDefaultNamespace('MyNamespace\Controller');

    $router->addResource('User', '/user');
}
```
But this does:
```php
use Phalcon\Mvc\Router\Annotations as AnnotationsRouter;

$di->set('router', function() use ($config) {

    //Use the annotations router
    $router = new AnnotationsRouter(false);

    $router->setDefaultNamespace('MyNamespace\Controller');

    $router->addResource('MyNamespace\Controller\User', '/user');
}
```

Green-Cat:


Green-Cat:
Travis build issue seems unrelated.

sergeyklay:
@Green-Cat Can you please provide a small test?

PR #11278, opened 2016-01-07
Fixed for building on windows.

ifsnow:
There were some problems when 'gen-build.php' was executed on windows.


andresgutierrez:
Thanks

PR #11280, opened 2016-01-07
Require at least php 5.4 (composer)

Green-Cat:
Since support for php 5.3 has been deprecated, we should not allow it to be used.


Green-Cat:
@sergeyklay

mattvb91:
:+1:

PR #11284, opened 2016-01-07
Fixed doc in Uniqueness validator.

SidRoberts:
See [line 92](https://github.com/phalcon/cphalcon/blob/902c32675776fbcfbd3a5a6178d4cd07833c7e79/phalcon/validation/validator/uniqueness.zep#L92).


sergeyklay:
Thanks

Issue #11267, opened 2016-01-04
[NFR] Bind model in router


mattvb91:
Couldn't find any discussions on this maybe someone can link me.

Would there be any interest in having the ability to bind models in the router like we can in laravel 5:
https://laravel.com/docs/5.2/routing#route-model-binding

That way we can then in the controller actions inject the correct model for you already without having to manually do a find() or similar.

```
public function edit(Product $product)
{
      //bla
}
```

I guess the syntax could be similar too so instead of this:

```
$router->addGet("/products/edit/{id}", "Products::edit");
```

it would be:
```
$router->addGet("/products/edit/{product}", "Products::edit");

```

Which then looks up do we have a model of type "product", if we do lookup the identifier on the model and load it so that we can inject into the action in the controller.

Green-Cat:
Symfony has this too http://symfony.com/doc/current/bundles/SensioFrameworkExtraBundle/annotations/converters.html

andresgutierrez:
https://docs.phalconphp.com/en/latest/reference/routing.html#using-conversors

mattvb91:
@andresgutierrez thanks for the link I will have a play around with this. If this works the way I think it does I will open a pull request to update the docs with an example with a model.

mattvb91:
wow that is awesome. Thats a hidden gem right there.

So this works perfectly:

```
            $router->add('/user/{user}', [
                'controller' => 'index',
                'action'     => 'test',
            ])->convert('user', function ($id) {
                return User::find($id)->getFirst();
            });
```

and in the controller:

```
 public function testAction(User $user)
 {
    //bla
 }
```

Now I guess the next question is, what would be involved in getting that to register/check for matching models automatically instead of having to manually do that for every route?

Jurigag:
I guess in dispatcher it has to get type of parameters from action. If its class which extends model, then it could do this behaviour you want maybe ? But idk.

spuy767:
@mattvb91 Perhaps you were just shorthanding it in your demo code, but if you use User::findFirst($id) you get to skip a step.

mattvb91:
@spuy767 thanks yea I have updated documentation with findFirst().

@Jurigag thanks for the suggestions I will look into that.

mattvb91:
@Jurigag I have managed to make it work in the dispatcher by using the following:

This is just thrown together to make it work haven't refined anything yet.

```php
$reflectionMethod = new \ReflectionMethod($handlerClass, $actionMethod);
[...]

PR #11288, opened 2016-01-09
Load model in dispatcher if type hinted in controller action

mattvb91:
Issue: #11267 

TODO:
- [x] Update changelog
- [x] Implement tests - Can someone give some advice?

Just a few other notes that I need some help with:

Is there a zephir equivalent of Class::class to get a classname? I dont like the way I have it here defining the classes in strings?

Not sure about the naming of "BindModelInterface" under Phalcon\Mvc\Controller. Any suggestions?

Can someone give me an idea what the best approach is for testing this? Or point out an already existing test class that I can add into.

---

This allows for typehinting models in the controller actions which will then be loaded for you in the dispatcher:

``` php
    public function viewAction(Article $article)
    {
        //$article is already instantiated
    }
```

It also adds a  \Phalcon\Mvc\Controller\BindModelInterface which allows you to set the controllers associated model:

``` php
static function getModelName()
{
     return Article::class;
}
```

This is useful in cases where you have a BaseController that handles a certain action for you but still want to be able to have the model loaded:

BaseController:

``` php
public function viewAction(Model $model)
{
      //Will check the child controller to "getModelName()" to load the correct entity
}
```

Currently this will only work based off /route/action/{id}. Something for a next release would be to allow modifying the table column that it performs the select on instead of ID.


Jurigag:
Test case:
Create some route, handle it, and check if parameter in controller is instanceof class ?

Is'nt SomeClass::class a constant ? Maybe try constant("Model::class") but its still doesnt look nice.

sergeyklay:
`SomeClass::class` is php 5.5 syntax. We still work with php 5.4

mattvb91:
Updated change log and added tests. 

If I do a "git merge upstream/2.1.x" so I can resolve the conflict that wont mess up this pull request will it?

Jurigag:
Yea, just pull changes from upstream/2.1.x and merge them.

mattvb91:
@sergeyklay can you restart the tests for this? Looks like composer failed?

Jurigag:
You can restart it yourself by close and open PR.

unit-tests/controllers/Test9Controller.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+use Phalcon\Mvc\Model;
+
+class Test9Controller extends Phalcon\Mvc\Controller implements Phalcon\Mvc\Controller\BindModelInterface
+{
+	public static function getModelName()
+	{
+		return People::class;
```
sergeyklay: We still use PHP 5.4

mattvb91:
Thanks have fixed but why would that cause tests to fail on 5.5 and 5.6? 

Will squash it all once I can get travis to pass

sergeyklay:
Did you tested/compelled locally?

Jurigag:
I have no idea whats going on with your travis build. You have segmntation fault and some tests are incomplete/skipped where you didnt even edit anything.

mattvb91:
@sergeyklay yes I always compile locally and run tests here before pushing. Although I only have 5.6 available to test on at the moment I will look into getting 5.5 and 5.4 too.

I need to spend a bit more time looking into travis as im not sure what im looking for in the output.

Jurigag:
I got exactly the same errors here: https://github.com/phalcon/cphalcon/pull/11306

mattvb91:
Yea looks like for the 5.6 log its lines 196 and 227 that are the only ones that have "install failed". I wonder if travis sees that as the reason for returning a 1 at the end.

Jurigag:
Not only 5.6, 5.5 and 5.4 have problems as well, i dont know why.

Jurigag:
@sergeyklay 
Checked on locally machine, have segmentation fault as well. In syslog:
kernel: [3919505.262140] phpunit[17151]: segfault at 14 ip 00007ff63a5dbf18 sp 00007ffecf767930 error 6 in phalcon.so[7ff63a2bd000+38b000]
I can send you a crash file if needed.

sergeyklay:
Looks good to me. @andresgutierrez ?

andresgutierrez:
Make a ReflectionMethod on every request could be slower. Can we make this optional?

mattvb91:
Good point! What about a new method/parameter on dispatcher to enable/disable it?

```php
$dispatcher->setModelBinding(true);
```

So then I can do a 

```
if this->_modelBinding {
 //Code here
}
```

mattvb91:
@andresgutierrez see a51aeac for making it optional.

SidRoberts:
With regards to "Something for a next release would be to allow modifying the table column that it performs the select on instead of ID", you could get an array of the dispatcher parameters and whitelist them against the columns in the model with http://php.net/manual/en/function.array-diff.php. You could then do a findFirst and bind those parameters. Once I get home, I'd be happy to provide a proof-of-concept if it's not clear what I mean.

mattvb91:
Thanks for the feedback @SidRoberts I have updated to change the naming as you suggested, I agree it makes it alot clearer.

:+1: for proof of concept, the thing that I am not sure about is how to decide which param is the correct one to choose if multiple get passed in that match. So just adds alot of complexity from the get go. I would like to have that functionality too though, maybe even a new model interface that if implemented gives you access to intercept the parameter/input and transform it before continuing?

Jurigag:
Shouldnt we got getColumnMap from certain model from models metadata and then check if parameter exists as key in this array ?

SidRoberts:
@Jurigag yes, exactly that.

@mattvb91 I'd argue that if model binding is enabled and the action is expecting a model, then it is up to the developer to choose parameter names that do not interfere with that.

phalcon/dispatcher.zep
```diff
@@ -305,6 +307,16 @@ abstract class Dispatcher implements DispatcherInterface, InjectionAwareInterfac
 	}
 
 	/**
+	 * Enable/Disable model binding during dispatch
+	 *
+	 * @param boolean value
```
sergeyklay: `@param boolean value` not needed here. It will be generated automatically

SidRoberts:
OK, so I've been thinking about this PR and #11308 all night long. :stuck_out_tongue: I'm not thinking very clearly so consider this some crazy guy's opinion until somebody else agrees with me. :wink: 

I personally don't think we should be interacting with dispatcher parameters via the action method's parameters (`public function viewAction($param1, $param2)`) - `$this->dispatcher->getParam()` should be used instead. If we didn't have to worry about compatibility, I'd be happy to disable the method parameters entirely.

I think getting model instances automatically from a route is a fantastic idea and one that Phalcon should have. However, I'm averse to changing the format of method parameters (which I believe to be already flawed) to achieve it. I'd also like to avoid using Reflection if at all possible.

I've come up with a slightly different take on this idea in the form of a new DI service:

```php
<?php

namespace Sid;

class BoundModels extends \Phalcon\Mvc\User\Plugin
{
    /**
     * @param string     $className
     * @param array|null $acceptableAttributes
     *
     * @return \Phalcon\Mvc\ModelInterface|false
     */
    public function get($className, $acceptableAttributes = null)
    {
        $model = new $className();

        if (!$acceptableAttributes) {
            $dispatcherParams = array_keys($this->dispatcher->getParams());
            $modelAttributes  = $this->modelsMetadata->getAttributes($model);

            $acceptableAttributes = array_intersect($dispatcherParams, $modelAttributes);
        }

        $conditions = [];
        $bind       = [];

        foreach ($acceptableAttributes as $attribute) {
            $conditions[]     = $attribute . " = :" . $attribute . ":";
            $bind[$attribute] = $this->dispatcher->getParam($attribute);
        }

        $conditions = implode(" AND ", $conditions);

        $boundModel = call_user_func_array(
            [$className, "findFirst"],
            [
                [
                    "conditions" => $conditions,
                    "bind"       => $bind
                ]
            ]
        );

        return $boundModel;
    }
}
```

```php
$di->set(
    "boundModels",
    function () {
        $boundModels = new \Sid\BoundModels();

        return $boundModels;
    }
);
```


```php
<?php

namespace Sid\Models;

class Posts extends \Phalcon\Mvc\Model
{
    public $categorySlug;
    public $postSlug;

    // ...
}
```

```php
<?php

namespace Sid\Controllers;

/**
 * @RoutePrefix("/post")
 */
class PostController extends \Phalcon\Mvc\Controller
{
    /**
     * @Route("/{categorySlug:[A-Za-z0-9\-]+}/{postSlug:[A-Za-z0-9\-]+}")
     */
    public function postAction()
    {
        // Infer attributes
        $post = $this->boundModels->get("Sid\\Models\\Posts");

        // Force only categorySlug to be used to find the model
        $post = $this->boundModels->get("Sid\\Models\\Posts", ["categorySlug"]);

        // ...
    }
}
```

mattvb91:
I do like the idea of being able to hook into it separately outside of the dispatcher. However personally if the end result is still having to do:

```php
 $post = $this->boundModels->get("Sid\\Models\\Posts");
```

instead of 
```php
function postAction(Posts $post)
```

Then I do think its nearly the same result as just simply doing a findFirst() ? 
I wonder if we could create a new event "beforeDispatch" or something and then hook something in like that? Personally I was looking for the Laravel style loading passed into the function instead of then having to fetch it again inside.

mattvb91:
Anyone else have an opinion regarding the above?

Jurigag:
I think di service is just some thing you can do always. To be honest its even mean thay you still have to do everything by yourself, i like original idea more, but its good idea too that it wouldnt necessary create reflectionmethod each time on each request. Maybe go back to interface - and make it necessary to implement it if we want this behaviour ? I guess its faster to check that object is instanceof something instead of creating reflection and check type of parameters etc.

mattvb91:
Agreed it would be nice to not have to create a reflection method each time. Personally I would be happy with it if it was necessary to implement the interface for it to work in order to get rid of the reflectionmethod. Although it is a pretty slick looking controller when you can just typehint the model without having to implement the getModelName method.

Another thing to think about: Like I said in my initial RFC issue personally I only use 1 controller for 1 model so the above will work fine for me. However what about someone who uses multiple models within a controller and would like to typehint different models on different functions? This wouldnt work without the reflection method.

If people are worried about that reflectionmethod on every request, what about taking it even one step further and instead of on the dispatcher having the setModelBinding(true) we instead have a method per controller? That way ONLY if the controller specifically returns true for the setModelBinding method only then will it use the reflectionMethod. Any other controller that doesnt setModelBinding(true) will just skip over it?

Green-Cat:
How significant is the performance impact from the reflection method anyway?

mattvb91:
I wouldn't even know how to test that through zephir? If it was being used in a loop of some kind I would be worried about some performance hit but it would only be used once during the request to get to the controller.

mattvb91:
@andresgutierrez any feedback since making it optional?

andresgutierrez:
Thanks

sergeyklay:
@mattvb91 [The docs](https://github.com/phalcon/docs) should probably be updated too. Can you please create PR at least for English?

mattvb91:
@sergeyklay yes was planning on doing that! Will have it in this weekend!

sergeyklay:
@mattvb91 You have to note, this feature is available from 2.1.x

daison12006013:
Is this more likely a method injection? this is cool similar with other frameworks :+1:

sergeyklay:
@mattvb91 @daison12006013 We understand the similarities with what framework more just  :wink:

Jurigag:
@mattvb91 

Could you possibly add new event - ```afterBinding``` to dispatcher and method like ```getBindedModel()``` ? I'm currently implementing some firewall to phalcon, currently using annotations, but i want to add option to using internally phalcon's ACL, and this binded model will be passed to user defined function in acl(new ability in 2.1.x).

PR #11294, opened 2016-01-11
Updated flash alerts

sergeyklay:


sergeyklay:
I will return to this later

PR #11296, opened 2016-01-12
Fixes for building on windows

ifsnow:
I'm sorry to say that I had some misunderstanding in my previous commit. (#11278)
I realized that config.m4 is required file when installing the extension from source code.

so, I changed the rule to generate config.w32 file on windows only.

Thanks.

### DIFF
--- a/build/_resource/Phalcon/Build/Generator/Safe.php
+++ b/build/_resource/Phalcon/Build/Generator/Safe.php
@@ -37,9 +37,9 @@ class Generator_Safe
     /**
      * Generator for config.m4 or config.w32
      *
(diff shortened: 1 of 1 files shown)


ifsnow:
[+] Here is the way to unset limit. (#11298)

Green-Cat:
Could you please edit the Change log to add the changes you made.

sergeyklay:
Also fix for #11298 should be in a separated commit

ifsnow:
@sergeyklay I've removed "fix for #11298" commit.

sergeyklay:
Thanks

Issue #11298, opened 2016-01-12
Phalcon\Mvc\Model\Query\Builder limit bug


marhub:
There is no way to unset limit.
For example:

```PHP 
$qb = new \Phalcon\Mvc\Model\Query\Builder;
$qb->limit(5);
$qb->limit(null); // expected result is no limit

var_dump($qb->getLimit()); // returns 'int(0)'
```

I belive it's because of casting limit in method to int.

sergeyklay:
Fixed in `2.0.x` branch

PR #11301, opened 2016-01-13
Fixed for Builder bug (#11298)

ifsnow:
I think it's reasonable to solve #11298.

Please let me know if you have another thought.


sergeyklay:
Thanks

PR #11316, opened 2016-01-17
[IGNORE] Models type hinting test

SidRoberts:


mattvb91:
btw just to point out the BindModelInterface is for when you have something like a base class that has 

```php
public function view(Model $model)
```
So that way you can define the actual model to use within the child controller.

Issue #11286, opened 2016-01-08
Phalcon\Mvc\Model __set method failure


bmoore:
After reviewing the documentation, and working with phalcon models. I found an issue using model setters. https://docs.phalconphp.com/en/latest/reference/models.html#public-properties-vs-setters-getters mentions that we can automagically use a method named "set" + camelize(property) to handle setting non-public properties. This however is not what is happening.

Consider the example
```php
class User extends \Phalcon\Mvc\Model
{
    protected $id;
    public function setId($id) {
        echo "Got here.";
        $this->id = $id;
    }
}

...

$user = new User();
$user->save(['id' => 1], ['id']); // This will print "Got here."
...
$user = User::findFirst(1);
$user->id = 13; // This does assign 13 as the id but does not print "Got here."
```

After digging through the code, I noticed that the model magic `__set` method made no check for the possible setter, and instead just assigns the property.

```zephir
phalcon/mvc/model.zep
...
abstract class Model implements EntityInterface, ModelInterface, ResultInterface, InjectionAwareInterface, \Serializable, \JsonSerializable
{
...
    public function __set(string property, value)
    {
        var lowerProperty, related, modelName, manager, lowerKey,
        relation, referencedModel, key, item, dirtyState;
        ...
        /**
         * Fallback assigning the value to the instance
         */
        let this->{property} = value;
        return value;
    }
    ...
```
This makes no mention or use of the possible setter.
Furthurmore second to last line completely bypasses the purpose of making a property private or protected, as it will always set the property regardless.

I can see the line in `Model::assign()` that utilizes the possible setter.
```zephir
	let possibleSetter = "set" . camelize(attributeField);
        if method_exists(this, possibleSetter) {
            this->{possibleSetter}(value);
        } else {
            let this->{attributeField} = value;
        }
```

I believe this should be refactored so the possible setter is utilized in `__set` and also fix the issue with `__set` completely bypassing private and protected properties.

I will provide a proof-of-concept patch in the near future.

bmoore:
Does anyone here have any thoughts with this one? I've been trying to rewrite it possibly using the ReflectionProperty class, but was getting segfault errors from it. Now I am getting real errors on TravisCI though I cannot yet run the unit tests locally.

P.S. Is there a guide somewhere about getting set-up for running test locally?

sergeyklay:
Could you please check now `2.1.x` branch, after accepting #11401

michanismus:
[...]

Issue #11171, opened 2015-11-29
[2.1.x] Inconsistency with Model::assign() when Creating new Model Instances with Array Data


virgofx:
> The following issue is specific to 2.1.x which introduced the ability to create a new `Model()` instance with an array of instance field/value data.

Creating new Model instances with array data can result in data being transformed twice because `Model::assign()` will use internal setters.

This is a common issue when using snapshots or dynamic updates.

For example:

``` php
<?php

class Robot extends \Phalcon\Mvc\Model {

    public $specialField;

    public function initialize() {
        $this->useDynamicUpdate(true);
    }

    public setSpecialField($rawValue) {
       $this->specialField = $rawValue . '-transformed';
    }
}
```

``` php
$robot = new Robot();
$robot->setSpecialField('raw');

// Snapshot data contains the resulting transformed data. 
// $snapshot['specialField'] = 'raw-transformed';
$snapshot = $robot->getSnapshotData();

// Assume, the application needs to recreate this model.
$newRobot = new Robot($snapshotData);

// After construction, the instance data is not accurate since it's been transformed twice.
// $newRobot->specialField = 'raw-transformed-transformed';

// Expected:
// $newRobot->specialField = 'raw-transformed';

```

**Recommendations:**
1. Maintain backwards compatibility and add a `$useModelSetters = true` parameter to `Model::assign()` as the fourth parameter. The model construct should call this with `false`.
   
   https://github.com/phalcon/cphalcon/blob/2.1.x/phalcon/mvc/model.zep#L170
2. Since 2.1.x already  breaks backwards compatibility with Model instantiation, simply remove the setters in `Model::assign()`. 
   
   Not sure why this was ever introduced? For 2.1 this would break backwards compatibility but it should be the correct design. There exists ambiguity when the model arbitrarily looks for extending methods as the various export functions do not check for ambiguous functions (nor should they) (e.g. `toArray()`, `getSnapshotData()`, etc).
   
   https://github.com/phalcon/cphalcon/blob/2.1.x/phalcon/mvc/model.zep,  Lines 493-496, 498

sergeyklay:
@virgofx Could you please check now 2.1.x branch

virgofx:
Will recheck this soon as I get some time ... hopefully within a couple days.

virgofx:
Small note: Model.zep @line 492/493 needs corrected notes

sergeyklay:
@virgofx ping

sergeyklay:
[...]

PR #11320, opened 2016-01-18
Minor rewrite to assign and __set issue #11286

bmoore:
Background: https://github.com/phalcon/cphalcon/issues/11286

Fixes issue with __set setting anything regardless of visibility.

### DIFF
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -490,10 +490,7 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 				}
 
 				// Try to find a possible getter
-				let possibleSetter = "set" . camelize(attributeField);
-				if method_exists(this, possibleSetter) {
-					this->{possibleSetter}(value);
-				} else {
+				if !this->_possibleSetter(attributeField, value) {
 					let this->{attributeField} = value;
 				}
 			}
@@ -4106,7 +4103,7 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 	public function __set(string property, value)
 	{
 		var lowerProperty, related, modelName, manager, lowerKey,
-			relation, referencedModel, key, item, dirtyState, method;
+			relation, referencedModel, key, item, dirtyState;
 
 		/**
 		 * Values are probably relationships if they are objects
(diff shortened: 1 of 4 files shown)


sergeyklay:
This PR fixes #11286 ?

bmoore:
@sergeyklay This PR is an attempt to solve #11286 but it is failing some Model tests (I cannot find any details about the failed tests). I would like to track down what the issues are, but I have yet to get phpunit running locally.

virgofx:
This PR should also address issues with assign using **set** with raw data when it may not be recommended. See #11171.

sergeyklay:
@bmoore Could you please check now `2.1.x` branch, after accepting #11401

bmoore:
@sergeyklay This handles part of the fix where the setter is used in __set.

Unfortunately, this does not handle the main point of contention, which is the last assignment statement in the __set method.

By having `let this->{property} = value;` you bypass the purpose of the property visibility, mainly protected variables without setters are just set, instead of throwing an illegal access error. Another thing, given the way php behaves, is you can just begin attaching random properties to models...

If you have a `User` model that has an `email` and `password`, you can also write `$this->foobar = 'banana';` and your user instance will now have a public `foobar` property.

I believe the appropriate fix is to just remove that last assignment line, as it should be unnecessary. I will rebase from 2.1.x and try to remove the line, but my suspicion is that the tests will fail because that assignment is missing. With the tests failing because of that, I'm afraid that the tests may be inadequate, depending on a direct assignment in the __set method, as it should only be called if a direct assignment is called on a private or protected property.

sergeyklay:
As you can see tests fails.

bmoore:
@sergeyklay I predicted that the tests will fail. This is a problem, as the design should not depend on direct assignment in the `Model::__set` method. Currently, it seems the design does depend on this piece of code, which unfortunately bypasses the notion of property visibility. 

Looking through the `Robots` class in the unit-tests folder. There are no properties defined for the `Robots` model, which is why the tests are failing, because the direct assignment isn't dynamically creating the property in `Model::__set` when `instance->{attributeName} = castValue;` is declared in `Model::cloneResultMap`.

This is only a problem if you want to correctly use these fundamentals of classical architecture in the model classes.

If that does not matter, then we can close this issue, and just look the other way.

bmoore:
This latest commit will attempt to use the reflection class again, and wraps testing the reflection property with a try catch handle when a property isn't defined. Therefore if the visibility explicitly set to protected or private, it will not directly assign. Otherwise, it will, which leaves the ambiguous case of undefined properties being run-time defined as public, which should handle the case of the Robots class not defining any properties.

bmoore:
There were 2 errors:

1) ConfigTest::testYamlConfig
Phalcon\Config\Exception: Yaml extension not loaded

phalcon/config/adapter/yaml.zep:71
/home/travis/build/phalcon/cphalcon/unit-tests/ConfigTest.php:296

2) ConfigTest::testYamlConfigCallback
Phalcon\Config\Exception: Yaml extension not loaded

phalcon/config/adapter/yaml.zep:71
/home/travis/build/phalcon/cphalcon/unit-tests/ConfigTest.php:310

In https://travis-ci.org/phalcon/cphalcon/jobs/109072750 which looks unrelated to what I've done.

sergeyklay:
Well, all ok. Could you please create a small test which will cower your PR? Also please update `CHANGELOG.md`

bmoore:
@sergeyklay please review.

phalcon/mvc/model.zep
```diff
@@ -4154,24 +4151,66 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
...
+		//Try reflection on the property.
+		let reflectionClass = new \ReflectionClass(this);
+		try {
+			let reflectionProp = reflectionClass->getProperty(property);
+			if !reflectionProp->isPublic() {
+				return false;
+			}
+		} catch \Exception, e {
+			// The property doesn't exist.
+			return true;
```
sergeyklay: We shouldn't return `false`  here?

bmoore: @sergeyklay the reason this is return true, is because it looks like phalcon models depend on runtime declaration if public properties. In the unit-test model Robots, there are no column properties defined.

If this was false, then the tests would fail, because   undeclared properties would not be set, and the robots model would not get fields set on load. Furthermore, if it were to return false, then the whole function would be redundant and we would have the same functionality removing the final assignment call in __set.

By catching the thrown exception and returning true, we're saying, "if there is no property, make one and assume it's public."

Is there a third party to review this pull request to ensure it is the direction phalcon wants to go with this?

bmoore: @phalcon ^^

bmoore:
Also, this may handle the issues cited in https://github.com/phalcon/cphalcon/issues/3033

phalcon/mvc/model.zep
```diff
@@ -4154,24 +4151,66 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
...
+	/**
+	 * Check whether a property is declared private or protected.
+	 * This is a stop-gap because we do not want to have to declare all properties.
+	 *
+	 * @param string property
+	 * @return boolean
+	 */
+	protected final function _isVisible(property)
+	{
+        var reflectionClass, reflectionProp, e;
```
niden: Is this just how github shows code or are we missing a tab here?

andresgutierrez:
Can you please rebase?

sergeyklay:
Thanks

Issue #11117, opened 2015-11-10
Model has a small possibility because of replication latency.


ifsnow:
I think it's a low priority issue.
but, we should know the possibility of an unintended result.

_exists() function always executes before create() or save() function in Model.

```
_exists() function => use read connection.
save() function => use write connection.
```

There may be long latency to replicate between master and slave servers.
We can't be guaranteed that slave servers always are same as master server.
It could be a bit problematic in high transaction.

In my opinion, you'd better use write connection intead of read connection in _exists() function.

Mechzeit:
Logistically it would make more sense to override the Model methods 'create'/'save' to temporarily change the read connection service for these methods.

Altering the code base to account for anyone running a master-slave database infrastructure by 'reading' from the 'write' DB service doesn't make the most sense out of all possible solutions, in my opinion.

Based on the code at https://github.com/phalcon/cphalcon/blob/c8d4916bbfd0cb18395fe25d8c5a663fc5077d8f/phalcon/mvc/model.zep#L2959
You could do something like

```php
namespace Custom;

class Model extends \Phalcon\Mvc\Model {
    public function create($data = NULL, $whiteList = NULL)
    {
        $originalReadService = $this->getReadConnectionService();
        $this->setReadConnectionService($this->getWriteConnectionService());
        $result = parent::create($data, $whiteList);
        $this->setReadConnectionService($originalReadService);
        return $result;
    }
}

dschissler:
@sergeyklay Can we add this to the 4.0 milestone?  We can evaluate it later for its validity.  It seems though that it could eliminate some rare gnarly impossible to reproduce race conditions.

guweigang:
Maybe we should not check if record exists, it's not just a mater-slave latency problem. Even use the same connection here, unless you use ```select...for update``` under a transaction，or else you can not avoid this  'cause of concurrency.

_Sent from my OnePlus ONEPLUS A5000 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

acmenna:
Master-slave setup does not work with transactions either. If a transaction is active, all reads should be kept in master regardless the read connection.  Doctrine has some cool rules to handle all these scenarios mentioned in this issue. I marked in bold the most important one that will covered the latency and the transaction issue.

http://www.doctrine-project.org/api/dbal/2.0/class-Doctrine.DBAL.Connections.MasterSlaveConnection.html

[...]

Issue #11106, opened 2015-11-09
Model has unnecessary select queries in specific case.


ifsnow:
I'm always grateful for phalcon :)

I wish it would be improved at some future day. Thanks you.

---

For example, robot table has multiple primary keys and auto_increment column.

``` sql
CREATE TABLE `robot` (
    `seq` INT(11) NOT NULL AUTO_INCREMENT,
    `sub_key` INT(11) NOT NULL,
    `name` VARCHAR(100) NULL DEFAULT NULL,
    PRIMARY KEY (`seq`, `sub_key`)
);
```

create() function generates 3 quries. I think 2 select quires are unnecessary in this case.

``` php
$robot = new Robot();
$robot->sub_key = 1;
$robot->name = "Robot 1";
$robot->create();
```

``` sql
call _exists() function of create() function.
SELECT COUNT(*) "rowcount" FROM `robot` WHERE `seq` = NULL AND `sub_key` = 1;

call _exists() function of save() function.
SELECT COUNT(*) "rowcount" FROM `robot` WHERE `seq` = NULL AND `sub_key` = 1;

INSERT INTO...
```

Parameters are checked about null already. but, It needs more consideration.

If type of filed is AUTO_INCREMENT(or NOT NULL) and value is null, I think _exists() function should return false.

sergeyklay:
I'm closing this issue due to the lack of any reaction. I'll open it again if the need arises

PR #11321, opened 2016-01-19
Improves the logic of `Model`

ifsnow:
I think there're some issues to work with `Model` in my project.
so, I tried to improve `Model` for me. I hope it would be valuable to others.

Please let me know If you have an another opinion or idea.
- It's related to #11106, #11117
## 1. Only use write connection on save.

I usually prefer write(master) connection rather than read(slave) connection while the processes to save data. That's because it's possibility to have unexpected network latency between master and slave servers.

I just want to use write connection selectively.
- global config

```
phalcon.orm.use_write_connection_on_save = On (default : On)
```
- example for Model::setup()

```
Model::setup(array(
    'useWriteConnectionOnSave' => false
));
```
## 2. Not check if the record exists.

I don't want to check the existing record before create or update function.
It's just for performance. and, I think it's often unnecessary in most cases.
- global config

```
phalcon.orm.check_existing_record_before_save = Off (default : Off)
```
- example for Model::setup()

```
Model::setup(array(
    'checkExistingRecordBeforeSave' => true
));
```
## 3. Fixed for checking the unique key.

Sometimes, `Model` has unnecessary select queries in specific case. (#11106)
If the field is AUTO_INCREMENT and NULL, It would not occur additional select query.

### DIFF
--- a/config.json
+++ b/config.json
@@ -96,6 +96,14 @@
         "orm.ignore_unknown_columns": {
             "type": "bool",
             "default": false
+        },
+        "orm.check_existing_record_before_save": {
+            "type": "bool",
+            "default": false
+        },
+        "orm.use_write_connection_on_save": {
+            "type": "bool",
+            "default": true
         }
     },
     "destructors": {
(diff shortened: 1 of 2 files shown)


wedgwood:
Agree with @ifsnow. I have same problem. Additional select query is totally not necessary, exceptions like `duplicate entry 'balabala' ...` is enough.

Green-Cat:
:+1:

Green-Cat:
:+1: 
Just two comments on the pull request:
1) Please update the changelog file so that users of the framework can find out about the two new config vars.
2) This should probably target the 2.1.x branch since it does change the behaviour of Model::save()

PR #11326, opened 2016-01-19
Array of status codes is no longer a property in Http\Response

SidRoberts:
Adding to #11325, the `statusCodes` array shouldn't be a property only assigned on certain conditions.

(I've built this on top of @sergeyklay's PR to avoid a merge conflict).


sergeyklay:
Looks good to me. @andresgutierrez?

andresgutierrez:
Thanks

Kachit:
Maybe move status codes to other class for example:
https://github.com/phalcon/incubator/blob/master/Library/Phalcon/Http/Response/StatusCode.php

sergeyklay:
@Kachit https://github.com/phalcon/cphalcon/pull/10129

PR #11328, opened 2016-01-21
Getter for actionSuffix in dispatcher

renatomjr:


sergeyklay:
This breaks BC and should be sent to 2.1.x branch

PR #11342, opened 2016-01-25
Modify paginators case style

r3jack:


sergeyklay:
Could you please send this PR onto `2.1.x`? Also it should be rebased onto `2.1.x`

PR #11344, opened 2016-01-25
Paginator names

r3jack:


sergeyklay:
Could you please submit just the commits that belong to the PR? Also, as you can see https://travis-ci.org/phalcon/cphalcon/jobs/104604163#L4275-L4301 tests fails

PR #11355, opened 2016-01-31
Add metaDataFileMode option to Phalcon\Mvc\Model\Metadata\Files

ariarijp:


sergeyklay:
Can you please explain a bit more  what issue can be solved (or what feature can be introduced) with your PR?

ariarijp:
In my case, I'm running Web Application and CLI Task in one server.

The Web server run as "www-data" user, but CLI Task run as other user(e.g. "ubuntu" user).

If Web App creates Metadata file, I guess CLI Task unable to overwrite Metadata file when Metadata is expired.

Thanks for asking.

ariarijp:
@sergeyklay I think this issue is an edge-case.
Should I close this PR?

sergeyklay:
The 2.0.x branch is not longer supported. Could you please resend PR to the 2.1.x

Issue #11139, opened 2015-11-15
[fix] please add support for enum in db/column


Dublerq:
Right now phalcon sees enums as char(1)

andresgutierrez:
Could you please submit this in the devtools repo? https://github.com/phalcon/phalcon-devtools

Dublerq:
Well... I did. The problem is in Phalcon though. 

the type $_type = 5 => it's recognized as char

[code]class Phalcon\Db\Column#34 (16) {
  protected $_name =>
  string(4) "type"
  protected $_schemaName =>
  NULL
  protected $_type =>
  int(5)
  protected $_typeReference =>
  int(-1)
  protected $_typeValues =>
  NULL
  protected $_isNumeric =>
  bool(false)
  protected $_size =>
  int(0)
  protected $_scale =>
  int(0)
  protected $_default =>
  NULL
  protected $_unsigned =>
  bool(false)
  protected $_notNull =>
  bool(true)
  protected $_primary =>
  bool(false)
  protected $_autoIncrement =>
  bool(false)
  protected $_first =>
  bool(false)
  protected $_after =>
  string(2) "id"
  protected $_bindType =>
  int(2)
}[/code]

afk4life:
I can confirm an issue with enum. It will not save with a field of type enum. In one method it simply ignores the set value, in another it will not save a field enum, responding that the field is required even though it is set.

Jurigag:
Weird, on mysql its working for me with enum. While saving i putting just a number, while getting i got a string from enum, not a number.

afk4life:
It isn't consistent. There is one that was working just fine as enum, but when I saw the others having issues I changed all my enums to varchar to be safe.

Dublerq:
@sergeyklay A year has passed, is there a possibility to have enums anytime soon?

Dublerq:
@Jurigag @sergeyklay 
It doesn't seem like a big deal to implement...

Issue #10216, opened 2015-05-01
[NFR] Add ENUM to column types


ykurtbas:
``` php
class Column implements \Phalcon\Db\ColumnInterface {
        const TYPE_INTEGER = 0;
        const TYPE_DATE = 1;
        const TYPE_VARCHAR = 2;
        const TYPE_DECIMAL = 3;
        const TYPE_DATETIME = 4;
        const TYPE_CHAR = 5;
        const TYPE_TEXT = 6;
        const TYPE_FLOAT = 7;
        const TYPE_BOOLEAN = 8;
        const TYPE_DOUBLE = 9;
}
```

We would appreciate if you added enum into the metadatareader so we could operate on automated database schema.

Currently we have nice handles for all the types and have edge case for the ENUMs,

Thank you.

Green-Cat:
ref #11368

PR #11368, opened 2016-02-04
adding TYPE_ENUM to Phalcon\Db\Column

mattvb91:
PR to devtools will follow

### DIFF
--- a/phalcon/db/adapter/pdo/mysql.zep
+++ b/phalcon/db/adapter/pdo/mysql.zep
@@ -190,6 +190,14 @@ class Mysql extends PdoAdapter
 				let definition["type"] = Column::TYPE_BLOB;
 			} else {
 				/**
+				 * Enum
+				 */
+				if memstr(columnType, "enum") {
+					let definition["type"] = Column::TYPE_ENUM;
+					break;
+				}
+
+				/**
 				 * By default is string
 				 */
 				let definition["type"] = Column::TYPE_VARCHAR;
@@ -200,7 +208,10 @@ class Mysql extends PdoAdapter
 			 */
 			if memstr(columnType, "(") {
 				let matches = null;
-				if preg_match(sizePattern, columnType, matches) {
+
+				if definition["type"] == Column::TYPE_ENUM {
+					let definition["size"] = substr(columnType, 5, -1);
+				} elseif preg_match(sizePattern, columnType, matches) {
 					if fetch matchOne, matches[1] {
 						let definition["size"] = (int) matchOne;
 					}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
-# [3.2.1](https://github.com/phalcon/cphalcon/releases/tag/v3.2.1) (2017-XX-XX)
+# [4.0.0](https://github.com/phalcon/cphalcon/releases/tag/v4.0.0) (2017-XX-XX)
+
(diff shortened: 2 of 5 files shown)


sergeyklay:
Hi @mattvb91 
What about other databases?
If someone moves from MySQL to Postgresql what would happen with those enum types.

In my opinion this PR this PR not completed.

We have to support that type virtually for the other database engines supported.

mattvb91:
@sergeyklay do you mean just defaulting it to something like varchar for all the other adapters?

sergeyklay:
As I said above we have to support that type for the other database. For example we can transparent convert from Myslq's enum to Postgres' varchar. I'm not saying what you need to do exactly. I only express my opinion. Perhaps you'll propose a better idea

mattvb91:
@sergeyklay all the adapters default to either varchar or text if a column type isn't supported which in this case would still be ok as the result would still allow insertion of the data into the column.

sergeyklay:
And back?

mattvb91:
Im not sure if im missing something in your question but if porting data back INTO mysql from a data source not supporting enum it would also default the column to varchar.

Dublerq:
Please consider merging this pull request to 2.x. We're waiting for enums since 2014

phalcon/db/adapter/pdo/mysql.zep
```diff
@@ -280,7 +280,10 @@ class Mysql extends PdoAdapter implements AdapterInterface
 			 */
 			if memstr(columnType, "(") {
 				let matches = null;
-				if preg_match(sizePattern, columnType, matches) {
+
+				if definition["type"] == Column::TYPE_ENUM {
+					let definition["size"] = substr(columnType, 5, -1);
```
mattvb91: @sergeyklay this is where we were trying to use a regex but its easier to simply get the contents out of enum(CONTENTS) by doing a substr select. 

So essentially if you have ENUM('val1', 'val2', 'val3') then definition["size"] = "'val1', 'val2', 'val3'"

Does that make more sense?

sergeyklay:
#10216 #11139

sergeyklay:
@mattvb91

dcosta-ptc:
Any news regarding this topic?

sergeyklay:
@dcosta-ptc Will be part of 3.1.x

Dublerq:
I need to find some time to configure zephir and learn how to test/contribute. Looks like I'll need to speed up needed features development myself because waiting more than half a year to get a simple feature or fix is a bit too long.

sergeyklay:
@Dublerq souds fine!

mattvb91:
@sergeyklay @dcosta-ptc @Dublerq sorry guys im not sure why im only now receiving notifications about this... I've rebased the branch and repushed. Looks like your builds are failing? 

Surprised this hadn't been merged 2 years ago as its a pretty minor change from code side.

sergeyklay:
@mattvb91 I'll take a look a bit later. Thank you for contributing.

jesuGMZ:
3.2.1 and problems too. Any plan when it could be?

aat2703:
4.0.0 according to Sergey... Not usable without ENUM type for me... sadly... Migrations are pretty rough atm...

ruudboon:
Hoping that this and some other types like TIME will get in at 4.0

niden:
@mattvb91 Could you be so kind as to resolve the conflict with the changelog and have a look at the tests? It appears there is a syntax error somewhere. After that I can merge this.

I will need to add more tests on this one

niden:
Thank you!

PR #11373, opened 2016-02-05
Removed pointless property from Volt compiler

SidRoberts:


phalcon/mvc/view/engine/volt/compiler.zep
```diff
@@ -480,38 +478,33 @@ class Compiler implements InjectionAwareInterface
...
-						"password_field": true,
-						"text_area": true,
-						"text_field": true,
-						"email_field": true,
-						"date_field": true,
-						"tel_field": true,
-						"numeric_field": true,
-						"image_input": true
-					];
-					let this->_arrayHelpers = arrayHelpers;
```
sergeyklay: You removed static cache

PR #11376, opened 2016-02-06
Added Phalcon\Dispatcher::hasParam()

SidRoberts:


sergeyklay:
Thanks

phalcon/dispatcher.zep
```diff
@@ -279,6 +279,17 @@ abstract class Dispatcher implements DispatcherInterface, InjectionAwareInterfac
 	}
 
 	/**
+	 * Check if a param exists
+	 *
+	 * @param  mixed param
```
aleksandrzen: `string`?

sergeyklay: How about `int`?

Issue #11372, opened 2016-02-05
[NFR] Query Builder "having" like "where" functions 


michanismus:
Are there reasons why the builder does not have functions like "andHaving", "orHaving", "inHaving" and "notInHaving"?

The function accepts a string only. It could be compiled like where conditions...

It would be possible to add having conditions after calling events where a builder is passed, for example.

michanismus:
Introduced in 3.2.x branch.

PR #11381, opened 2016-02-07
#11372 Added HAVING functions

michanismus:


phalcon/mvc/model/query/builderinterface.zep
```diff
@@ -196,13 +196,65 @@ interface BuilderInterface
...
-	public function having(having);
+    /**
+     * Sets the HAVING conditions
+     *
+     * @param string conditions
+     * @param array bindParams
+     * @param array bindTypes
+     * @return \Phalcon\Mvc\Model\Query\BuilderInterface
+     */
+    public function having(conditions, bindParams = null, bindTypes = null) -> <BuilderInterface>;
```
duythien: Styles code should be to 4 space

michanismus: Strange thing is, I used 4 spaces...

sergeyklay:
1. Provide please small test case for each method
2. Update `CHANGELOG.md`
3. We use tabs in `zep` files for indents

michanismus:
Ok, i will add tests, update CHANGELOG.md and change spaces to tabs

michanismus:
Ok, added fixed files with tabs for indents, a test for each function and I updated CHANGELOG.md.

sergeyklay:
Could you please return back files mode  100644 → 100755.
Also rebase please onto 2.1.x

PR #11388, opened 2016-02-08
Fixed bug in `Http\Request::isValidHttpMethod()`

SidRoberts:
Previously, `$request->isValidHttpMethod("get")` would have returned false.


sergeyklay:
Thanks

PR #11401, opened 2016-02-11
Fix setting model attribute with setter to use setter

nsossonko:
This syncs up the behavior with assign() and setting the value directly and also aligns the behavior of getting the property with setting the property. Comes with a test.

### DIFF
--- a/unit-tests/ModelsGetterSetterTest.php
+++ b/unit-tests/ModelsGetterSetterTest.php
@@ -1,0 +1,114 @@
+<?php
+
+/*
+  +------------------------------------------------------------------------+
+  | Phalcon Framework                                                      |
+  +------------------------------------------------------------------------+
+  | Copyright (c) 2011-2016 Phalcon Team (http://www.phalconphp.com)       |
+  +------------------------------------------------------------------------+
+  | This source file is subject to the New BSD License that is bundled     |
+  | with this package in the file docs/LICENSE.txt.                        |
+  |                                                                        |
+  | If you did not receive a copy of the license and are unable to         |
+  | obtain it through the world-wide-web, please send an email             |
+  | to license@phalconphp.com so we can send you a copy immediately.       |
+  +------------------------------------------------------------------------+
+  | Authors: Nochum Sossonko                                               |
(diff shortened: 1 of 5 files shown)


sergeyklay:
Can you please merge your efforts with #11320 into one PR please. 

Amend CHENGELOG, update tests, explain a bit more what issue can be solved (or what feature can be introduced) with your PR, etc.

Thanks.

nsossonko:
Sorry I didn't even notice that PR. This PR is strictly about looking for a setter when assigning a value directly to a hidden property. That PR solves some other things too.

Basically, Phalcon supports defining getters/setters for model attributes that are called automatically when they are retrieved and/or set. The problem is, when the attribute is set directly it is not called, instead the attribute is assigned whatever value is being passed.

This would fix Issue #11286. Feel free to close if the other PR precedes this one (although feel free to take the test from this one and use in the other one too).

nsossonko:
@sergeyklay what do you think should be done? This change should be completely compatible with 2.0.x and master as well...

sergeyklay:
Can you please rebase?

nsossonko:
@sergeyklay done (I'm pretty sure). Thanks.

unit-tests/ModelsGetterSetterTest.php
```diff
@@ -0,0 +1,114 @@
...
+		}
+
+		$connection = new Phalcon\Db\Adapter\Pdo\Mysql($configMysql);
+
+		$di->set('db', $connection, true);
+
+		$this->_executeSetGet();
+	}
+
+	/*public function testModelsPostgresql()
```
SidRoberts: Why are these commented out?

nsossonko: Because this doesn't really touch the database anyhow (it just checks the getters/setters for the model object).

sergeyklay:

nsossonko:
@sergeyklay done. I messed up with my git rebasing which caused that issue but should be fully fixed and up to date now.

sergeyklay:
Thanks

PR #11403, opened 2016-02-13
Simplified if statement in the Annotations Router.

SidRoberts:


Green-Cat:
:+1: Nice catch

sergeyklay:
Thanks

Issue #11358, opened 2016-02-02
[Phalcon 2.0.9] Radio field unchecked if default value is 0


Jeffzor:
Hello,

Using PHP 5.6.2, when I set a default value to 0, no radio fields are checked.
```php
/*
 Class someClass 
*/
class someClass extends \Phalcon\Mvc\Controller
{
    public function someFunction()
    {
        $this->tag->setDefault('someRadioButton', 0);
    }
}
```

When I set my radio fields in the view:
```php
$this->tag->radioField(array('someRadioButton', 'id' => 'someRadio0', 'value' => 0));
$this->tag->radioField(array('someRadioButton', 'id' => 'someRadio1', 'value' => 1));
```
No radio fields are checked.

It happens just with values 0, '0' and '' (actually all values that are interpreted as false in PHP).

In the tag.zep file, the protected method "_inputFieldChecked", a condition causes the problem (lines 539 and 549):

```
if value && currentValue == value {
				let params["checked"] = "checked";
}
```

```
/**
* Evaluate the value in POST
*/
if value {
    let params["checked"] = "checked";
}
```

The condition "if value" with those values is not working properly (never enters in the condition).
I should be able to create Radio fields with these default values without problem.

It should be like in PHP : 
```php
if (isset($value))
```

It would be great if you could fixed it.

Thanks!
Jeff

bmoore:
https://github.com/phalcon/cphalcon/pull/11412

sergeyklay:
Fixed in `2.1.x`

PR #11412, opened 2016-02-16
Fix issue #11358

bmoore:
Since `value` comes from `Tag::getValue`, which returns `null` if it's not set, then we can test that value is not null, and we should still be able to allow for falsy values.


sergeyklay:
Can you please create a small test that shows that these issues has successfully solved?

bmoore:
@sergeyklay ready for merge.

sergeyklay:
Thanks

aleksandrzen:
@bmoore, thx!

PR #10429, opened 2015-05-28
[2.0.x] setLogLevel method to Multiple Logger

tmihalik:

PR #11418, opened 2016-02-17
[2.0.x] setLogLevel method to Multiple Logger

bmoore:
Port setLogLevel to multiple as was done by tmihkil but gone stale. Add unit tests to verify the default log level is applied across multiple logs.

https://github.com/phalcon/cphalcon/pull/10429 went stale. So I recreated it and added a unit-test.


sergeyklay:
Thanks

PR #11428, opened 2016-02-18
Update postgresql.zep

fs523577192:
Improve describeColumns to make it accord with PostgreSQL.
1. PostgreSQL does not support "tinyint" and does not return the numeric size in the data type column, so the "tinyint(1)" if-branch is useless.
2. PostgreSQL returns "real" instead of "float" when the data type is single-precision floating-point number and returns "double precision" when the data type is double-precision floating-point number.
3. PostgreSQL returns "bit varying" when the data type is variable-length bit-string, so the if-branch for VARCHAR should search for "char" as well as "varying".


sergeyklay:

fs523577192:
@sergeyklay I don't understand what you mean by 

> belong to the PR
Do you mean "belong to 2.0.x" or "not including the unit test file"?

sergeyklay:
I mean you need rebase in order to include to this PR only our commits

PR #11437, opened 2016-02-23
fix modifyColumn

Daredzik:
If currentColumn is missing use column parametr.

prevent:

```
ERROR: RuntimeException Object
(
    [message:protected] => Trying to call method getname on a non-object
...
)
```


sergeyklay:
Can you please also port this to the 2.0.x branch?

Daredzik:
ye sure:
https://github.com/phalcon/cphalcon/pull/11438

PR #11438, opened 2016-02-23
Port from:  fix modifyColumn #11437

Daredzik:
If currentColumn is missing use column parametr.

prevent:

```
ERROR: RuntimeException Object
(
    [message:protected] => Trying to call method getname on a non-object
...
)
```


sergeyklay:
Thanks

PR #11450, opened 2016-02-25
Added a couple tests for \Phalcon\Http\Request

rkeplin:
When looking how these functions were implemented, I noticed that there weren't any tests for them, and figured it would be an easy first contribution.


sergeyklay:
Thanks

PR #11460, opened 2016-02-26
Fix di array access

Jurigag:
This is related to this problem on forum : https://forum.phalconphp.com/discussion/10583/set-service-to-be-not-shared#C30309

I think it adds some consistency for array syntax and option to set if service should be shared or not.

Also i guess i could possibly do PR for 2.0.x.


Jurigag:
So:

1. If service is registered as shared for example "request" accesing it will return shared service. If its not registered as shared it returns new instance like normally get would.
2. You can register shared service using array syntax like this:
$di['someService'] = [function(){ return new SomeService(); },true];
If we register service only with definition by default its shared.

Well im not sure that this code dont break some logic with current apps.

PR #11462, opened 2016-02-27
Add "OR" operator to query builder

dzubchik:
Recently I faced with problem of building complex query and I should do this using query builder. I have not found possibilities to do this, so this is pull-request that does not break compatibility.

### DIFF
--- a/phalcon/mvc/model/query/builderinterface.zep
+++ b/phalcon/mvc/model/query/builderinterface.zep
@@ -26,7 +26,7 @@ namespace Phalcon\Mvc\Model\Query;
  */
 interface BuilderInterface
 {
-    const OPERATOR_OR = "or";
+	const OPERATOR_OR = "or";
 
 	const OPERATOR_AND = "and";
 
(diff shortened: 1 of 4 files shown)


phalcon/mvc/model/query/builderinterface.zep
```diff
@@ -26,6 +26,10 @@ namespace Phalcon\Mvc\Model\Query;
  */
 interface BuilderInterface
 {
+    const OPERATOR_OR = "or";
```
Green-Cat: The indentation is a bit off here.

sergeyklay:
For indents we use tabs in zep files instead of spaces

dzubchik:
@sergeyklay , just a second and I will correct it. I think this possibility should be highlighted in documentation. How can I do this?

sergeyklay:
Looks good to me.

@andresgutierrez

xboston:
:+1:

sergeyklay:
Thank you very much, man

Issue #1299, opened 2013-09-26
[NFR] Multiple memcached server support for memcache clusters


tsnyder91:
I'm building out a distributed memcache system over 3 servers in the near future and would like to support for a distributed memcached system in the memcache cache code natively.  Currently it only supports one host.  

http://www.php.net/manual/en/memcached.addservers.php 
http://www.php.net/manual/en/memcached.addserver.php

Since the constuct is already created with requiring server info, I was thinking of supporting both string and array for host. Similarly for the port, if only one port is given then use that port for all hosts. If it's an array then map the ports to the hosts (it would need to match the count of the number of hosts or throw an exception). 
There should be new support for the weight param found in the docs above as I may one to not do an even distribution among the cluster.
Something like the following.
$cache = new Phalcon\Cache\Backend\Memcache($frontCache, array(
    'host' => array('localhost','localhost2','localhost3')
    'port' => array(11211,11212,11211),
    'weight' => array(40,50,10),
    'persistent' => false
 ));
##

jaskaransingh156:
I fixed this issue by adding a function "addServers" to the Phalcon\Cache\Backend\Memcache class. Code is available here:

https://github.com/jaskaransingh156/cphalcon

Example of using addServers method is following:

$frontCache = new \Phalcon\Cache\Frontend\Data( array ( "lifetime" => 172800 ));
$memcache = new \Phalcon\Cache\Backend\Memcache($frontCache, array ( 'host' => 'memcache_host1', 'port' => 11211,
)); 
$memcache->addServers('memcache_host2', 11211);

Jurigag:
There is `addServers` method in memcache backend, not sure though how it's working exactly.

PR #11474, opened 2016-02-29
Adding addServers method to enable pool of servers for memcache

jaskaransingh156:
Adding addServers method to enable pool of servers for memcache


jaskaransingh156:

https://github.com/phalcon/cphalcon/issues/1299

sergeyklay:
For indents we use tabs in zep files instead of spaces

jaskaransingh156:
Thanks sergeyklay for your feedback. I have corrected the indentation and committed the changes.

phalcon/cache/backend/memcache.zep
```diff
@@ -117,6 +117,28 @@ class Memcache extends Backend implements BackendInterface
...
 
+        /**
+	 * Add servers to memcache pool
+	 *
+	 * @param  string host
+	 * @param  long port
+         * @param  boolean persistent	 
+         * @return boolean
+	 */
+	public function addServers(var host, var port, var persistent = false)
```
sergeyklay: I think should use such declaration:
```zephir
public function addServers(string! host, int port, boolean persistent = false) -> boolean
```

and remove `@param`s doc

See please PHP.net doc:

```php
bool Memcache::addServer ( string $host [, int $port = 11211 [, bool $persistent [, int $weight [, int $timeout [, int $retry_interval [, bool $status [, callable $failure_callback [, int $timeoutms ]]]]]]]] )
```

What with other params?

sergeyklay:
Update please `CHANGELOG.md`

jaskaransingh156:
Thanks @sergeyklay. I have changed declaration of addServers and removed @params doc and also updated CHANGELOG.md file. Regarding other arguments of addServer method, currently only 3 options (host, port and persistent) are being accepted in the constructor of Phalcon\Cache\Backend\Memcache. This host acts as the first server which is added to the memcache pool.  So, if i provide other arguments in addServers method then constructor should be also be changed accordingly in order to maintain consistency among first and all other servers. In that case, instead of using connect method of Memcache library, we should use addServer method to add even the first server. This has been explained in the following comment also:

http://php.net/manual/en/memcache.addserver.php#Hcom104484

Please confirm if constructor of Phalcon\Cache\Backend\Memcache is to be changed so that i will make the changes accordingly.

sergeyklay:
Could you please rebase?

jaskaransingh156:
@sergeyklay Done.

sergeyklay:
Thanks

Issue #11458, opened 2016-02-26
[2.0.10] dispatch:beforeException not executing when exception is thrown in dispatch process


rkeplin:
When an exception is thrown in a dispatch loop event, beforeExecuteRoute for instance, beforeException never gets fired off.

This had been working in Phalcon 1.3.5, until I recently updated to Phalcon 2.0.10.  Not sure if the desired functionality has changed, or if this is a bug.  But I had found it useful when beforeException was fired off whenever an exception was thrown in a controller or other dispatch loop event.

**Reproducible Test**
```php
<?php
use Phalcon\Mvc\View;
use Phalcon\DI\FactoryDefault;
use Phalcon\Mvc\Application;
use Phalcon\Mvc\Controller;
use Phalcon\Mvc\Dispatcher;
use Phalcon\Events\Event;

class IndexController extends Controller
{
    public function indexAction() {
        echo 'This is a test.';
    }
}

class AuthPlugin
{
    public function beforeExecuteRoute(Event $event, Dispatcher $dispatcher, $exception)
    {
        throw new Exception('Catch me in ExceptionPlugin, please!');
    }
}

class ExceptionPlugin
{
    public function beforeException(Event $event, Dispatcher $dispatcher, $exception)
    {
        echo 'I should end up here first.';
    }
}

$di  = new FactoryDefault();

$di->set('dispatcher', function() use($di) {
    $dispatcher = new Dispatcher();

    $eventsManager = $di->getShared('eventsManager');
    $eventsManager->attach('dispatch', new AuthPlugin());
    $eventsManager->attach('dispatch', new ExceptionPlugin());

    $dispatcher->setEventsManager($eventsManager);

    return $dispatcher;
}, true);

$di->set('view', function () {
    $view = new View();

    return $view;
}, true);

$app = new Application();
$app->setDi($di);

try {
    echo $app->handle()->getContent();
} catch (Exception $e) {
    echo 'I should end up here second.';
}
```

**Expected Result**
`I should end up here first. I should end up here second.`

**Observed Result**
`I should end up here second.`

sergeyklay:
Fixed in `2.1.x` branch

PR #11475, opened 2016-02-29
Fires dispatch:beforeException event when there is any exception during dispatching

rkeplin:
Fires dispatch:beforeException event when there is any exception during the dispatch process. #11458


sergeyklay:
Could you please send this PR to the `2.1.x` branch?

PR #11476, opened 2016-02-29
Fires the dispatch:beforeException event when there is any exception during dispatching

rkeplin:
Fires dispatch:beforeException event when there is any exception during the dispatch process. #11458 #11475


sergeyklay:
Could you please rebase onto 2.1.x branch?

sergeyklay:
Thanks

PR #11485, opened 2016-03-03
Use of short syntax in case of compilation of Volt's templates

Studentsov:
Use of syntax of arrays and "echo" in PHP 5.4 style

### DIFF
--- a/phalcon/mvc/view/engine/volt/compiler.zep
+++ b/phalcon/mvc/view/engine/volt/compiler.zep
@@ -514,7 +514,7 @@ class Compiler implements InjectionAwareInterface
 				}
 
 				if isset arrayHelpers[name] {
-					return "$this->tag->" . method . "(array(" . arguments . "))";
+					return "$this->tag->" . method . "([" . arguments . "])";
 				}
 				return "$this->tag->" . method . "(" . arguments . ")";
 			}
@@ -563,7 +563,7 @@ class Compiler implements InjectionAwareInterface
 			/**
 			 * By default it tries to call a macro
 			 */
-			return "$this->callMacro('" . name . "', array(" . arguments . "))";
+			return "$this->callMacro('" . name . "', [" . arguments . "])";
 		}
 
 		return this->expression(nameExpr) . "(" . arguments . ")";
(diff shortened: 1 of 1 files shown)


phalcon/mvc/view/engine/volt/compiler.zep
```diff
@@ -1779,10 +1779,10 @@ class Compiler implements InjectionAwareInterface
 		 * Echo statement
 		 */
 		if this->_autoescape {
-			return "<?php echo $this->escaper->escapeHtml(" . exprCode . "); ?>";
+			return "<?= $this->escaper->escapeHtml(" . exprCode . ") ?>";
 		}
 
-		return "<?php echo " . exprCode . "; ?>";
+		return "<?= " . exprCode . " ?>";
```
sergeyklay: What exactly Phalcon should do here if short syntax is disabled?

sergeyklay:
As you can see, tests fails.

Studentsov:
http://php.net/manual/en/migration54.new-features.php
`<?= is now always available, regardless of the short_open_tag php.ini option. `
Conditions of tests need to be changed, I didn't consider it.

sergeyklay:
Really?

```sh
$ php -i | grep short
short_open_tag => Off => Off
```

sergeyklay:
Also user/hoster/provider/someone else can disble it

Studentsov:
Use of this syntax doesn't depend now on settings in any way. Above I gave the reference to official documentation.

sergeyklay:
Anyway I don't understand what issue aims to solve your PR.

Studentsov:
Only abbreviation of quantity of a code on an output from the compiler and coercion to PHP 5.4 style

phalcon/mvc/view/engine/volt/compiler.zep
sergeyklay: what issue will be solved by removing the semicolon?

Studentsov: Existence or absence of a semicolon doesn't influence behavior in this case. Moreover, in official documentation such option of syntax is given in such look: http://php.net/manual/en/function.echo.php

PR #11490, opened 2016-03-04
Use of short syntax in case of compilation of Volt's templates

Studentsov:
Use of syntax of arrays and "echo" in PHP 5.4 style.
This pull request it is created for the purpose of the general coercion of a code to new style, besides, it reduces the size of the output file a little.
Official documentation: http://php.net/manual/en/migration54.new-features.php


sergeyklay:
Thanks

PR #11495, opened 2016-03-05
Fixed Logger\FormatterInterface inheritance

SidRoberts:
Now, a class that extends `\Phalcon\Logger\Formatter` is forced to implement methods in the interface.


sergeyklay:
Thanks

PR #11496, opened 2016-03-05
Removed redundant validity checks DI and ModelsManager in Mvc\Model.

SidRoberts:
The DI and the Models Manager are set in the constructor so we don't need to check if they are objects. See https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/model.zep#L134 and https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/model.zep#L145.


sergeyklay:
Thanks

Issue #11491, opened 2016-03-04
[NFR] Add setLastModified to Phalcon\Http\Response


ruudboon:
You can set the expire time using the public method setExpires (DateTime $datetime).
Can we add a setLastModified also?

sergeyklay:
Implemented in `2.1.x` branch

ruudboon:
👍

akashkumardce:
@jaskaransingh156 - Very Helpful, looking for it since long time. :)

PR #11498, opened 2016-03-06
Added setLastModified method to Phalcon\Http\Response class

jaskaransingh156:

https://github.com/phalcon/cphalcon/issues/11491


sergeyklay:
Thanks

PR #11499, opened 2016-03-06
Removed _processed property from Annotations Router

SidRoberts:
The `_processed` property stops the Router from being used multiple times. I can't understand the restriction as it works fine without it and the other Routers don't have this.


Green-Cat:
:+1:

sergeyklay:
Thanks

PR #11506, opened 2016-03-08
Removed unnecessary gc from Mongo cache.

SidRoberts:
`flush()` removes everything anyway so we don't need to collect any garbage.

### DIFF
--- a/phalcon/cache/backend/mongo.zep
+++ b/phalcon/cache/backend/mongo.zep
@@ -438,10 +438,6 @@ class Mongo extends Backend implements BackendInterface
 	{
 		this->_getCollection()->remove();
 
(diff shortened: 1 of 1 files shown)


sergeyklay:
```
There was 1 failure:

1) CacheTest::testCacheMongoFlush

Failed asserting that Array (

    'n' => 2

    'connectionId' => 1

    'err' => null

    'ok' => 1.0

) is true.

/home/travis/build/phalcon/cphalcon/unit-tests/CacheTest.php:1371
```

SidRoberts:
I tried to be clever. :stuck_out_tongue:

sergeyklay:
Thanks

PR #11511, opened 2016-03-09
Fix view selection when multiple view directories have been set

Ragnis:
The problem was that when the target view was not found in the very first views directory, the search was aborted and an exception got thrown. This meant that the views were never searched for in the other views directories.

I am not completely sure, how the `view:notFoundView` event should be handled here -- should it fire for every view not found or only when no views are found?

### DIFF
--- a/phalcon/mvc/view.zep
+++ b/phalcon/mvc/view.zep
@@ -726,7 +726,7 @@ class View extends Injectable implements ViewInterface
 			}
 
 			if !silence {
-				throw new Exception("View '" . viewsDirPath . "' was not found in the views directory");
+				throw new Exception("View '" . viewPath . "' was not found in any of the views directory");
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


SidRoberts:
FWIW, I think the event should only be triggered if Phalcon can't find the view in any folder. That's what I would expect to happen.

certainlysylvia:
I missed this - check whether this is the same issue I address in pull request #11561 ? ... I addressed it is the same line that SidRoberts mentioned: the event is only triggered if the view is not found in ANY path.

Ragnis:
Yeah, it seems to be the same issue.

sergeyklay:
Could you please check now 2.1.x branch?

sergeyklay:
@Ragnis

Ragnis:
What exactly do you want me to do?

sergeyklay:
Could you please rebase?

phalcon/mvc/view.zep
```diff
@@ -713,6 +714,8 @@ class View extends Injectable implements ViewInterface
 					}
 					break;
 				}
+
+				let viewEnginePaths[] = viewEnginePath;
```
sergeyklay: Could you please explain why we need this

PR #11522, opened 2016-03-11
Fixed Mvc\Model\Transaction::isManaged()

### LINKED ISSUE #11488: [2.1.x] Transaction isManaged returns wrong result
Looking at sources Phalcon\Mvc\Model\Transaction:
```c
/**
 * Checks whether transaction is managed by a transaction manager
 */
public function isManaged() -> boolean
{
	return !this->_manager;
}
```

Now it returns if transaction is not managed. Shouldn't it be the other way around? 

P.S. Dat issue ID

SidRoberts:
Fixes #11488.


sergeyklay:
Thanks

Issue #11486, opened 2016-03-03
Change mcrypt on openssl


Studentsov:
Due to the obsolescence of mcrypt (absence of updates of nearly 10 years) and its loss on openssl speed, I offer an exception it from dependences of a framework and its changeover by openssl. It is possible to mention as arguments that it want to exclude from PHP kernel long ago (https://wiki.php.net/rfc/removal_of_dead_sapis_and_exts, https://wiki.php.net/rfc/mcrypt-viking-funeral), and also the fact that Phalcon already depends on openssl.

sergeyklay:
:+1:

Studentsov:
It is necessary to break backward compatibility as these extensions aren't interchanged. It would be interesting to listen to judgements of the people who are responsible for development of a framework.

sergeyklay:
I agree. 
@andresgutierrez @niden @xboston @carvajaldiazeduar

xboston:
:+1:

niden:
I think that this is a good idea. Yes we will break backwards compatibility but we will break it for a very good reason and that is security. It is the same argument for using an older version of PHP vs a newer one and thus we dropped support for PHP 5.3.

Nothing beats a good design with solid security enhancements in place, but the tools have to match the effort and this move IMHO is more than welcome. We just have to make sure everyone knows about it so that they can adjust their code accordingly.

sergeyklay:
Implemented in 2.1.x branch

PR #11528, opened 2016-03-12
Mcrypt is replaced with openssl

Studentsov:
Mcrypt is replaced with openssl in `Phalcon\Crypt`. (https://github.com/phalcon/cphalcon/issues/11486)
Removed methods `setMode()`, `getMode()`, `getAvailableModes()` as not applicable to the openssl in `Phalcon\CryptInterface`.
Unfortunately, the reverse compatibility was lost for the reason that cryptography techniques are called differently in mcrypt and openssl libraries. Moreover, the algorithm of rijndael-256 used by default doesn't have changeover in openssl. I hope, it is not too heavy payment for transition to faster, safe and modern cryptography library.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,8 @@
 - Added `Phalcon\Cache\Backend\Memcache::addServers` to enable pool of servers for memcache
 - Added `setLastModified` method to `Phalcon\Http\Response`
 - Added `Phalcon\Validation\Validator\Date`
+- Mcrypt is replaced with openssl in `Phalcon\Crypt`
+- Removed methods setMode(), getMode(), getAvailableModes() in `Phalcon\CryptInterface`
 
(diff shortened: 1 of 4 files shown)


sergeyklay:
@Studentsov See https://github.com/phalcon/cphalcon/blob/2.1.x/tests/unit/CryptTest.php

Studentsov:
There were some difficulties which are generally connected to distinctions of implementations of similar algorithms in mcrypt and openssl, and also imperfection of the current implementation of `Phalcon\Crypt`. I will be glad if someone undertakes carrying out additional tests.

sergeyklay:
Could you please squash it into a single commit to keep the history clean? Let me know if you need instructions.

Studentsov:
If I understand correctly, I need to create a new branch and copy the changed files to it, and then create a new pull-request.

sergeyklay:
if it will be convenient - yes, also you can to squash commits into one in this PR

PR #11530, opened 2016-03-13
Mcrypt is replaced with openssl

Studentsov:
Mcrypt is replaced with openssl in `Phalcon\Crypt`. (https://github.com/phalcon/cphalcon/issues/11486)
Removed methods `setMode()`, `getMode()`, `getAvailableModes()` as not applicable to the openssl in `Phalcon\CryptInterface`.
Unfortunately, the reverse compatibility was lost for the reason that cryptography techniques are called differently in mcrypt and openssl libraries. Moreover, the algorithm of rijndael-256 used by default doesn't have changeover in openssl. I hope, it is not too heavy payment for transition to faster, safe and modern cryptography library.
There were some difficulties which are generally connected to distinctions of implementations of similar algorithms in mcrypt and openssl, and also imperfection of the current implementation of `Phalcon\Crypt`. I will be glad if someone undertakes carrying out additional tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,8 @@
 - Added `Phalcon\Cache\Backend\Memcache::addServers` to enable pool of servers for memcache
 - Added `setLastModified` method to `Phalcon\Http\Response`
 - Added `Phalcon\Validation\Validator\Date`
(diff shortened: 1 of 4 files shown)


andresgutierrez:
Thanks :+1:

Issue #11275, opened 2016-01-06
[NFR] Pdo for sqlrelay


tigran-a:
Hi, 

Any plans of adding support of sqlrelay? Would be helpful!

Thanks!

spuy767:
You should really prefix this with [NFR] so it's easy to distinguish from bugs.

PR #11532, opened 2016-03-13
Added support for sql relay

jaskaransingh156:

https://github.com/phalcon/cphalcon/issues/11275

Sample code to use this is following:

``` php
$config = [
    'host'        => 'localhost',
    'port'        => 9000,
    'username'    => 'mysqltest',
    'password'    => 'mysqltest',
    'backendType' => 'mysql'
];

$connection = new Phalcon\Db\Adapter\Pdo\Sqlrelay($config);
```

Here, `backendType` is the DBMS to which Sqlrelay connects to. Sample sqlrelay.conf is:

``` xml
<?xml version="1.0"?>
<!DOCTYPE instances SYSTEM "sqlrelay.dtd">
<instances>
    <instance id="mysqltest" dbase="mysql" connections="5" cursors="50">
        <users>
            <user user="mysqltest" password="mysqltest"/>
        </users>
        <connections>
            <connection string="user=test;password=test;db=testing;host=localhost"/>
        </connections>
    </instance>
</instances>
```


sergeyklay:
Here is many things that need discussion. But first need to fix the code so the test passed

jaskaransingh156:
@sergeyklay I have fixed the code.

jaskaransingh156:
I have changed the code to remove sqlrelay dialect. Now, dialect used in sqlrelay would any one out of mysql, oracle, postgresql and sqlite. It would be taken as input from user in the `backendType` parameter.

sergeyklay:
I'm closing this issue due to the lack of any reaction. I'll open it again if the need arises. So far no one has been able to understand why this feature is needed in the framework' core.

PR #11533, opened 2016-03-13
Tidied up Mvc classes a bit

SidRoberts:


sergeyklay:
As you can see tests fails

sergeyklay:
Thanks

PR #11535, opened 2016-03-13
Model::find...By...() fields are now escaped properly (fixes #11253).

### LINKED ISSUE #11253: Models apparently have reserved keywords
The following keywords (when used as column names) were found to cause issues with phalcon's models (although I could not find a list in the docs). IMHO they should not cause issues and should be allowed to be used, but at least it should be documented:

`group` - causes an exception during PHQL parsing as it sees it as meaning GROUP.
`transaction` - meddles with transactions used when nesting models.
`for` - Causes error similar to `group`'s.
`source` - Confuses phalcon to think we are declaring the table source.

Edit: I should note, this is using mysql backend.

SidRoberts:


phalcon/mvc/model.zep
```diff
@@ -4024,7 +4024,7 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 		 * Execute the query
 		 */
 		return {modelName}::{type}([
-			"conditions": field . " = ?0",
+			"conditions": "[" . field . "] = ?0",
```
sergeyklay: I like this

sergeyklay:
Thanks

PR #11540, opened 2016-03-14
Fixed incorrect detection of column type with enums (fixes #11505)

### LINKED ISSUE #11505: Incorrect model ENUM validation
There is an issue with ENUM values handling.
This scenario wont work:

Phalcon Version: ```2.0.10```

Database:
```
CREATE TABLE `robots` (
	`id` INT(11) NOT NULL AUTO_INCREMENT,
	`name` VARCHAR(50) NULL DEFAULT NULL,
	`type` ENUM('internet','member','prepaid','terminal') NOT NULL,
	PRIMARY KEY (`id`)
)
```

Model:
```
class Robots extends Phalcon\Mvc\Model
{
    public $id;
    public $name;
    public $type;
}
```

Example:
```
$robot = new Robots;
$robot->name = 'TestMember';
$robot->type = 'member';

if(!$robot->save())
{
     echo $robot->getMessages();
}
```
Result:
```
type is required
```

When i remove "internet" from ENUM types in database, the code is working fine.
I think its because the string "INT" exists in word "INTernet" like in this old issue: https://github.com/phalcon/cphalcon/issues/709

kristoffer-ingemansson:
Moved enum detection to top.

### DIFF
--- a/phalcon/db/adapter/pdo/mysql.zep
+++ b/phalcon/db/adapter/pdo/mysql.zep
@@ -121,6 +121,14 @@ class Mysql extends PdoAdapter implements AdapterInterface
 			loop {
 
 				/**
+				 * Enum are treated as char
+				 */
+				if memstr(columnType, "enum") {
+					let definition["type"] = Column::TYPE_CHAR;
+					break;
+				}
+
+				/**
 				 * Smallint/Bigint/Integers/Int are int
 				 */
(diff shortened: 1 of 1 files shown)


sergeyklay:

kristofferingemansson:
Sorry, first pull request ever. I'll try and find correct branch.

PR #11555, opened 2016-03-16
The beforeException dispatcher event can now forward

### LINKED ISSUE #2851: [BUG] Action forwarding is not working in 'dispatch:beforeException'.
Hi all,

It is related to this: http://docs.phalconphp.com/en/latest/reference/dispatching.html#handling-not-found-exceptions. Whenever an exception is thrown, action forwarding to Error controller does not work and results in a blank page.

Steps to repro:
* Check out Phalcon's "INVO" example.
* Create ErrorController.php:
```php
    class ErrorController extends ControllerBase
    {
        public function indexAction()
        {
                error_log('Error Controller');
        }
    }
```
* Create view script "error/index.volt" with a message "KABOOM!"
* Edit /index.php and use the following code:
```php
	/**
	 * We register the events manager
	 */
	$di->set('dispatcher', function() use ($di) {

		$eventsManager = $di->getShared('eventsManager');
        $eventsManager->attach('dispatch:beforeDispatchLoop', function(Phalcon\Events\Event $event, Phalcon\Mvc\Dispatcher $dispatcher) {
            throw new \Exception('Just because');
        });

        $eventsManager->attach('dispatch:beforeException', function ($event, $dispatcher, $exception) use (&$di) {
            error_log('dispatch:beforeException');
            $dispatcher->forward(
                array(
                    'controller' => 'error',
                    'action'     => 'index',
                    'error' => $exception
                )
            );
            return false;
        });

		$security = new Security($di);

		/**
		 * We listen for events in the dispatcher using the Security plugin
		 */
		$eventsManager->attach('dispatch', $security);

		$dispatcher = new Phalcon\Mvc\Dispatcher();
		$dispatcher->setEventsManager($eventsManager);

		return $dispatcher;
	});
```

Exception "Just because" is thrown and caught in 'dispatch:beforeException'. However, forwarding to /error/index does not work. Execution terminates immediately after "return false". I am using PHPEd debugger to trace it.

Also,

Can Phalcon guys explain the following cryptic sentence from the docs:

```txt
Only exceptions produced by the dispatcher and exceptions produced in the executed action are notified in the ‘beforeException’ events. Exceptions produced in listeners or controller events are redirected to the latest try/catch.
```

What does that really mean?? 

Thanks!

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/4913060-bug-action-forwarding-is-not-working-in-dispatch-beforeexception?utm_campaign=plugin&utm_content=tracker%2F50707&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F50707&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SidRoberts:
If an Exception is thrown, it needs to be caught within the loop and be able to continue if we need it to.

Fixes #2851.

### DIFF
--- a/phalcon/dispatcher.zep
+++ b/phalcon/dispatcher.zep
@@ -362,7 +362,7 @@ abstract class Dispatcher implements DispatcherInterface, InjectionAwareInterfac
 			actionName, params, eventsManager,
 			actionSuffix, handlerClass, status, actionMethod, reflectionMethod, methodParams,
 			className, paramKey, methodParam, modelName, bindModel,
-			wasFresh = false;
+			wasFresh = false, e;
 
 		let dependencyInjector = <DiInterface> this->_dependencyInjector;
 		if typeof dependencyInjector != "object" {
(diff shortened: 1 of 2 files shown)


sergeyklay:
Could you please provide a small test?

sergeyklay:
Thanks

PR #11559, opened 2016-03-17
Update view.zep

certainlysylvia:
Where you have multiple view paths/directories ... searching for anything (partial etc) fails (errors out) if the first path it tests does not have the item.

Perhaps this is expected behaviour but it seems to me that the error should only rise if the item is not in ANY of the multiple paths.

If so, then the fix is simple as seen.


certainlysylvia:
*headdesk* - the break should also become a return (near end) ... will loook at refiling this soon as can :)

PR #11560, opened 2016-03-17
Update view.zep

certainlysylvia:
Allows multiple view directories where, as long as an item is found in any one of them, it is found. Currently, if it fails before it succeeds (in the loop) an error rises. Patch fixes this so, for example, a partial is 'found' if it exists in ANY of the paths.


certainlysylvia:
The check against notFound at end should be to check if it is true ... not false ... having a day :)

PR #11561, opened 2016-03-17
Update view.zep

certainlysylvia:
Finally!

Ok - this fixes an issue with multiple view paths. As it was, if a resource was not found in the first view path, then and error was raised and the other paths were not checked. Now - a resource (partial / whatever) is marked as 'found' if it exists in ANY of the multiple view paths.


certainlysylvia:
Intersects with #11511 :)

andresgutierrez:
Thanks

PR #11566, opened 2016-03-18
Fix namespace resolution

Bellardia:
Fixed an issue that caused the fully qualified class to have the namespace duplicated in sufixed when handle contains a namespace itself.

### DIFF
--- a/phalcon/mvc/router/annotations.zep
+++ b/phalcon/mvc/router/annotations.zep
@@ -147,7 +147,7 @@ class Annotations extends Router
 			 */
 			fetch moduleName, scope[2];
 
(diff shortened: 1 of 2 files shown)


Green-Cat:
:+1: 
Could you a small test to prevent the issue from being reintroduced in the future?

Bellardia:
I'll try and get something up sometime tomorrow

Bellardia:
@Green-Cat I updated the test with the failing condition best I could without increasing complexity of test itself. The existing branch will throw an exception.

Green-Cat:
:+1:

sergeyklay:
Thanks

Issue #11568, opened 2016-03-19
Volt template names and camelized controller actions


makerlabs:
Hi, 
Im having a strange issue In work we use a compiled version of phalcon from homebrew and it works fine I mean when we create a controller action named "addProductAction" the phalcon is calling the "addProduct.volt" template. But Now I switched to linux and compiled phalcon from sources and ran our application phalcon don't see the template "addProduct.volt" it tries to search for "addproduct.volt" why?

sergeyklay:
Possible refs: https://github.com/phalcon/cphalcon/issues/284

makerlabs:
This is actually a bug in the Annotation Router I created a fix for that

PR #11569, opened 2016-03-20
Annotation Router - Removed converting the action name to lowercase

makerlabs:
I found this bug by analyzing the issue https://github.com/phalcon/cphalcon/issues/11568. I still don't know why the the action name was converted to lowercase.


Green-Cat:
This change seems to be breaking tests.

makerlabs:
Sorry about that I forgot to update the unit test now it's fixed.

sergeyklay:
Could you please rebase? Also update CHANGELOG.md

makerlabs:
OK but what should I put in the changelog? The 2.0.10 is already out you could also better describe the whole pull request workflow in the documentation because there is no reference to that.

Jurigag:
```Removed converting the action name to lowercase```

andresgutierrez:
This PR contains many commits that don't belong to the PR, could you please submit a new PR?

PR #11571, opened 2016-03-20
Clear queue

shopiz:
日志事务提交，清空队列


phalcon/logger/adapter.zep
```diff
@@ -122,6 +122,7 @@ abstract class Adapter
 					message->getContext()
 				);
 			}
+			this->_queue = [];
```
Green-Cat: This should be ```let this->_queue = [];``` in zephir.

PR #11572, opened 2016-03-20
clear queue

shopiz:
日志事务提交，清空队列

### DIFF
--- a/phalcon/logger/adapter.zep
+++ b/phalcon/logger/adapter.zep
@@ -122,6 +122,7 @@ abstract class Adapter
 					message->getContext()
 				);
 			}
+			this->_queue = [];
 		}
 
 		return this;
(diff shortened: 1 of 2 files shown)


sergeyklay:

shopiz:
I need to submit to which branch？2.1.x？

sergeyklay:
If this is bug related to the 2.0.x - send to the 2.0.x. Anyway  please describe a bit more what exactly are you trying to fix. And I'm not sure that you need to change Phalcon\Mvc\Model

shopiz:
When the table design for not null, and set the default value (0 or the empty string), at the time of update, can not succeed.

shopiz:
When the table field is designed as null not, and set the default value (0 or empty string), the original 0 or empty string will be modified to null value at update, leading to the update failed.

shopiz:
If I submit it to 2.0.x, do I need to submit it to 2.1.x?

sergeyklay:
Could you please separate PRs?

PR #11575, opened 2016-03-21
Removed converting the action name to lowercase in `Phalcon\Mvc\Route…

makerlabs:
Fixed issue #11568


andresgutierrez:
This will break several applications

sergeyklay:
@makerlabs 

This PR will result in serious breaks in many applications.
All developers will have to rename all view folders. 

We will return to this PR when the 2.2.x branch is ready.

Jurigag:
But there is inconsistency now. The framework should be all about consistency, you can define actions with annotatons and it will be camel sensitive, but when define it in annotations it won't ?

makerlabs:
OK no problem for now I will just use my branch. We could also add additional parameter that will change this behavior for routing but I don't think that this will be a good approach as Jurigag stated this is an inconsistency.

fenixphp:
If it plays such a role in the views, why transform "lowercase" does not move directly the call view and keep the original name of the method?

PR #11577, opened 2016-03-21
Form validation : fetch elements after "bind" and "beforeValidation" have been called

challet:
They are called from form->isValid() method, and can be custom methods. Any use of form->remove or form->add inside one of this function can change the set of elements to be validated.
Moving the elements fetch after they are called remove the case where the elements being valdated are not the actual stored ones.

### DIFF
--- a/phalcon/forms/form.zep
+++ b/phalcon/forms/form.zep
@@ -270,11 +270,6 @@ class Form extends Injectable implements \Countable, \Iterator
 			validators, name, preparedValidators, filters,
 			validator, validation, elementMessages;
 
-		let elements = this->_elements;
-		if typeof elements != "array" {
-			return true;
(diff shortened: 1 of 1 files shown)


phalcon/forms/form.zep
```diff
@@ -299,7 +294,12 @@ class Form extends Injectable implements \Countable, \Iterator
 				return false;
 			}
 		}
-
+		
+		let elements = this->_elements;
+		if typeof elements != "array" {
```
SidRoberts: I recommend removing `if` statement so that the afterValidation event is run.

challet:
Looks like it has been done here 542d17e1bbb58d40e3641dd9f5e33a1c6040ab1d

Issue #11550, opened 2016-03-16
PHP 7 Tests related issues


Jurigag:
WORK IN PROGRESS - not all yet

**UP TO DATE FOR NOW(26.07.2016 8:48 GMT+1)**

I will update this list if something is fixed or new problem occurs:

- ```Phalcon\Assets\Resource``` needs to be changed cuz Resource is PHP 7 reserved word - http://php.net/manual/en/reserved.other-reserved-words.php - **NOT AN ISSUE**, but still it would be nice to change cuz it some day can be reserved word in php
- [x] - First usage of ```Phalcon\Test\Proxy\Assets\Collection``` cause error but then runs fine, if we return true in ```Phalcon\Test\Unit\Assets\CollectionTest::testAssetsResourceCollection``` then it runs fine, there is problem just when we construct this proxy class, if we change it to internal phalcon class then there is no problem - _fixed by https://github.com/phalcon/cphalcon/commit/99e197735199b46854b0ff6142a5ba043a2e88da_
- [x] - ```Phalcon\Config\Ini::toArray()``` returns ```Phalcon\Config``` object as a key - i think it's possible that it's related to https://github.com/phalcon/zephir/issues/1203 beacause its using ```method_exists``` too - _Fixed by https://github.com/phalcon/zephir/commit/68a1a2291b25d71112a557186d3dd507f309c420_
- [x] - First usage of ```Phalcon\Test\Proxy\Config``` cause error but then runs fine, if we change ```Config``` to ```PhConfig``` in ```Phalcon\Test\Unit\ConfigTest::testConfigToArray``` then it runs fine and again error now in ```Phalcon\Test\Unit\ConfigTest::testConfigCount```
- ```Phalcon\Config::merge()``` is just changing value of keys instead of merging, beacause type of key is not integer, it's related to https://github.com/phalcon/zephir/issues/1214 - _not an issue, https://github.com/phalcon/zephir/issues/1214#issuecomment-217577420_
[...]

PR #11578, opened 2016-03-21
Fix segfault on PHP 7

Bellardia:
Back porting some code from 2.0.0 branch, and including an update for PHP 7.0.

Current implementation results in a SEGFAULT when run under PHP 7.0, stemming from the below code in `mvc/model/query/builder.zep`

```
let query = <QueryInterface> dependencyInjector->get("Phalcon\\Mvc\\Model\\Query", [phql, dependencyInjector]);
```

I could potentially fix through `create_instance_params`, but I can't find where that function is defined - nor `is_php_version` for that matter.

No unit tests since PHP 7 is not running on travis.

### DIFF
--- a/phalcon/di.zep
+++ b/phalcon/di.zep
@@ -205,7 +205,7 @@ class Di implements DiInterface
 	 */
 	public function get(string! name, parameters = null) -> var
 	{
-		var service, eventsManager, instance = null;
+		var service, eventsManager, reflection, instance = null;
 
 		let eventsManager = <ManagerInterface> this->_eventsManager;
 
@@ -232,7 +232,12 @@ class Di implements DiInterface
 				}
 
 				if typeof parameters == "array" && count(parameters) {
-					let instance = create_instance_params(name, parameters);
+					if is_php_version("5.6") || is_php_version("7.0") {
+						let reflection = new \ReflectionClass(name),
+							instance = reflection->newInstanceArgs(parameters);
+					} else {
+						let instance = create_instance_params(name, parameters);
+					}
 				} else {
 					let instance = create_instance(name);
 				}


Jurigag:
create_instance_params is zephir function, it's defined in c code i guess, i created issue there : https://github.com/phalcon/zephir/issues/1218

Bellardia:
@Jurigag Thanks a lot, I'll poke around

Jurigag:
Also it's no point in merging it. You would need to change ALL create_instance_params in all places to make it sure it's working on php7, and still you would fix only one possible place for seg faults and there are much more(check zephir repo). Just it's need to be fixed in zephir, not changed in phalcon imho. You can use yourself version with reflection but there is no point in doing PR.

Bellardia:
@Jurigag I'm still evaluating the issue fully, but it seems relatively limited in scope. `create_instance_params` does work the majority of the time.

Jurigag:
Because it's a bug in zephir and only php7, not phalcon. It needs to be fixed in zephir

Bellardia:
Yes, agreed.

Jurigag:
Also there is list oh php 7 related issues which are happening currently in tests:

https://github.com/phalcon/cphalcon/issues/11550

Izopi4a:
i dont think `create_instance_params` is "managed" by zephir.

https://github.com/phalcon/zephir/blob/f68701626f413379565b6e1ebc33df23c8e41bcb/Library/FunctionCall.php#L248

i think this pull request is legit, after dinner i will try it with my app.

Jurigag:
``` https://github.com/phalcon/zephir/blob/f68701626f413379565b6e1ebc33df23c8e41bcb/Library/FunctionCall.php#L260
```

?

Green-Cat:
It is a built-in function in zephir, here's the optimizer: https://github.com/phalcon/zephir/blob/7cfd109f599fc0e554318b66afc3fc4d0433c297/Library/Optimizers/FunctionCall/CreateInstanceParamsOptimizer.php

Izopi4a:
ha, fuck me indeed

Jurigag:
@Bellardia 

This PR can be closed. This issue is already fixed on newest zephir. Actually when using php7-parsers branch and newest zephir generated in new way the PHQL is already working fine.

PR #11588, opened 2016-03-24
Update CHANGELOG.md

Studentsov:
Obsolete information on changes in `Phalcon\Crypt` is corrected.


Studentsov:
It would also be quite good to update information in the official blog.

PR #11277, opened 2016-01-06
[NFR] RoleAware and ResourceAware Interface

Jurigag:

PR #11590, opened 2016-03-24
Firewall component, afterBinding in dispatcher

Jurigag:
@sergeyklay @daison12006013 @valVk @SidRoberts

TODO, Requires discussion:
- [x] - After binding tests
- [x] - Discussion about `Phalcon\Firewall\Adapter::setRoleCallback()` - maybe diffrent name, maybe we could pass $di as parameter which will be get from dispatcher instead of writing use ($di), don't know if there is any performance difference ?
- [x] - `Phalcon\Firewall\Adapter\Acl` - what if we have controllers with same names(multi-module application i guess), maybe some convenient, standardized way of naming in acl service when we have multi-module application ? Like Module:Controller as resource name perhaps ?
- [x] - Check tests and classes, especially tests of `Phalcon\Firewall\Adapter\Annotations` - not sure if the current logic is fine(especially when we provide user roles as array and we have multiple roles in `@Deny` or `@Allow` and also if logic with controller/method checking is fine, for me it's obvious when i put `@Allow` before controller it means for me that i have access to whole controller
- [x] - Maybe whole Firewall namespace should be put into Phalcon\Mvc\Dispatcher, beacause it depends from this dispatcher ?
- ~~Methods in firewall to access current controller/action/user/role - for handling exception.~~

In future(or now) - implement firewall adapter on adding routes - not sure if we should add event manager to router, or just internally in router there should be calls into our firewall.

### DIFF
--- a/phalcon/dispatcher.zep
+++ b/phalcon/dispatcher.zep
@@ -76,6 +76,8 @@ abstract class Dispatcher implements DispatcherInterface, InjectionAwareInterfac
 
 	protected _modelBinding = false;
 
+	protected _boundModel = null;
+
 	const EXCEPTION_NO_DI = 0;
 
 	const EXCEPTION_CYCLIC_ROUTING = 1;
@@ -328,6 +330,14 @@ abstract class Dispatcher implements DispatcherInterface, InjectionAwareInterfac
 	}
 
 	/**
+	 * Gets a binded model
+	 */
+	public function getBoundModel() -> object
+	{
+		return this->_boundModel;
+	}
+
+	/**
 	 * Dispatches a handle action taking into account the routing parameters
 	 *
 	 * @return object
@@ -564,6 +574,7 @@ abstract class Dispatcher implements DispatcherInterface, InjectionAwareInterfac
 									let modelName = call_user_func([handlerClass, "getModelName"]);
 									let bindModel = call_user_func_array([modelName, "findFirst"], [params[paramKey]]);
 									let params[paramKey] = bindModel;
+									let this->_boundModel = bindModel;
 									break;
 								}
 							}
@@ -572,10 +583,16 @@ abstract class Dispatcher implements DispatcherInterface, InjectionAwareInterfac
 							if is_subclass_of(className, "Phalcon\\Mvc\\Model") {
 								let bindModel = call_user_func_array([className, "findFirst"], [params[paramKey]]);
 								let params[paramKey] = bindModel;
+								let this->_boundModel = bindModel;
 								break;
 							}
 						}
 					}
+				}
+			}
+			if typeof eventsManager == "object" {
+				if eventsManager->fire("dispatch:afterBinding", this) === false {
+					continue;
 				}
 			}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@
 - Fixed `Filter::add` method handler [#11581](https://github.com/phalcon/cphalcon/issues/11581)
 - Removed `Phalcon\Session` [#11340](https://github.com/phalcon/cphalcon/issues/11340)
 - Phalcon\Tag::getTitle() shows a title depending on prependTitle and appendTitle
+- Added `Phalcon\Mvc\Dispatcher\Firewall\Adapter\Acl` and `Phalcon\Mvc\Dispatcher\Firewall\Adapter\Annotations`, built-in classes for securing controllers/actions
 
 # [2.0.11](https://github.com/phalcon/cphalcon/releases/tag/phalcon-v2.0.11) (????-??-??)
 - Fix Model magic set functionality to maintain variable visibility and utilize setter methods.[#11286](https://github.com/phalcon/cphalcon/issues/11286)
--- a/phalcon/mvc/dispatcher/firewall/adapter.zep
+++ b/phalcon/mvc/dispatcher/firewall/adapter.zep
@@ -1,0 +1,150 @@
+
+/*
+ +------------------------------------------------------------------------+
+ | Phalcon Framework                                                      |
+ +------------------------------------------------------------------------+
+ | Copyright (c) 2011-2016 Phalcon Team (http://www.phalconphp.com)       |
+ +------------------------------------------------------------------------+
+ | This source file is subject to the New BSD License that is bundled     |
+ | with this package in the file docs/LICENSE.txt.                        |
+ |                                                                        |
+ | If you did not receive a copy of the license and are unable to         |
+ | obtain it through the world-wide-web, please send an email             |
+ | to license@phalconphp.com so we can send you a copy immediately.       |
+ +------------------------------------------------------------------------+
+ | Authors: Andres Gutierrez <andres@phalconphp.com>                      |
+ |          Eduar Carvajal <eduar@phalconphp.com>                         |
+ |          Wojciech Ślawski <jurigag@gmail.com>                          |
+ +------------------------------------------------------------------------+
+ */
+
+
+namespace Phalcon\Mvc\Dispatcher\Firewall;
+
+use Phalcon\Events\EventsAwareInterface;
+use Phalcon\Events\ManagerInterface;
+use Phalcon\Mvc\Dispatcher\Firewall\AdapterInterface;
+use Phalcon\Acl;
+use Phalcon\Mvc\Dispatcher;
+
+/**
+ * Phalcon\Mvc\Dispatcher\Firewall\Adapter
+ *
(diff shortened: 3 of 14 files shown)


phalcon/dispatcher.zep
```diff
@@ -328,6 +330,14 @@ abstract class Dispatcher implements DispatcherInterface, InjectionAwareInterfac
 	}
 
 	/**
+	 * Gets a binded model
+	 */
+	public function getBindedModel() -> object
```
nsossonko: Technically this should be bound model...

Jurigag: Oh thanks, i didn't have idea how to name it, my english sometimes are bad.

Jurigag:
Also im guessing that firewall should have methods to get controller name, action, roles, module name, to make them be available in handleException i guess ?

Jurigag:
@sergeyklay @daison12006013 @valVk @SidRoberts

Could you guys look into it and comment about things i mentioned when you have time and i will update it ? All of you were writing in ACL on router issue. I think it needs discussion before adding it to phalcon in any time. I just need some feedback and will do the rest.

phalcon/firewall/adapter/acl.zep
```diff
@@ -0,0 +1,161 @@
...
+	 * Bound model
+	 * @var object
+	 */
+	protected _boundModel = null;
+
+	/**
+	 * Phalcon\Firewall\Adapter\Acl constructor
+	 *
+	 * @param string aclServiceName
+	 * @param string bindedModelParameterName
```
Green-Cat: Should be ```boundModelParameterName```

phalcon/firewall/adapter/acl.zep
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	protected function handleDispatcher(<DispatcherInterface> dispatcher)
+	{
+		var role, identity, actionName, eventsManager, controllerAccess, actionAccess, acl, aclServiceName, aclRole, aclAccess, defaultAccess, parameters, boundModelParameterName, value, dependencyInjector, controllerName;
+		string controllerPrefix;
+		let dependencyInjector = dispatcher->getDI();
+		if typeof dependencyInjector != "object" {
+        	throw new Exception("A dependency injector container is required to obtain ACL service");
+        }
```
Green-Cat: Wrong indentation

Green-Cat:
Major :+1:  A very useful feature to have.

sergeyklay:
It looks more something for the incubator than the core. Keeping this open for discussion

Jurigag:
Well, not everyone knows about incubator and using it, if it would be in core and docs then someone will know that there is such behavior. I guess there are already similar thing in incubator, so no reason for adding this to it.

Green-Cat:
This might belong under the ```Phalcon\Security\Firewall``` namespace, since it's a security feature and the namespace is mostly underused. But then again I also think the ACL should be under security too.

@sergeyklay I would say it belongs in the core, because:
1) It's a very useful feature, which can save a considerable amount of work in doing a rather common task of checking the user role before executing a controller action. Most web apps have some form of usergroups or roles, at least an admin with added permissions. This feature makes securing an admin panel, for example, a single annotation line, which is a great developer experience.
2) we already have many features in the core that are much more situational than securing access to routes, for example we have the image adapters which I doubt are used by more than 1% of the users of the framework.
3) Most other frameworks have something similar. For example http://symfony.com/doc/current/components/security/firewall.html . Therefore many new users coming to the framework may expect to find a similar feature.

daison12006013:
This is really a helpful, upon checking the test sample, it's kind a more easy way to handle rather than building a set of ACL, this firewall feature is awesome I guess. :+1: @Jurigag

sergeyklay:
I don't say NO.

I just want to discuss this issue in more detail. I don't like the idea of having this feature in the core. Frankly speaking, I have spoken to @andresgutierrez, and he also thinks that this feature must remain in the Incubator. Your suggestion is too optional. Let's just discuss the use case of your feature. Let's take opinion by @niden, @xboston or @carvajaldiazeduar.

Maybe, I don't understand something. Show me a real example where it can be used. Explain why the framework will be poor without this feature at the core level.

rianorie:
I'll chime in as well. First, I think @Green-Cat said most of it in https://github.com/phalcon/cphalcon/pull/11590#issuecomment-204161951, already.

I don't think the framework will be worse-off without it, but I do feel it's an extension to the existing ACL and therefor _ok_ to have in the core. Providing more native features in the framework is not a bad thing, I also always understood the incubator as a place where functionality lives for a while until it gets ported to the framework definitely. Moving it there would by that reasoning simply postpone the merging.

I think the only real arguments that can be brought against it are:
 - *Phalcon doesn't aim to be a full-stack framework so we shouldn't implement "non-core" features.* Which to my knowledge is false, Phalcon does aim to be that, in the long run, attested by the copious amount of libraries already available like Image, Security, Crypt, Queue, etc.

 - *The functionality already exists.* It doesn't exactly though, the ACL we have is different.

Jurigag:
@rianorie 
Actually it's not really extension to acl, there is only option to be able to use it with acl, (but you can as well use it without acl, only as annotations) so we can pretty much change names(or even dont if we name resources as controller prefixes) and add firewall and we ready to go.

I already done extension to acl in this commit: https://github.com/phalcon/cphalcon/pull/11277/files

rianorie:
I'll clarify, I meant it can serve as an extension, as in, provide extra functionality together with the existing ACL. I didn't mean they were actually related as such.

Jurigag:
Yea but it doesn't changing anything in ACL, acl remains the same as it is. Only one problem maybe is that i thought that controller prefix as resource name is fine, but maybe someone else have already other naming convention and he would need to change it to work with built-in firewall.

valVk:
@Jurigag, sorry seems like I late to the party. I'll take a look into this.

I agree that this feature have to live in incubator and later after some time if it'll need than it have to be implemented to the core.

In the same time I belive that this functionality have to be in the core, because this is the very usefull thing in full stack frameworks.

Jurigag:
Well if you need use cases then in like most of the CRM/CMS systems there are roles, and mostly some roles have access to some action, some don't.

sergeyklay:
Just so you now guys

```
$ ls -al ext/modules/
total 300828
drwxr-xr-x 2 root root      4096 Apr 13 17:59 .
drwxrwxr-x 9 klay klay      4096 Apr 13 17:59 ..
-rwxr-xr-x 1 root root 308032250 Apr 13 17:59 phalcon.so
```

308 MB

Jurigag:
@sergeyklay it's only on php 5.6 like that, on php 7 for example is much smaller. Still most of hostings that have phalcon support have higher size. And VPS have moslty 20++ gbs.

rianorie:
That's a 293M module that PHP needs to load **every single time**.. It's not just about the diskspace.

Jurigag:
Wtf ? It's loaded only once on php startup bro.

ogarbe:
How can it be 300Mb whereas in Phalcon 2.0.9 is 4.6Mb ?

sergeyklay:
@ogarbe because some people forget that Phalcon is framework but not all-in-one

ogarbe:
my question is what have change which add 295Mb in the compiled code ??? (the 300Mb code is perhaps the debug one ?)

Jurigag:
Yes, 300 mb is debug one. Phalcon 2.1.x is 4.6 mb.

Jurigag:
Okay. Added the rest of tests and things i wrote here lately and rebased.

If anyone have any idea or something to add that it should work in diffrent way maybe(annotations firewall or something else) then tell me.

Jurigag:
As i see there is no chance for this in 2.1.0 ? How about 2.2.x ?

sergeyklay:
most likely 2.2.x

Jurigag:
Fine with me. Well until 2.1.0 is released then leave it as open so more people can see this and write some idea or comment if have any.

Jurigag:
3.1.x ?

Jurigag:
@Green-Cat well im planning to add ability to use it with `Phalcon\Mvc\Micro` as well. So i would need to add it in other namepsace. So `Phalcon\Firewall` or `Phalcon\Security\Firwall` ? @sergeyklay

sergeyklay:
I prefer `Phalcon\Firewall`

virgofx:
I think this definitely belongs in incubator as it's controversial and adds performance overhead as well. Needs a lot more discussion prior to being included.

Jurigag:
So in php it will be even slower - and this won't be used out of box. You would need to create class and attach it to dispatcher.

PR #11608, opened 2016-03-30
Modify cache redis backend

pantaovay:
# Changes
1. Modify cache redis backend, delete _PHCR
2. Cache redis backend supports "client" option, now you can pass a redis client
3. Redis's `set` method supports `ttl`,the call of `settimeout` is removed
4. Redis has `exists`method to check if a key exists, the previous implementation uses `get` which is wrong
# Why
1. The redis cache backend use a special key _PHCR to store all keys used by cache. There is a issue: #10905 .According to @Green-Cat ,it is disabled in 2.0.x branch which is not true. If use _PHCR, every time we set a cache, we have to run sAdd and set commands. Store all keys in `_PHCR` will cause it expands quickly and this will influence performance.
2. The previous implementation mixes `statsKey` and `prefix`, it uses `_PHCR` as super prefix which is not needed.
3. The cache backend doesn't support redis option which means that we can not pass a redis connection. So if we use modelsCache and modelsMetadata, there will be two redis connections. So I modify this part and let redis option be supported, and through the lifetime of a request we can use one redis connection.

### DIFF
--- a/phalcon/cache/backend/redis.zep
+++ b/phalcon/cache/backend/redis.zep
@@ -3,7 +3,7 @@
  +------------------------------------------------------------------------+
  | Phalcon Framework                                                      |
  +------------------------------------------------------------------------+
- | Copyright (c) 2011-2015 Phalcon Team (http://www.phalconphp.com)       |
+ | Copyright (c) 2011-2016 Phalcon Team (http://www.phalconphp.com)       |
  +------------------------------------------------------------------------+
  | This source file is subject to the New BSD License that is bundled     |
  | with this package in the file docs/LICENSE.txt.                        |
@@ -29,7 +29,7 @@ use Phalcon\Cache\FrontendInterface;
  *
  * Allows to cache output fragments, PHP data or raw data to a redis backend
  *
- * This adapter uses the special redis key "_PHCR" to store all the keys internally used by the adapter
+ * This adapter uses the special redis key "_PHCR" by default to store all the keys internally used by the adapter
  *
  *<code>
  *
@@ -44,6 +44,13 @@ use Phalcon\Cache\FrontendInterface;
  *		'port' => 6379,
  *		'auth' => 'foobared',
  *  	'persistent' => false
+ * ));
+ * 
+ * //You can also pass a redis client
+ * $redis = new \Redis();
+ * $redis->connect('localhost', 6379);
+ * $cache = new Phalcon\Cache\Backend\Redis($frontCache, array(
+ *		'client' => $redis
  * ));
  *
  * //Cache arbitrary data
@@ -185,7 +192,9 @@ class Redis extends Backend implements BackendInterface
 	 */
 	public function save(keyName = null, content = null, lifetime = null, boolean stopBuffer = true)
 	{
-		var lastKey, frontend, cachedContent, preparedContent, ttl, success, specialKey, isBuffering;
+		var redis, lastKey, frontend, cachedContent, preparedContent, ttl, success, specialKey, isBuffering;
+
+		let redis = this->_getRedis();
 
 		if !keyName {
 			let lastKey = this->_lastKey;
@@ -221,9 +230,9 @@ class Redis extends Backend implements BackendInterface
 		}
 
 		if is_numeric(cachedContent) {
-			let success = this->_getRedis()->set(lastKey, cachedContent, ttl);
+			let success = redis->set(lastKey, cachedContent, ttl);
 		} else {
-			let success = this->_getRedis()->set(lastKey, preparedContent, ttl);
+			let success = redis->set(lastKey, preparedContent, ttl);
 		}
 		if !success {
 			throw new Exception("Failed storing the data in redis");
@@ -233,7 +242,7 @@ class Redis extends Backend implements BackendInterface
 			throw new Exception("Unexpected inconsistency in options");
 		}
 		if specialKey != "" {
-			this->_getRedis()->sAdd(specialKey, lastKey);
+			redis->sAdd(specialKey, lastKey);
 		}
 
 		let isBuffering = frontend->isBuffering();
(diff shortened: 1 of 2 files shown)


pantaovay:
See https://github.com/phalcon/cphalcon/pull/11452

pantaovay:
@sergeyklay 😊

sergeyklay:
I'll look into this as soon as I can

phalcon/cache/backend/redis.zep
```diff
@@ -88,13 +99,21 @@ class Redis extends Backend implements BackendInterface
 		}
 
 		if !isset options["statsKey"] {
-			// Disable tracking of cached keys per default
-			let options["statsKey"] = "";
+			let options["statsKey"] = "_PHCR";
```
sergeyklay: > Modify cache redis backend, delete _PHCR

As I remember it was done in  https://github.com/pantaovay/cphalcon/commit/6b8842267d875e5d2a1a58b701e758b909e53dc9

Why are we again added _PHCR ?

phalcon/cache/backend/redis.zep
```diff
@@ -88,13 +99,21 @@ class Redis extends Backend implements BackendInterface
...
 		}
 
 		parent::__construct(frontend, options);
 	}
 
+	public function _getRedis()
```
sergeyklay: the ` _ ` prefix is inconsistent with common style of public Phalcon methods

sergeyklay: It seems to me `getInternalHandler` would be better

phalcon/cache/backend/redis.zep
```diff
@@ -182,39 +192,26 @@ class Redis extends Backend implements BackendInterface
...
 	public function save(keyName = null, content = null, lifetime = null, boolean stopBuffer = true)
 	{
-		var prefixedKey, lastKey, prefix, frontend, redis, cachedContent, preparedContent,
-			tmp, tt1, success, options, specialKey, isBuffering;
+		var redis, lastKey, frontend, cachedContent, preparedContent, ttl, success, specialKey, isBuffering;
+
+		let redis = this->_getRedis();
 
-		if keyName === null {
+		if !keyName {
```
sergeyklay: could you please explain this a bit more

phalcon/cache/backend/redis.zep
```diff
@@ -312,29 +290,19 @@ class Redis extends Backend implements BackendInterface
...
 		if specialKey == "" {
-			throw new Exception("Cached keys need to be enabled to use this function (options['statsKey'] == '_PHCM')!");
+			throw new Exception("Cached keys need to be enabled to use this function (options['statsKey'] == '_PHCR')!");
 		}
 
 		/**
 		* Get the key from redis
 		*/
-		let keys = redis->sMembers(specialKey);
+		let keys = this->_getRedis()->sMembers(specialKey);
```
sergeyklay: anyway 
```zephir
this->_getRedis()->sMembers(specialKey);
```

is expensive than
```zephir
var redis;
let redis = this->_getRedis();

redis->sMembers(specialKey);
```

phalcon/cache/backend/redis.zep
pantaovay: Because it is consistent with the document. Current implementation is not consistent with other cache backends. I will modify the document instead.

phalcon/cache/backend/redis.zep
pantaovay: Would `private` be better?

phalcon/cache/backend/redis.zep
pantaovay: All the use of `keyName` in other methods is `if !keyName`, I think this is inconsistent. In `libmemcached` and `memcache`, `if keyName === null` is used in the `save` method too. Maybe all of them should be `if !keyName`?

phalcon/cache/backend/redis.zep
pantaovay: I don't know much about this. I will modify this part😊

phalcon/cache/backend/redis.zep
pantaovay: If user passed empty string or `0`... An empty string is not allowed but `0` should be allowed.

I think the check should be

```zephir
		if keyName === null || keyName === "" {
			let lastKey = this->_lastKey;
		} else {
			let lastKey = this->_prefix . keyName;
			let this->_lastKey = lastKey;
		}
```

pantaovay: And I find that `statsKey` can not be set as `"0"`

JCMais:
Is that PR still being worked on?

tigran-a:
Hi, just to let you know, this _PHCR is a pain for us, it eats all the memory from our Redis: the keys are stored with TTL, so after they expire, redis removes it, but _PHCR keeps growing regardless if the keys exist or not (there is no explicit key removal from php, we rely on TTL).  Why not to modify all cache backends that they'd interpret null value as not to use the stats key?

tigran-a:
Sorry for using a language that is not related to the project, but this what we have to run from time to time to clean up that key without affecting the production, maybe it might be useful for someone else: 
```
import redis
import socket
KEY = '_PHCR'
r = redis.Redis(host=socket.gethostname())
c = r.scard(KEY)
print("%s has %d elements"%(KEY, c)) 
print("Starting removing the key values. This can take some time") 
t = r.sscan(KEY)
if t[1]:
    r.srem(KEY, *t[1])
while t[0]:
    t = r.sscan(KEY, t[0])
    if t[1]:
        r.srem(KEY, *t[1])
c = r.scard(KEY)
print("%s has %d elements"%(KEY, c)) 
```

wcoppens:
Just to let you know, the _PHCR prefix really is annoying. We for instance use the redis cache across multiple applications. Phalcon in our case created the redis cache data since it acts as a receiver of measurement data. In Phalcon 1.x with the incubator package we did not have the prefix behavior and so all the cached data in Redis is not prefixed. However since we can't stay legacy we are forced to upgrade resulting in upgrading our whole application stack to support the prefix... It would really help us if there would come a solution in which the key is optional.

aat2703:
Will this be added to 3.3.3? Really Looking forward to it 😁

PR #11612, opened 2016-03-30
Fixed SQL error in Paginator when the query builder has set groupBy() specified by string

ivanfetischev:
### Code to reproduce:

``` php
$di['modelsManager']->createBuilder()
    ->columns('cedula, nombres')
    ->from('Personnes')
    ->groupBy('email');
```
### Get error

``` sql
Exception SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error 
in your SQL syntax; check the manual that corresponds to your MySQL server version 
for the right syntax to use near ') AS `rowcount` FROM
```


sergeyklay:

PR #11618, opened 2016-04-01
Cleanup beanstalk doc

sergeyklay:
Removed extra PHPDoc. I'm not sure that Phalcon is right place for copypasta of beanstalkd protocol


sergeyklay:
@Chameleon-m

Chameleon-m:
I personally do not care, I do know where it is...
I would simplify people
--------------------------------------------------------------
Sorry, I'm using google translate)

PR #11624, opened 2016-04-03
Forms\Form::hasMessagesFor() now checks if the Group is empty

SidRoberts:


sergeyklay:
@SidRoberts Could you please create the test in new Phalcon test format? See https://github.com/phalcon/cphalcon/pull/11038#issuecomment-202520081

sergeyklay:
Thanks

PR #11627, opened 2016-04-04
Added ability to call afterFetch method in Simple::toArray(), always call afterFetch even when select columns

Jurigag:
Added ability to call afterFetch method in:
- `Phalcon\Mvc\Model\Resultset\Simple::toArray()`
- `Phalcon\Mvc\Model::find` when searching with columns
- `Phalcon\Mvc\Model::findFirst` when searching with columns
- `Phalcon\Mvc\Model\Query\Builder` when searching with columns(but without joins)

It's using reflectionMethod and stdClass, it's not actually creating object of our model(and using then to Array() method from it), well acutally we have to create it once(to retrieve closure) - actual way is pretty fast(like 5 times than `cloneResultMap`).

**What is important it's not working for joins. Maybe i will implement this some day if there will be need for it.**

If casting object type to array in zephir would work, we could possibly save some time. Right now it's using ArrayObject class.

**Also what is important if unset could work on dynamic properties we could save some time in current() method.**

### DIFF
--- a/phalcon/mvc/model/resultset/simple.zep
+++ b/phalcon/mvc/model/resultset/simple.zep
@@ -23,6 +23,10 @@ use Phalcon\Mvc\Model;
 use Phalcon\Mvc\Model\Resultset;
 use Phalcon\Mvc\Model\Exception;
 use Phalcon\Cache\BackendInterface;
+use ReflectionMethod;
+use ReflectionException;
+use Closure;
+use ArrayObject;
 
 /**
  * Phalcon\Mvc\Model\Resultset\Simple
@@ -36,9 +40,21 @@ class Simple extends Resultset
 
 	protected _model;
 
+	protected _realModel;
+
 	protected _columnMap;
 
 	protected _keepSnapshots = false;
+
+	protected _fireAfterFetch = false;
+
+	protected _closure = null;
+
+	protected _keysDifference = null;
+
+	protected _checkedRemoving = false;
+
+	protected _needsRemoving = false;
 
 	/**
 	 * Phalcon\Mvc\Model\Resultset\Simple constructor
@@ -48,11 +64,14 @@ class Simple extends Resultset
 	 * @param \Phalcon\Db\Result\Pdo|null result
 	 * @param \Phalcon\Cache\BackendInterface cache
 	 * @param boolean keepSnapshots
+	 * @param \Phalcon\Mvc\ModelInterface realModel
 	 */
-	public function __construct(var columnMap, var model, result, <BackendInterface> cache = null, keepSnapshots = null)
+	public function __construct(var columnMap, var model, result, <BackendInterface> cache = null, keepSnapshots = null, var realModel = null, bool fireAfterFetch = false)
 	{
 		let this->_model = model,
-			this->_columnMap = columnMap;
+			this->_columnMap = columnMap,
+			this->_realModel = realModel,
+			this->_fireAfterFetch = fireAfterFetch;
 
 		/**
 		 * Set if the returned resultset must keep the record snapshots
@@ -67,7 +86,8 @@ class Simple extends Resultset
 	 */
 	public final function current() -> <ModelInterface> | boolean
 	{
-		var row, hydrateMode, columnMap, activeRow, modelName;
+		var row, hydrateMode, columnMap, activeRow, modelName, closure,
+		countOriginal, countParsed, objectClosure, rowObject, key, value, arrayObject, arrayCopy, parsedKeys;
 
 		let activeRow = this->_activeRow;
 		if activeRow !== null {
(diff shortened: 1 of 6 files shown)


sergeyklay:
Could you please create a small test which covers this feature

Jurigag:
Well i will do it when have some time.

Jurigag:
@sergeyklay
`renameColumns` was never documented in docs. As well it's not used anywhere(i think) expect https://github.com/phalcon/cphalcon/blob/2.1.x/phalcon/mvc/model/resultset/simple.zep#L237 

I guess no one even know there can be passed parameters to `toArray()`. So could i swap `toArray()` parameters so `fireAfterFetch ` parameters would be first and change this 237 line and `renameColumns` no documented(beacause it's used pretty much internally, i don't really see point why someone would not to rename columns) ?

phalcon/mvc/model/resultset/simple.zep
```diff
@@ -181,48 +188,80 @@ class Simple extends Resultset
...
 
+	private function _executeClosure(<Closure> closure, array record) -> array
+	{
+		var rowObject, objectClosure, arrayObject;
+
+		let rowObject = (object)record;
+		let objectClosure = closure->bindTo(rowObject);
+		{objectClosure}();
+		// in zephir casting object to array doesn't work, so we need to use ArrayObject, we could possibly save some time if we could just (array)rowObject
+		let arrayObject = new ArrayObject(rowObject);
```
sergeyklay: frankly I not rested, but it seems to me

```php
return json_decode(json_encode(rowObject), true);
```

faster than

```php
let arrayObject = new ArrayObject(rowObject);
return arrayObject->getArrayCopy();
```

anyway need to test

Jurigag:
Actually ArrayObject is faster, maybe zephir is optimizing somehow json_decode ?

http://sandbox.onlinephpfunctions.com/code/6e495c0dffb998052ee3043880f507d38d9aae35

phalcon/mvc/model/resultset/simple.zep
Jurigag: Actually ArrayObject is faster in this case(beacause it's not recursive and we don't need this in this case), maybe zephir is optimizing somehow json_decode ?

http://sandbox.onlinephpfunctions.com/code/6e495c0dffb998052ee3043880f507d38d9aae35

Jurigag:
@sergeyklay 
Finally done. There are still failing in tests folder, i will look into it tomorrow(well, today tbh).

Also it's so bad that zephir don't support unset method for dynamic property, beacause i don't know what to do here: https://github.com/phalcon/cphalcon/pull/11627/files#diff-0a0014cdd026fc43956d243d1ecebbcdR163

Jurigag:
Okay tests passing, changed idea a bit. There are some results from testing how about performance. With pretty simple class and afterFetch method like this:

```php
    public function afterFetch()
    {
        $this->datetime=new DateTime($this->datetime);
        $this->name = "asdasdasdas";
    }
```

And code for tests like this:
http://pastebin.com/CGRjPeDc

It returns those results:
```
Array ( [datetime] => DateTime Object ( [date] => 2016-04-09 21:53:42.000000 [timezone_type] => 3 [timezone] => Europe/Warsaw ) ) 0.011757135391235
Array ( [datetime] => 2016-04-09 21:53:42 ) 0.0013010501861572
Array ( [datetime] => DateTime Object ( [date] => 2016-04-09 21:53:42.000000 [timezone_type] => 3 [timezone] => Europe/Warsaw ) [name] => asdasdasdas ) 0.0016560554504395
Array ( [datetime] => DateTime Object ( [date] => 2016-04-09 21:53:42.000000 [timezone_type] => 3 [timezone] => Europe/Warsaw ) ) 0.0017580986022949
Array ( [datetime] => 2016-04-09 21:53:42 ) 0.00095105171203613
Array ( [datetime] => DateTime Object ( [date] => 2016-04-09 21:53:42.000000 [timezone_type] => 3 [timezone] => Europe/Warsaw ) ) 0.00047397613525391
Array ( [datetime] => 2016-04-09 21:53:42 ) 0.00042581558227539
```

tests/unit/Mvc/Model/Resultset/SimpleTest.php
```diff
@@ -0,0 +1,98 @@
...
+            function () {
+                $expectedRobots = [
+                    ['datetime'=>new DateTime('1972-01-01 00:00:00')],
+                    ['datetime'=>new DateTime('1952-01-01 00:00:00')],
+                    ['datetime'=>new DateTime('2029-01-01 00:00:00')]
+                ];
+                
+                $actualRobots = TestAfterFetchRobots::find([
+                    'columns'=>['datetime'],
+                    'fireAfterFetch'=>true
```
sergeyklay: Could you please provide test with `'fireAfterFetch' => false`

Jurigag: Okay. Im also doing some optimaztions. Beacause there are no reason to fetch closure each time and also we can use array_diff_key and then unset them instead of intersect.

PR #11631, opened 2016-04-06
Fix FilterFloatTest\testSanitizeFloatHex()

googlle:


Jurigag:
You forgot about ```FilterIntegerTest``` too.

PR #11632, opened 2016-04-06
Fix FilterFloatTest\testSanitizeFloatHex()

googlle:


Jurigag:
Just if you wonder why even change it @sergeyklay - to fix those tests in php 7.

sergeyklay:
Thanks

PR #11648, opened 2016-04-09
Fixed Beanstalk default tube & optimized connection

sergeyklay:


phalcon/queue/beanstalk.zep
```diff
@@ -71,7 +71,7 @@ class Beanstalk
 	 * Default tube name
 	 * @const string
 	 */
-	const DEFAULT_TUBE = 'default';
+	const DEFAULT_TUBE = "default";
```
sergeyklay: @Chameleon-m

PR #11651, opened 2016-04-09
Added ability for loading aliased namespaces in manager

Jurigag:
With this ability we can define our relations using namespace aliases.

It's just loading proper models instead of throwing that some class don't exists.


Jurigag:
Well im not sure but maybe i should use addBelongsTo and use normal relations in class - it should work this way too ?

Also is it fine that i like copy-pasted Albums and Aritists from old tests structure ? In future i guess we are going to remove unit-tests in favour of tests ?

@sergeyklay

Jurigag:
Well fail is not related, it passed on my VM. So mention me when it will be fixed.

sergeyklay:
Yes, in future we will remove `unit-tests` in favor of `tests`.

Jurigag:
@sergeyklay 
Test added if you didn't notice.

sergeyklay:
Thanks

PR #11663, opened 2016-04-13
Sqlite data types

basilfx:
The lists of column types has been extended. However, not all column types conversions exist for SQLite.

This PR adds support for the additional column types. While SQLite has no statically typed columns, they have dynamic typing and use the column name for determining the type affinity. The rules are explained in [Section 2.0](http://www.sqlite.org/datatype3.html). I have chosen column names that match MySQL the most, while still adhering to the type affinity rules. This way, as little as possible information is lost when converting from SQLite to another DBMS.

I also extended the dialect tests.

Together with https://github.com/phalcon/phalcon-devtools/pull/639, I have been using this patch for almost two weeks without problems. I successfully converted from MySQL -> Phalcon Devtool Migrations -> SQLite (clean migration).


sergeyklay:
Could you please update CHANGELOG.md

basilfx:
Adapted `CHANGELOG.md`.

sergeyklay:
Thanks

PR #11695, opened 2016-04-21
Comment updated

quasipickle:
getTempName() is the _temporary_ name, not the _temporal_ name


sergeyklay:

quasipickle:
To which branch should I submit the pull request then?  2.0.x?

sergeyklay:
Could you please submit PR to the both branches? 2.0.x and 2.1.x

PR #11696, opened 2016-04-22
[RFC] Added autoescape() helper to PHQL Query Builder

SidRoberts:


Jurigag:
Well its adding not needed function calls. As i know less function calls = better for performance. But i guess in this case it will be like nothing.

sergeyklay:
Thanks

PR #11697, opened 2016-04-22
getTempName() comment fix

quasipickle:
getTempName means _temporary_ name, not _temporal_ name


quasipickle:
Whoops - did that wrong.

Issue #11709, opened 2016-04-24
[2.1.x] Service resolve raise a warning


ogarbe:
Hello,

These lines raises a warning in 2.1.x : https://github.com/phalcon/cphalcon/blob/2.1.x/phalcon/di/service.zep#L173-L178

I initialize all my code with error Exception when warning or notice : 

``` php
        /**
         * Set generic error handler
         */
        set_error_handler(
            function ($errno, $errstr, $errfile, $errline) {
                if (0 === error_reporting()) {
                    return false;
                }
                throw new \ErrorException($errstr, 0, $errno, $errfile, $errline);
            },
            E_ALL
        );

        /**
         * Set generic exception handler
         */
        set_exception_handler('handleException');
```

and when trying to resolve a closure, here is the error i face : 

``` shell
Fatal error: Uncaught exception 'ErrorException' with message 'Cannot bind an instance to a static closure'

ErrorException: Cannot bind an instance to a static closure in /home/olivier.garbé/www/lib_phalcon/Library/Loader.php on line 134

Call Stack:
    0.0467     714272   1. handleException() 
    0.0467     714416   2. Phalcon\Di->get()
    0.0467     714512   3. Phalcon\Di\Service->resolve() 
    0.0467     714648   4. Closure::bind()
```

In 2.0.x these lines didn't exists, hence no error.

andresgutierrez:
This happens because of this:

> Phalcon\Di is now bound to services closures allowing use Phalcon\Di as $this to access services within them

https://github.com/phalcon/cphalcon/blob/2.1.x/CHANGELOG.md

https://github.com/phalcon/cphalcon/issues/11029#issuecomment-200612702

ogarbe:
thanks andres, but how can we get rid of the warning ?

andresgutierrez:
Some service is returning a closure which was created in a static context thus generating the exception. To avoid this you have to avoid return a static-created closure to the di->resolve method.

ogarbe:
@andresgutierrez 

https://docs.phalconphp.com/en/latest/reference/di.html 

the use os static closure is in the documentation for Phalcon, it

```php
// Using an anonymous function
$di->set('logger', function () {
    return new LoggerFile('../apps/logs/error.log');
});
```

how would you change this (which is very handy) ?

andresgutierrez:
@ogarbe That example works for me in 2.1.x

sergeyklay:
@ogarbe ping

[...]

PR #11710, opened 2016-04-25
refs #11709 - Add static closure

ogarbe:
This PR permits to add static closure to service resolution hence backward compatibility with Phalcon 2.0 service definition.

### DIFF
--- a/phalcon/di/service.zep
+++ b/phalcon/di/service.zep
@@ -126,7 +126,7 @@ class Service implements ServiceInterface
 	public function resolve(parameters = null, <DiInterface> dependencyInjector = null)
 	{
 		boolean found;
-		var shared, definition, sharedInstance, instance, builder;
+		var shared, definition, sharedInstance, instance, builder, reflectionFunction;
 
 		let shared = this->_shared;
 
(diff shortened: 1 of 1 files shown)


Jurigag:
@sergeyklay 

I think that some test would be nice that everything is working as it should be.

phalcon/di/service.zep
```diff
@@ -174,7 +174,13 @@ class Service implements ServiceInterface
...
 					 */
 					if typeof dependencyInjector == "object" {
-						let definition = \Closure::bind(definition, dependencyInjector);
+						let reflectionFunction = new \ReflectionFunction(definition);
+    						if (
+							reflectionFunction->getClosureScopeClass() === null ||
+							reflectionFunction->getClosureThis() !== null
+    						) {
+							let definition = \Closure::bind(definition, dependencyInjector);
+						}
```
sergeyklay: else?

ogarbe: in the case we've given a static closure (which can't be bound), the use of $this in the closure won't work (not really a pb as this shouldn't be used inside a closure...), but the system will continue fo function as it was in Phalcon 2.0.

Jurigag: That's why there should be some test for this, in which cases we want $this in the closure, in which don't.

andresgutierrez:
Maybe implementing a Zephir optimizer to check this would be more performant than using Reflection

Jurigag:
Well, good to know it can be done there. You mean it will be checked in zend engine instead using php classes etc ?

andresgutierrez:
Yes, there must be an easy way in C to check whether the closure is static or not

Jurigag:
If someone can add it to zephir then great.

sergeyklay:
I apologize @ogarbe for completely overlooking this. Thank you for the patch. I know some people think this kinds of changes are trivial, but personally I believe it is important to get understanding right, including hypothetical risks and errors. We still need to test it for performance regression and may be we'll need to create an optimizer on the Zephir side but in general looks good to me. So thanks again for catching this mistake and fixing it, and again I am sorry about the delay. 

I'm closing this PR in favor of https://github.com/phalcon/cphalcon/pull/13395

PR #11718, opened 2016-04-27
Fix undefined method call ::getConnectionService

AlexKomrakov:
Fix php error "Call to undefined method Phalcon\Mvc\Collection\Manager::getconnectionservice()"
And Cphalcon undefined method call https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/collection.zep#L259

Now it works in accordance with the documentation: https://docs.phalconphp.com/ru/latest/api/Phalcon_Mvc_Collection.html


Jurigag:
Please squash your commits into one.

PR #11719, opened 2016-04-28
[2.0.x] Fix undefined method call ::getConnectionService

AlexKomrakov:
- Fixed undefined method call `Phalcon\Mvc\Collection\Manager::getConnectionService`

PR #11720, opened 2016-04-28
[2.1.x] Fix undefined method call ::getConnectionService

AlexKomrakov:
- Fixed undefined method call `Phalcon\Mvc\Collection\Manager::getConnectionService`

Same as [2.0.x] https://github.com/phalcon/cphalcon/pull/11719


sergeyklay:
Thanks

PR #11723, opened 2016-04-29
Reset variables(params and matches) between loops in Router::handle()

Fishdrowned:
params and matches are bind to each route, they should not suvive a loop, which may interfere in next loop


sergeyklay:
Thanks

PR #11727, opened 2016-04-29
Update micro.zep in order to fix 'after' Middleware not stopping execution flow..

stamster:
Reference: https://forum.phalconphp.com/discussion/6347/canceling-route-execution-with-micros-before-middleware


sergeyklay:

stamster:
OK Sergey. :+1: 

I don't commit that often, so this was quick try to fix this issue.

stamster:
Is there a way to edit pull request category?

SidRoberts:
@stamster, There isn't a way to change the branch. You'll have to checkout 2.1.x, manually recreate the commit and then create another pull request.

PR #11729, opened 2016-05-01
Allow implicit joins to be disabled in the global settings.

dschissler:
Some time ago I added the ability to disable implicit joins on the query level and this completes that work by allowing it to be globally disabled.


dschissler:
Ooops.  I forgot to define the variable.  Its a simple commit and I'll do this soon.

PR #11730, opened 2016-05-02
modify phalcon/mvc/model/metadata/redis.zep, __construct, add share r…

xkos:
when  i use session, modelsCache and modelsMetadata service, i found phalcon create multi redis handle;
redis of session use Backend redis handle, but redis of metadata create a new instance in construct, can't use exist reids. so i modify phalcon/mvc/model/metadata/redis.zep:

> if \Phalcon\Cache\Backend\Redis handle exist in options, use it, 
>    if not exist, it create a new one.

### DIFF
--- a/phalcon/mvc/model/metadata/redis.zep
+++ b/phalcon/mvc/model/metadata/redis.zep
@@ -46,47 +46,58 @@ class Redis extends MetaData
 
 	protected _ttl = 172800;
 
-	protected _redis = null;
+	protected _redis = null { get, set };
 
 	protected _metaData = [];
 
 	/**
 	 * Phalcon\Mvc\Model\MetaData\Redis constructor
 	 *
+	 * if redis exist, use exist redis handle
+	 * if not exist, create a new instance of redis
+	 *
 	 * @param array options
 	 */
 	public function __construct(options = null)
 	{
-		var ttl;
+
 
 		if typeof options != "array" {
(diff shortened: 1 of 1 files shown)


slilin:
add identify of optiosn["redis"]

phalcon/mvc/model/metadata/redis.zep
```diff
@@ -46,47 +46,58 @@ class Redis extends MetaData
...
+
 
 		if typeof options != "array" {
 			let options = [];
 		}
 
-		if !isset options["host"] {
-			let options["host"] = "127.0.0.1";
-		}
+		if isset options["redis"] && options["redis"] instanceof Redis {
```
sergeyklay: I prefer something like this:
```zephir
use use Phalcon\Cache\Backend\Redis as RedisBackend;

var redis;

if fetch redis, myArray["redis"] && typeof redis == "object" && redis instanceof RedisBackend {
    // ...
} else {
   let redis = ...
}

let this->_redis = redis;
```

sergeyklay (changes requested):
I think this PR should be more accurate. For example, metadata adapter can get a service instead of injecting a dependency through a constructor. Or user can pass service name instead instance of redis backend

sergeyklay:
I'm closing this PR due to the lack of any reaction.  If there are real needs, we can consider this in new PR.

PR #11741, opened 2016-05-04
PHP7 compatibility improvements

andresgutierrez:
The contents of this PR greatly improves stability running Phalcon on PHP7. There are still couple of bugs that are causing segfaults. On the other hand, the number of memory leaks have been reduced nearly to zero when running on ZendEngine3.
- PHQL/Volt/Annotations parsers have been rewritten to take advantage of the PHP7 memory layout
- Mvc\Url now passes its tests (fixed phalcon_replace_paths)
- Code making use of the `iterator` optimizer has been rewritten to avoid memory corruption (still researching the root cause)
- Fixed dependencies to external extensions
- Disabled tests that won't work due to unsupported extensions for php7.


Jurigag:
@andresgutierrez when i will have time i will check what's going on with tests that are not passing

PR #11750, opened 2016-05-05
Url regexp fix

### LINKED ISSUE #11715: UrlManager get
In Url manager there is regular expression to check is it local URL or not, and we got problem with colon in url, if colon exists in URL, then manager use it as not local. But it is not true. 

Here: 
https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/url.zep#L188
`preg_match("#^(//)|([a-z0-9]+://)|([a-z0-9]+:)#i", uri)`
Line begin here used only for //, not for all components of regexp. 

http://refiddle.com/8re2 here is example how this regexp process such data.

So it looks like it should be 
`preg_match("#^((//)|([a-z0-9]+://)|([a-z0-9]+:))#i", uri)`
To validate different schemas from begin of the line. 
And here is example http://refiddle.com/8ren

Newbas:
#11715


sergeyklay:
Could you please use this file for test
https://github.com/phalcon/cphalcon/blob/2.1.x/tests/unit/Mvc/UrlTest.php

Newbas:
@sergeyklay Thanks, i started it from master, then moved changes to this branch, but could not find UrlTest. Thanks Will update.

Newbas:
@sergeyklay changed. Also fixed typo in Readme file.

tests/unit/Mvc/UrlTest.php
```diff
@@ -244,6 +244,31 @@ function () {
             }
         );
     }
+    
+    /**
+     * Test urls that contains colos in schema definition and as parameter
```
DannyFeliz: Colons*

Newbas: Thanks. Fixed.

sergeyklay:
Thanks

PR #11761, opened 2016-05-08
Fix method's argument in Phalcon\Flash

Studentsov:
In the current implementation Phalcon\Flash it is impossible to transfer an array to the method `success()`. It is the strange and differs from behavior of other methods. This patch corrects it.


sergeyklay:
Thanks. Could you please check the 2.0.x branch?

PR #11764, opened 2016-05-09
Fix method's argument in Phalcon\Flash [2.0.x]

Studentsov:
In the current implementation Phalcon\Flash it is impossible to transfer an array to the method `success()`. It is the strange and differs from behavior of other methods. This patch corrects it.
Just as the https://github.com/phalcon/cphalcon/pull/11761


sergeyklay:
Thanks

PR #11765, opened 2016-05-09
Fixed Volt::callMacro

sergeyklay:


phalcon/mvc/view/engine/volt.zep
```diff
@@ -285,14 +284,14 @@ class Volt extends Engine implements EngineInterface
...
+	public function callMacro(string! name, array arguments = []) -> var
 	{
 		var macro;
 
 		if !fetch macro, this->_macros[name] {
 			throw new Exception("Macro '" . name . "' does not exist");
 		}
 
-		return call_user_func_array(macro, arguments);
+		return call_user_func(macro, arguments);
```
Jurigag: This will break functions https://github.com/phalcon/cphalcon/blob/2.1.x/phalcon/mvc/view/engine/volt/compiler.zep#L555

sergeyklay: Provide please script to reproduce bug

PR #11769, opened 2016-05-09
Fixes #11759 and deletes expired keys

### LINKED ISSUE #11759: FileCache not expiring
Here is the sample code:

``` php
$di->setShared('cache', function() {
    //Cache data for one hour by default
    $frontCache = new \Phalcon\Cache\Frontend\Data(array(
        'lifetime' => 3600
    ));

    //File cache settings
    $cache = new \Phalcon\Cache\Backend\File($frontCache, array(
        'cacheDir' => __DIR__ . '/cache/'
    ));

    return $cache;
});
```

then 

``` php
$resetAfter = $rate['saved'] + $seconds - time();
$cache->save($cacheName, $rate, (float)$resetAfter);
```

After a while `$resetAfter` becomes negative, and `$cache->exists($cacheName)` still returns the value...

edit: not touched the file for > 3600 seconds, and the file is still there. Looks like a bug for me.

stratoss:
Fixes #11759


phalcon/cache/backend/file.zep
```diff
@@ -121,32 +121,45 @@ class File extends Backend implements BackendInterface
...
+            		}
+            	}
+            } else {
+            	let ttl = (int) lifetime;
+            }
+            if !this->isValidArray(cachedContent, "created") {
+            	let createdTime = (int) filemtime(cacheFile);
+            } else {
+            	let createdTime = (int) cachedContent["created"];
+            }
```
sergeyklay: For indents we use tabs in zep files instead of spaces

stratoss: Usually my IDE is using tabs as well, should be the .zep extension. Will add another commit fixing the indents.

stratoss:
Any ideas why it is failing for 5.4?

stratoss:
Sorry about the multiple commits, but I was unable to run the tests locally. Now the file cache is being invalidated properly if the lifetime is less than default.

phalcon/cache/backend/file.zep
```diff
@@ -160,6 +173,11 @@ class File extends Backend implements BackendInterface
 					let ret = frontend->afterRetrieve(cachedContent);
 					return ret;
 				}
+			} else {
+				/**
+				 * As the content is expired, deleting the key
+				 */
+				this->delete(keyName);
```
sergeyklay: And `return ???`

stratoss: I don't think we should return something specific here. The function wasn't returning anything anyway, I've just included this delete call to make sure that the cache file is deleted.

sergeyklay: I believe that the function has something to return or not return anything at all. But now see at 174 line:
```zephir
return ret;
```

stratoss: Yes, if the cache is valid, it returns its value. Check out the current code in master: https://github.com/phalcon/cphalcon/blob/master/phalcon/cache/backend/file.zep#L163 

`if !(time() - ttl > modifiedTime) {` e.g. if the cache is not expired, return its value. There is no else clause, nor return after the if. I've just appended an else clause, where we delete the cache file from the filesystem as it is expired, and still not returning anything as we don't at the moment.

stratoss: I'm going to add few more lines and checks + few tests in the next hour.

stratoss:
I believe this should be it... returning the null is now more obvious and refusing to accept negative lifetime values. What do you think?

sergeyklay:
The `2.0.x` branch is not longer supported. Could you please send PR to the `2.1.x` branch?

sergeyklay:
@stratoss

stratoss:
I'm going to send PR to 2.1.x

PR #11774, opened 2016-05-10
Add ability to globally disable PHQL implicit joining

dschissler:
Ability to globally disable PHQL implicit joining based on defined model relationship keys.

### DIFF
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -4454,7 +4454,8 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 	{
 		var disableEvents, columnRenaming, notNullValidations,
 			exceptionOnFailedSave, phqlLiterals, virtualForeignKeys,
-			lateStateBinding, castOnHydrate, ignoreUnknownColumns;
+			lateStateBinding, castOnHydrate, ignoreUnknownColumns,
+			enableImplicitJoins;
 
 		/**
 		 * Enables/Disables globally the internal events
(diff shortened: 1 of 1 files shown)


dschissler:
This commit just adds a global option to a setting that I added a while ago.  I don't particularly see this being used a lot for 99% of projects but it brings the feature in line with the other ones that are offered by allowing it to be defined on both the global and per query level.  I added this feature as an escape valve when I was having lots of difficulties getting a EXISTS subquery to work properly.  I ended up fixing that bug in the Zephir code and since my head was into the problem I figured that I'd add the ability to ignore the potentially problematic implicit joining so that in the future if I'm doing an exotic query that I can construct the PHQL exactly the way that I want the SQL to look.  Thus this is would be essentially a midway point between raw SQL and PHQL.

By adding this feature we are enabling a new pessimistic sort of developer who is unable to use implicit joining whether that is a trust issue or it is simply broken for his needs.  I personally found it scary to realize that a complicated query might not be possible and hitting a bug of that nature is a pretty terrible blocker.

The next step would be to document this obscure feature.

sergeyklay:
Could you please amend PR with test which use `ini_set(...)`

dschissler:
I've never done that so this will probably sit for a while until I have time.  Also I'm not exactly sure how I would test it but perhaps by checking the SQL for a pattern (join condition...) that should or should not exist.

dschissler:
If this ends up sitting here for a while it won't be a big deal because it will be very easy to rebase.  It probably won't end up in tree until in 2.1.1.  At least the effort is documented now.

sjinks:

dschissler:
Likely nobody will use this as I'm just some weird guy who loves trouble.

PR #11782, opened 2016-05-12
Tidied up the Crypt

sergeyklay:


Green-Cat:
This has to be one of the best commit titles thus far.

Issue #11799, opened 2016-05-17
PHQL generating broken query for Oracle PDO if aliasing tables (extra AS in query)


Skydev0h:
Using PHQL to make a query with joins generates broken PDO query.
For example, next code:
```php
$res = $this->modelsManager->executeQuery('select p.payOrderPosId, o.payDate, a.lastName ' . 
    'from HL\\PayOrderPos p join HL\\PayOrder o on p.payOrderId = o.payOrderId ' .
            'join HL\\Anketa a on a.anketaId = o.anketaId limit 50');
```
Generates the following query:
```sql
SELECT * FROM (SELECT Z1.*, ROWNUM PHALCON_RN FROM 
(SELECT p.PAY_ORDER_POS_ID AS payOrderPosId,
o.PAY_DATE AS payDate, a.LAST_NAME AS lastName FROM PAY_ORDER_POS AS p  
INNER JOIN PAY_ORDER AS o ON p.PAY_ORDER_ID = o.PAY_ORDER_ID 
INNER JOIN ANKETA AS a ON a.ANKETA_ID = o.ANKETA_ID) 
Z1 WHERE ROWNUM <= 50)
```
```sql
SELECT * FROM (SELECT Z1.*, ROWNUM PHALCON_RN FROM (SELECT p.PAY_ORDER_POS_ID AS payOrderPosId, o.PAY_DATE AS payDate, a.LAST_NAME AS lastName FROM PAY_ORDER_POS AS p  INNER JOIN PAY_ORDER AS o ON p.PAY_ORDER_ID = o.PAY_ORDER_ID INNER JOIN ANKETA AS a ON a.ANKETA_ID = o.ANKETA_ID) Z1 WHERE ROWNUM <= 50)
```
which results in error:
SQLSTATE[HY000]: General error: 907 OCIStmtExecute: ORA-00907: missing right parenthesis (/root/build/php-5.6.17/ext/pdo_oci/oci_statement.c:148)

Trying to query without limit gives another error:
```php
$res = $this->modelsManager->executeQuery('select p.payOrderPosId, o.payDate, a.lastName ' . 
    'from HL\\PayOrderPos p join HL\\PayOrder o on p.payOrderId = o.payOrderId ' .
    'join HL\\Anketa a on a.anketaId = o.anketaId');
```
```sql
SELECT p.PAY_ORDER_POS_ID AS payOrderPosId, o.PAY_DATE AS payDate, a.LAST_NAME AS lastName 
[...]

PR #11800, opened 2016-05-17
Remove "AS" in prepareTable for Oracle dialect

Skydev0h:
In oracle dialect it is "TABLE alias", without the AS. AS is only used for variable aliasing.
I don't think it is possible without copying code from parent, because you need to remove the "AS" part.


Skydev0h:
#11800

Skydev0h:
#11799

sergeyklay:
Could you please update CHANGELOG?

Skydev0h:
@sergeyklay Done. I hope the CHANGELOG is alright now?

sergeyklay:
Thanks

Issue #10532, opened 2015-06-22
Unique validation fails: Column 'ROWCOUNT' doesn't make part of the column map


skaosYan:
Hi,

I have Phalcon 2.0.2 on Oracle 12c. Unique validation for a model fails with error.
Adding of ROWCOUNT into a column map doesn't help. How to fix this error?

```
Column 'ROWCOUNT' doesn't make part of the column map<br>#0 [internal function]: Phalcon\Mvc\Model::cloneResultMap(Object(Phalcon\Mvc\Model\Row), Array, Array, 0, false)<br />
#1 [internal function]: Phalcon\Mvc\Model\Resultset\Simple-&gt;current()<br />
#2 [internal function]: Phalcon\Mvc\Model\Resultset-&gt;getFirst()<br />
#3 [internal function]: Phalcon\Mvc\Model::_groupResult('COUNT', 'rowcount', Array)<br />
#4 [internal function]: Phalcon\Mvc\Model::count(Array)<br />
#5 [internal function]: Phalcon\Mvc\Model\Validator\Uniqueness-&gt;validate(Object(Vokuro\Models\UserCompany))<br />
#6 /home/var/www/html/v1/app/models/UserCompany.php(40): Phalcon\Mvc\Model-&gt;validate(Object(Phalcon\Mvc\Model\Validator\Uniqueness))<br />
#7 [internal function]: Vokuro\Models\UserCompany-&gt;validation()<br />
#8 [internal function]: Phalcon\Mvc\Model-&gt;fireEventCancel('validation')<br />
#9 [internal function]: Phalcon\Mvc\Model-&gt;_preSave(Object(Phalcon\Mvc\Model\MetaData\Memory), false, false)<br />
#10 /home/var/www/html/v1/app/controllers/UserController.php(144): Phalcon\Mvc\Model-&gt;save()<br />
#11 [internal function]: Vokuro\Controllers\UserController-&gt;addCompanyAction('155', '1')<br />
#12 [internal function]: Phalcon\Dispatcher-&gt;dispatch()<br />
#13 /home/var/www/html/v1/public/index.php(43): Phalcon\Mvc\Application-&gt;handle()<br />
#14 {main}
```

SQL log contains:

```
15 08:04:51 +0300][INFO] SELECT CASE WHEN COUNT(*) > 0 THEN 1 ELSE 0 END RET FROM ALL_TABLES WHERE TABLE_NAME='USER_COMPANY'
[...]

Issue #10490, opened 2015-06-09
model->findFirstBy... doesn't work with columnMap() in Oracle 


skaosYan:
Hi,

Phalcon 2.0.2, Oracle 12c.

I have set columnMap() for my model.
Then, I invoke find on my model like this:
Users::findFirstByEmail('test@test.com');

In SQL debug.log I see:

[Tue, 09 Jun 15 02:58:38 +0300][INFO] SELECT COUNT(*) "numrows" FROM (SELECT * FROM (SELECT Z1.*, ROWNUM PHALCON_RN FROM (SELECT USERS.USERID, USERS.NAME, USERS.EMAIL, USERS.PASSWORD, USERS.MUSTCHANGEPASSWORD, USERS.PROFILESID, USERS.SEX, USERS.PHONE, USERS.BIRTHDATE FROM USERS WHERE USERS.EMAIL = :0) Z1 WHERE ROWNUM <= 1))

But the call itself results in error:

Column 'PHALCON_RN' doesn't make part of the column map<br>#0 [internal function]: Phalcon\Mvc\Model::cloneResultMap(Object(Vokuro\Models\Users), Array, Array, 0, false)<br />
#1 [internal function]: Phalcon\Mvc\Model\Resultset\Simple-&gt;current()<br />
#2 [internal function]: Phalcon\Mvc\Model\Resultset-&gt;getFirst()<br />
#3 [internal function]: Phalcon\Mvc\Model\Query-&gt;execute(Array, NULL)<br />
#4 [internal function]: Phalcon\Mvc\Model::findFirst(Array)<br />
#5 /home/var/www/html/v1/app/controllers/UsersController.php(93): Phalcon\Mvc\Model::__callStatic('findFirstByEmai...', Array)<br />
#6 /home/var/www/html/v1/app/controllers/UsersController.php(93): Vokuro\Models\Users::findFirstByEmail('test@test.ru')<br />
#7 [internal function]: Vokuro\Controllers\UsersController-&gt;createAction()<br />
#8 [internal function]: Phalcon\Dispatcher-&gt;dispatch()<br />
#9 /home/var/www/html/v1/public/index.php(41): Phalcon\Mvc\Application-&gt;handle()<br />

How this bug can be fixed?

dreamsxin:
Please submit model code?

skaosYan:
I have fixed it by adding 	    'PHALCON_RN' => 'phalcon_rn', into the columnMap(). But this should be done att core level.


[...]

Issue #1652, opened 2013-12-07
[BUG] Phalcon\Mvc\Model::find() fails when using "columns" and "limit" params with Oracle


rlaffers:
~~I am using **Phalcon 1.2.1** (unable to upgrade at the moment but seeing the changelog up to 1.2.4, this issue has presumably not been addressed yet)~~
The issue confirmed in Phalcon 1.2.4 with **Oracle 10g** (release 10.2.0.3.0),             PDO::ATTR_CASE is set to PDO::CASE_LOWER.
Related issue: https://github.com/phalcon/cphalcon/issues/760

```php
$robots = Models\Robots::find(array(
    'columns' => 'id' 
));
```
Fails with the following error:
> Phalcon\Mvc\Model\Exception: Column 'id' doesn't belong to any of the selected models (1), when preparing: SELECT id FROM [MyApp\Models\Robots]

My db table is called "robots":
```
        ID   NAME                   TYPE                         YEAR   CURRENT_STATUS
----------   --------------------   --------------------   ----------   --------------------
         1   Robotina               mechanical                   1972   on
         2   Astro Boy              mechanical                   1952   off
         3   Terminator             cyborg                       2029   on

```

My model:
```php
namespace MyApp\Models;
class Robots extends \Phalcon\Mvc\Model {
    public function getSource() {
        return 'robots';
    }
}
```

According to the docs, this *ought* to work but does not.

A workaround is to specify $columnMap in the model (which is not optimal and should not be necessary):

```php
class Robots extends \Phalcon\Mvc\Model {
    public $id;
    public $name;
    public $type;
    public $year;
    public $current_status;
    public $db_rownum;

    public function getSource() {
        return 'robots';
    }

    public function columnMap() {
        return array(
            'id' => 'id',
            'name' => 'name',
            'type' => 'type',
            'year' => 'year',
            'current_status' => 'current_status',
            'db_rownum' => 'db_rownum',
        );
    }
}
```
Now it works fine, however, it will fail again if I try:
```php
$robots = Models\Robots::find(array(
    'columns' => 'id' 
    'limit' => 1 
));
```
> Phalcon\Mvc\Model\Exception: Column "db_rownum" doesn't make part of the column map

Please note that I included db_rownum in the columnMap as suggested in https://github.com/phalcon/cphalcon/issues/760 . It helps only when:
```php
$robots = Models\Robots::find(array(
    'limit' => 1 
));
```
[...]

Issue #760, opened 2013-07-02
[BUG] Invalid Oracle Limiting


mingalevme:
$p = Package::findFirst(3);

**Exception: Column "DB_ROWNUM" doesn't make part of the column map**

Generates the SQL:

```sql
SELECT Z2.* FROM (SELECT Z1.*, ROWNUM DB_ROWNUM FROM ( SELECT [COLUMS] FROM "RADIO"."RADIO_PACKAGE" WHERE "RADIO_PACKAGE"."RADIO_PACKAGE_ID" = 3 ) Z1 ) Z2 WHERE Z2.DB_ROWNUM BETWEEN 1 AND 1
SELECT COUNT(*) AS "numrows" FROM (SELECT Z2.* FROM (SELECT Z1.*, ROWNUM DB_ROWNUM FROM ( SELECT [COLUMNS] FROM "RADIO"."RADIO_PACKAGE" WHERE "RADIO_PACKAGE"."RADIO_PACKAGE_ID" = 3 ) Z1 ) Z2 WHERE Z2.DB_ROWNUM BETWEEN 1 AND 1)
```

andresgutierrez:
This is fixed in the 2.0.x branch (Phalcon 2.0.4), you have to change a global ORM setting to allow unexpected columns:

```php
\Phalcon\Mvc\Model::setup(['ignoreUnknownColumns' => true]);
```

PR #11802, opened 2016-05-18
ORM: Case-insensitive column map option

Skydev0h:
Added a global ORM setting that allows to try to find a key in column map case-insensitively if it was not found by an exact match at first time. Can be a real fix for many Oracle bugs, because Phalcon does not expect that column names are converted to uppercase by Oracle in the query result.

This is a real fix for #10532, because orm.ignore_unknown_columns just suppresses error, while _groupResult will return NULL which may have unintended consequences, and of course Model::count() will not work in this case. This may be a fix for #1652, #760, #10490, but did not check all those cases yet.

### DIFF
--- a/config.json
+++ b/config.json
@@ -99,7 +99,11 @@
         "orm.ignore_unknown_columns": {
             "type": "bool",
             "default": false
-        }
+        },
+		"orm.try_ci_column_map": {
+			"type": "bool",
+			"default": false
+		}
     },
     "destructors": {
         "request": [
--- a/config.json
+++ b/config.json
@@ -100,10 +100,10 @@
             "type": "bool",
             "default": false
         },
-		"orm.try_ci_column_map": {
-			"type": "bool",
-			"default": false
-		}
+        "orm.try_ci_column_map": {
+            "type": "bool",
+            "default": false
+        }
     },
     "destructors": {
         "request": [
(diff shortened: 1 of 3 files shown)


config.json
```diff
@@ -99,7 +99,11 @@
         "orm.ignore_unknown_columns": {
             "type": "bool",
             "default": false
-        }
+        },
+		"orm.try_ci_column_map": {
+			"type": "bool",
+			"default": false
+		}
     },
```
sergeyklay: Fix please indents

phalcon/mvc/model.zep
```diff
@@ -494,6 +494,20 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
...
+	private static function columnMapCIResolve(var columnMap, var key) -> string
+	{
+		var cmKey;
+		for cmKey in array_keys(columnMap) {
+			if strtolower(cmKey) == strtolower(key) {
+				return cmKey;
+			}
+		}
+		return key;
+	}
```
sergeyklay: For indents we use tabs in zep files instead of spaces

sergeyklay:
Could you please rebase?

phalcon/mvc/model.zep
Skydev0h: @sergeyklay I think in zep diffs the tabs are there already :)

Skydev0h:
@sergeyklay Tried to do the rebase, merge and fixing the indents. 
Commits are a little mess at the moment, but I couldn't properly squash them.

sergeyklay:
@Skydev0h ping

niden:
Closed in favor of https://github.com/phalcon/cphalcon/pull/13569

PR #11806, opened 2016-05-19
[2.1.x] Fixed endless loop in Phalcon\Mvc\Model\Resultset::update()

BesedinSasha:


Jurigag:
@sergeyklay im not sure if there is any test right now, but some would be need to check if it's working with this change.

sergeyklay:
@Besedin86 Could you please explain a bit more - what do you fix here

Besedin86:
The code below will update first record in endless loop. There was no next() method call when update records. 

```php 
$robots = Robot::find();
$robots->update(['name' => 'New name']);
```

Jurigag:
Oh you're right. That's why we need tests for such a things. Also well, udpating resultset like this it's not the best idea to be honest.

Besedin86:
It's just a hotfix. I think for other task we need more refactor and tests. Anyway, I use foreach to fix this locally.

sergeyklay:
Thanks

PR #11796, opened 2016-05-17
[NFR] getJoins() method in \Phalcon\Mvc\Model\Query\Builder

BesedinSasha:
In some cases it's useful to have `getJoins()` method in Query Builder to check which joins are already present in Builder. 

Usage:

``` php
$queryBuilder = new Phalcon\Mvc\Model\Query\Builder();
$queryBuilder->from(Robots::class);
$queryBuilder->leftJoin(Parts::class,sprintf('%s.id = %s.robot_id',Robots::class,Parts::class));

...

$queryBuilder->leftJoin(Parts::class,sprintf('%s.id = %s.robot_id',Robots::class,Parts::class)); 
$result = $queryBuilder->getQuery()->execute(); // ooops... it will be error here

```

Therefore we can use `getJoins()` to see which joins are present in Builder to avoid this situation.

PR #11812, opened 2016-05-20
[2.0.x][NFR] added new method getJoins() to \Phalcon\Mvc\Model\Query\Builder

BesedinSasha:
The same as #11796 but for 2.0.x branch


sergeyklay:
We finished support of `2.0.x` branch

PR #11826, opened 2016-05-24
Changes to validation

Jurigag:
What i changed in this commit:
- Added option for adding multiple fields to validate/filters with one(same) validator/filter - also it checking for messages, labels etc, so if we have 10 `Phalcon\Validation\Validator\PresenceOf` validators - we can shorten it to one add method and use array of fields, messages etc, it's working for every option except `model` in `Phalcon\Validation\Validator\Uniqueness` validator(if i missed something then write it here)
- In `Phalcon\Validation\Validator\Uniqueness::isUniqueness()` i added getting Entity from validation if we don't provide any `model` so we can get rid of "unlogical" `model => $this` when we validating from model
- Changed checking for strict in `Phalcon\Validation\Validator\ExclusionIn` and `Phalcon\Validation\Validator\InclusionIn` - it was checking for boolean value before getting it, i think it was wrong behaviour

What needs to be done:
- [x] tests
- [x] update documentation in phpdoc (examples)
- [ ] performance tests ?


Besedin86:
I can't see in this PR how can I check combination of fields for uniqueness. It would be great if it will be possible.

Jurigag:
Just by:

```php
$validation->add(['field1','field2'], new Uniqueness);
```

Besedin86:
f you see [here](https://github.com/phalcon/cphalcon/pull/11826/files#diff-86920296fe08950334f3b3f6d672f12cR170), you just add validator for each field separately. But I ask about combination of fields: when `field1 = value1 and field2 = value2`, not `field1 = value1 or field2 = value2`. It think this is different things.

Jurigag:
But it's not OR, it's still and, validation i succesfull when all validatiors are successfull.

Besedin86:
See example:

```sql
CREATE TABLE IF NOT EXISTS `test_table` (
  `id` int(11) NOT NULL,
  `field_1` int(11) NOT NULL,
  `field_2` int(11) NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1;

INSERT INTO `test_table` (`id`, `field_1`, `field_2`) VALUES
	(1, 1, 1),
	(2, 1, 2),
	(3, 2, 1);
```

I want to check for uniqueness combination `field_1` and `field_2`. For example I want to add values 2 and 2.

```sql
INSERT INTO `test_table` (`id`, `field_1`, `field_2`) VALUES (4, 2, 2);
```

First validator will fails (I have 2 in record no. 3) and second validator will fails too (I have 2 in record no.2). But I have no record with values 2 2 simultaneously. Therefore you need build 
```sql
SELECT COUNT(*) FROM test_table WHERE field_1 = 2 AND field_2 = 2.
```

Jurigag:
Yea after some thinking i think i got it. Will do commit soon with fix for it.

Jurigag:
How except should work for combination of field ? It should be `OR` or `AND` ?

Besedin86:
As I wrote - AND.

Jurigag:
I mean this EXCEPT OPTION https://github.com/phalcon/cphalcon/blob/2.1.x/phalcon/validation/validator/uniqueness.zep#L106

For both cases AND ?

Jurigag:
@sergeyklay 

Im not sure but shouldn't here:

https://github.com/phalcon/cphalcon/blob/2.1.x/phalcon/validation/validator/uniqueness.zep#L102

Be: 
```
if !except {

}
else {

}
```

Besedin86:
I work around query 
```sql
select * from test_table 
where field_1= 2 and field_2 = 2 and field_1 <> 2 and field_2 <> 2
```
and it works as we expect.

Jurigag:
Okay. Just wasn't sure about logic. What about above ? Not sure if this except even works as it should be ? Anyone even tested it ?

Besedin86:
I've never used it. But I'm sure it will work correctly. One small moment: if we use set of fields - except must be array too.

Jurigag:
Oh so this except just check additionally against our values. Roger that.

Besedin86:
Also you have to check type of `except` param.

Jurigag:
Oh right, except can be just string too.

Besedin86:
If `except` is array, I think it preferred be indexed array to avoid mistakes.

Jurigag:
So basically only this:

```php
'except'=>[
'field1'=>['value1'],
'field2'=>['value2']
]
```

Yea, i think the same.

Besedin86:
I fixed my query. So final version is
```sql
select * from test_table 
where field_1= 2 and field_2 = 2 and # field values
      (field_1 <> 2 or field_2 <> 2) # except
```

This is right query.

Jurigag:
Yea i know it's right. But what IF someone want fields not to be array of values ? I think it's logically possible, current uniqueness validator don't allow such a case.

Besedin86:
The query will be the same

```sql
select * from test_table 
where (field_1= 2) and   # field values
      (field_1 <> 2)     # except
```

Jurigag:
You don't understand, i mean such a case:

```php
$validation->add('field1', new Uniqueness([
'except'=>['value1','value2']
]));
```

Now it doesn't works. We can only check against one value(why ?).

Besedin86:
Oh. So we can modify our script to generate such sql query
```sql
select count(*) from test_table 
where (field_1 = 2 and field_2 = 2) and # field values
      ((field_1 <> 1 and field_1 <> 3) or (field_2 <> 2)) # except
```
it exclude 1,2 and 3,2 combination of values.

Jurigag:
Yea exactly i mean this. So basically i will add abiltiy to except value for field can be array of value.(now it can be only ONE VALUE)

Besedin86:
So that we have such difference of formats:

```php
// single field, except - single value
$validation->add('field1', new Uniqueness([
'except'=> 'value1'
]));

// single field, except - array of values
$validation->add('field1', new Uniqueness([
'except'=> ['value1', 'value2']
]));

// set of fields, except - single value
$validation->add(['field1', 'field2'], new Uniqueness([
'except'=> ['field1' => 'value1']
]));

// set of fields, except - array of values
$validation->add(['field1', 'field2'], new Uniqueness([
'except'=> ['field1' => ['value1', 'value2']]
]));

// set of fields, except - mixed values
$validation->add(['field1', 'field2'], new Uniqueness([
'except'=> ['field1' => ['value1', 'value2'], 'field2' => 'value3']
]));

```

Besedin86:
Optimized variant of query
```sql
select count(*) from test_table 
where (field_1 = 2 and field_2 = 2) and # field values
      ((field_1 not in (1,3)) or (field_2 not in (2))) # except
```

Jurigag:
I think really it should be not in AND not in ? So basically there are always be ands ? Well from mysql point there is no difference in performance between NOT IN and <> i think.

Besedin86:
It will be harder to parse query with `and and and and....`

Jurigag:
Yea you right. I just didn't understand it correctly. Will do commit soon after quick testing of just uniqueness.

Jurigag:
Almost done, but what Phalcon\Mvc\Model\Message field property should contain if we have two fields ? Just array or joined fields ?

Besedin86:
I think joined fields with general message

Jurigag:
Well as i see in old validation it was array of fields.

Jurigag:
@Besedin86 

now uniqueness validator should as it should be

sergeyklay:
@Jurigag

Jurigag:
Well, i will try to finish it today and tomorrow.

Jurigag:
There are some validators with really many options. I hope i don't have to provide test for each case(each option) beacause it's just similar code ?

sergeyklay:
do as many as you can

Jurigag:
Can i change everywhere in validator docs array to [] ?

sergeyklay:
yes

Jurigag:
Done, only some performance tests to do to check if it affects performance(i doubt it).

phalcon/validation.zep
```diff
@@ -162,30 +207,63 @@ class Validation extends Injectable implements ValidationInterface
...
 	{
-		let this->_validators[] = [field, validator];
+		var f;
+		if typeof field == "array" {
+			// Uniqueness validator for combination of fields is handled differently
+			if validator instanceof Uniqueness {
+				let this->_uniquenessValidatorsCombinedFields[] = [field, validator];
+			}
+			else {
+				for f in field {
```
andresgutierrez: Could you please rename variable 'f' to 'singleField'?

phalcon/validation.zep
```diff
@@ -478,13 +570,26 @@ class Validation extends Injectable implements ValidationInterface
 	/**
 	 * Internal validations, if it returns true, then skip the current validator
 	 */
-	protected function preChecking(string field, <ValidatorInterface> validator) -> boolean
+	protected function preChecking(var field, <ValidatorInterface> validator) -> boolean
 	{
-		if validator->getOption("allowEmpty", false) {
-			if method_exists(validator, "isAllowEmpty") {
-				return validator->isAllowEmpty(this, field);
+		var f;
```
andresgutierrez: Same here

Jurigag: Changed this f and e too everywhere they were(i hope).

tests/unit/Validation/Validator/BetweenTest.php
```diff
@@ -0,0 +1,92 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: Wojtek
+ * Date: 2016-06-05
+ * Time: 15:42
+ */
```
sergeyklay: @Jurigag Could you please add standard header here?

tests/unit/Validation/Validator/AlphaTest.php
```diff
@@ -0,0 +1,77 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: Wojtek
+ * Date: 2016-06-05
+ * Time: 15:35
+ */
```
sergeyklay: Same here

tests/unit/Validation/Validator/AlnumTest.php
```diff
@@ -0,0 +1,81 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: Wojtek
+ * Date: 2016-06-05
+ * Time: 15:21
+ */
```
sergeyklay: Same here

tests/unit/Validation/Validator/ConfirmationTest.php
```diff
@@ -0,0 +1,87 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: Wojtek
+ * Date: 2016-06-05
+ * Time: 16:23
+ */
```
sergeyklay: Same here

tests/unit/Validation/Validator/CreditCardTest.php
```diff
@@ -0,0 +1,81 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: Wojtek
+ * Date: 2016-06-05
+ * Time: 17:47
+ */
```
sergeyklay: Same here

tests/unit/Validation/Validator/DateTest.php
```diff
@@ -0,0 +1,85 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: Wojtek
+ * Date: 2016-06-05
+ * Time: 17:58
+ */
```
sergeyklay: Same here

tests/unit/Validation/Validator/DigitTest.php
```diff
@@ -0,0 +1,81 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: Wojtek
+ * Date: 2016-06-05
+ * Time: 18:08
+ */
```
sergeyklay: Same here

tests/unit/Validation/Validator/EmailTest.php
```diff
@@ -0,0 +1,81 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: Wojtek
+ * Date: 2016-06-05
+ * Time: 18:15
+ */
```
sergeyklay: Same here

sergeyklay:

Jurigag:
Heh, i totally forgot about those phpstorm comments. Done.

sergeyklay:
Thanks

Issue #11134, opened 2015-11-13
Phalcon\Mvc\Collection support for _id as array


challet:
When a mongo collection has its _ids made of arrays, for instance the result of a [group](https://docs.mongodb.org/v3.0/reference/operator/aggregation/group/) aggregation, it causes the `Collection::find` method to throw a "Array to string conversion" notice on the [iterator_to_array function call](https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/collection.zep#L432) ~~then a "You cannot serialize or unserialize PDO instances" PDOException~~ (**edit** : not part of the case).

How would it work with the [use_keys parameter](http://php.net/iterator_to_array#use_keys) set to false ? since the key doesn't seem to be used later in the code. By the way the `Collection::findFirst` call, which returns just before this operation, handles it correctly and makes it available as an array under `$result->_id`.

challet:
script to test the described beahvior

```php
<?php 
use Phalcon\Mvc\Collection as Collection;

class TestDocument extends Collection {
  
  public function getSource() {
    return "test";
  }
  
  public function initialize() {
    $this->useImplicitObjectIds(false);
  }

}

// ------------------------------------------
$doc = new TestDocument();
$doc->_id = array(
  'a' => 1,
  'b' => 1
);
$doc->save();

var_dump(TestDocument::find());
```

it leads to : 
```
Notice: Array to string conversion in test.php on line 8
Call Stack:
[...]
    0.0459     378464   6. Phalcon\Mvc\Collection::find() test.php:8
[...]

PR #11840, opened 2016-05-29
Collection: don't keep the raw keys when converting to array

challet:
These keys are not used anywhere further in the code. 
With the default setting as true and the use of array in MongoDb ids (as allowed through the use of `Phalcon\Mvc\Collection->useImplicitObjectIds` ), it can lead to an "Array to string conversion" notice.
See also #11134 and [the comment about the use_keys parameter](http://php.net/iterator_to_array#use_keys).


sergeyklay:
Thanks

PR #11844, opened 2016-05-31
Amended the code example for Phalcon\Mvc\Model\Message

CameronHall:
This should make it look more like PHP :)


sergeyklay:
We don't add changes directly to stable branches or `master`. Could you please send PR to the `2.1.x` branch?

CameronHall:
Whoops! Sorry about that, I'll create another for the other branch :)

PR #11846, opened 2016-05-31
Update escaper.zep

jormaj:
Allow setting whether double encoding takes place when escaping values.


sergeyklay:
We don't add changes directly to stable branches or master. Could you please send PR to the 2.1.x branch?

PR #11855, opened 2016-06-06
Update redis.zep

zhangshj:
prefix


sergeyklay:
1. We don't add changes directly to stable branches or master.
2. See `public function get(keyName, lifetime = null)`:
```zephir
	public function get(keyName, lifetime = null)
	{

		var redis, frontend, prefix, lastKey, cachedContent;
		...
		let prefix = this->_prefix;
		let lastKey = "_PHCR" . prefix . keyName;
		...
	}
```

PR #11856, opened 2016-06-06
Update memcache.zep

zhangshj:
prefix


sergeyklay:
1. We don't add changes directly to stable branches or master.
2. Could you please explain a bit more, what exactly are you trying to fix

PR #11859, opened 2016-06-06
Allow to set double encoding setting

jormaj:
For some situations, double encoding causes problems (f.ex. &#39; becomes &amp;#39;). Being able to disable double encoding solves this problem.


sergeyklay:
@jdrmar Could you please update  `CHANGELOG.md`

sergeyklay:
Thanks

PR #11874, opened 2016-06-11
FIX #11862 oracle lastinsertid should return int instead of array

### LINKED ISSUE #11862: Oracle lastInsertId returning array - breaks saving related records
Oracle lastInsertId returns array of value instead of the actual value.
This breaks saving of related records.
image

To fix you need to change line `228` of file `phalcon/db/adapter/pdo/oracle.zep` to this:
```PHP
return this->fetchAll("SELECT " . sequenceName . ".CURRVAL FROM dual", \Phalcon\Db::FETCH_NUM)[0][0];
```
instead of
```PHP
return this->fetchAll("SELECT " . sequenceName . ".CURRVAL FROM dual", \Phalcon\Db::FETCH_NUM)[0];
```

tembem:


sergeyklay:
Thanks

PR #11878, opened 2016-06-13
addResourceAccess and dropResourceAccess check the wrong accessList

hinablue:
It should be `this->_accessList`.


sergeyklay:
We don't add changes directly to stable branches or `master`. Could you please send this PR to the `2.1.x` branch?

hinablue:
OK.

PR #11880, opened 2016-06-13
addResourceAccess and dropResourceAccess check the wrong accessList

hinablue:
It should be `this->_accessList` 🎱


sergeyklay:
Thanks

PR #11891, opened 2016-06-15
New Mailer Class

fbuselli:
New class with tools for sending mails


Jurigag:
We don't accept PRs to master. Can you send it to 2.1.x ? Also update changelod.md and squash commits.

phalcon/mailer.zep
```diff
@@ -0,0 +1,513 @@
...
+    *     'port'       => 587,                  // set the SMTP port for the GMAIL server
+    *     'username'   => "noreply@server.it",  // SMTP account username
+    *     'password'   => "password",           // SMTP account password
+    *     'secure'     => "tls",
+    *     ];
+    *
+    * // Create the object Mailer
+    * // set in mail the value of the config
+    * $objMail = new Mailer();
+    * $objMail->setParams($config);
```
Jurigag: How about constructor where we can pass `$config` ?

phalcon/mailer.zep
```diff
@@ -0,0 +1,513 @@
...
+    * $objMail->addTo('prueba_to@address.com', 'pruba_name_to');
+    *
+    * // Send mail
+    * $objMail->send();
+    * </code>
+    *
+    **/
+
+class Mailer {
+	public server;
```
Jurigag: As far as i know we are using protected everywhere and getters/setters.

phalcon/mailer.zep
```diff
@@ -0,0 +1,513 @@
...
+	public username;
+	public password;
+	public secure;
+	
+	public SMTPAuth = false;
+	
+	public XMailer = "Phalcon Mailer";
+	public Priority = 3;
+	
+	public From = "";
```
Jurigag: Why this starts from upper case ?

phalcon/mailer.zep
```diff
@@ -0,0 +1,513 @@
...
+	 * Set parameters
+	 *
+	 * @param array   parameters
+	 * 
+	 */
+	public function setParams(var parameters){
+        var key, value;
+        
+        if (typeof parameters != "array") {
+            throw new Exception("Mail error: Parameters not array");
```
Jurigag: Create mailer exception class please.

phalcon/mailer.zep
```diff
@@ -0,0 +1,513 @@
...
+    }
+
+    /**
+	 * Set single parameter
+	 *
+	 * @param string  key
+	 * @param type_key  value
+	 * 
+	 */
+    public function setParam(var key, var value){
```
Jurigag: There should be 'string' key

phalcon/mailer.zep
```diff
@@ -0,0 +1,513 @@
...
+		this->getServerResponse();
+		if strtolower(trim(this->secure)) == "tls" {
+			this->sendServer(this->conn, "STARTTLS" . this->newline);
+			if substr(this->getServerResponse(),0,3) != "220" { return false; }
+			stream_context_set_option(this->conn, "ssl", "verify_peer", false);
+			stream_context_set_option(this->conn, "ssl", "verify_peer_name", false);
+			stream_context_set_option(this->conn, "ssl", "allow_self_signed", true);
+			stream_socket_enable_crypto(this->conn, true, STREAM_CRYPTO_METHOD_TLS_CLIENT);
+			this->sendServer(this->conn, "HELO " . this->localhost . this->newline);
+			if substr(this->getServerResponse(), 0, 3) != "250" { return false; }
```
Jurigag: I think we don't use if like this. Just do 

```
if substr(this->getServerResponse(), 0, 3) != "250" { 
    return false; 
}
```

phalcon/mailer.zep
```diff
@@ -0,0 +1,513 @@
...
+		return r;
+	}
+	
+	public function addTo(recipient, recipientName = "") {
+		let recipientName = trim(preg_replace("/[\r\n]+/", "", recipientName));
+		let this->recipients[] = "\"" . recipientName . "\" <" . recipient . ">";
+	}
+	
+	public function clearRecipients() {
+		unset(this->recipients);
```
Jurigag: Is those unset needed ?

phalcon/mailer.zep
```diff
@@ -0,0 +1,513 @@
...
+		
+		let parts .= "--" . boundary . "--";
+		
+		let message = ob_get_clean(); //Turn off output buffering
+		return parts;
+	}
+	
+	private function generateBoundary() {
+		return md5(uniqid(time()));
+		/*
```
Jurigag: remove those comments and others. Also i think we have some service for generating something like this.

Jurigag:
There's a lot to do here to be honest, more than i commented.

nsossonko:
This would be awesome, we currently use PHPMailer, but would be much better to use Phalcon native mailer...

phalcon/mailer.zep
```diff
@@ -0,0 +1,624 @@
...
+    /**
+     * Set parameters
+     *
+     * @param array   parameters
+     * 
+     */
+    public function setParams(var parameters) -> void {
+        var key, value;
+        
+        if (typeof parameters != "array") {
```
Jurigag: Instead of this if and exception just change `var parameters` to `array parameters`.

phalcon/mailer.zep
```diff
@@ -0,0 +1,624 @@
...
+    
+
+    /**
+     * Connect to the server
+     * return status conection
+     * 
+     */
+    private function connect() -> boolean {
+        if strtolower(trim(this->secure)) == "ssl" {
+            let this->server = "ssl://" . this->server;
```
Jurigag: Not sure but it's better to store any value of this->property or this->method() in zephir to variable before using it, especially multiple times.

phalcon/mailer.zep
```diff
@@ -0,0 +1,624 @@
...
+            if (substr(this->getServerResponse(), 0, 3) != "334") { return false; }
+            this->sendServer(this->conn, base64_encode(this->password) . this->newline);
+            if (substr(this->getServerResponse(), 0, 3) != "235") { return false; }
+        }
+        return true;
+    }
+
+    /**
+     * Format date
+     *
```
Jurigag: This blank lines before `*/` are not needed

phalcon/mailer.zep
```diff
@@ -0,0 +1,624 @@
...
+        let this->replyToName = name;
+    }
+    
+    /**
+     * Set message
+     *
+     * @param string  pMessage
+     * 
+     */
+    public function msg(pMessage) -> void {
```
Jurigag: Well if it should be string according to doc then set it in method too.

phalcon/mailer.zep
```diff
@@ -0,0 +1,624 @@
...
+    private conn;
+    private newline = "\r\n";
+    private localhost = "localhost";
+    private timeout = "15";
+    private debug = false;
+    
+    /**
+     * Constructor
+     */
+    public function __construct(var parameters){
```
sergeyklay: Same here:
```zephir
public function __construct(array parameters = [])
{
    this->setParams(parameters);
}
```

phalcon/mailer.zep
```diff
@@ -0,0 +1,624 @@
...
+     */
+    public function setParams(var parameters) -> void {
+        var key, value;
+        
+        if (typeof parameters != "array") {
+            throw new Exception("Mail error: Parameters not array");
+        }
+        else {
+            for key, value in parameters {
+                if property_exists(this,key) {
```
sergeyklay: move this chech to the setParam method

phalcon/mailer.zep
```diff
@@ -0,0 +1,624 @@
...
+    /**
+     * Connect to the server
+     * return status conection
+     * 
+     */
+    private function connect() -> boolean {
+        if strtolower(trim(this->secure)) == "ssl" {
+            let this->server = "ssl://" . this->server;
+        }
+        let this->conn = fsockopen(this->server, this->port, null, null, this->timeout);
```
sergeyklay: add ability to use fsockopen or p fsockopen

phalcon/mailer.zep
```diff
@@ -0,0 +1,624 @@
...
+        if substr(this->getServerResponse(), 0, 3) != "220" { return false; }
```
sergeyklay: Use common code style please

phalcon/mailer.zep
```diff
@@ -0,0 +1,624 @@
...
+    }
+
+    /**
+     * Set From Mail
+     *
+     * @param string  address
+     * @param string  name
+     * 
+     */
+    public function setFrom(address, name = "") -> void {
```
sergeyklay: public function setFrom(string! address, string! name = "") -> Mailer
{
    // ...
    return this;
}

phalcon/mailer.zep
```diff
@@ -0,0 +1,624 @@
...
+    }
+    
+    /**
+     * Send the email message
+     * Set up the headers and message body with attachments if necessary
+     *
+     * return status envio
+     * 
+     */
+    public function send(headers = null) -> boolean {
```
sergeyklay: What abot Headers/HeadersBug class?

phalcon/mailer.zep
```diff
@@ -0,0 +1,624 @@
...
+        
+        let email  = "Date: " . this->RFCDate() . this->newline;
+        let email .= "From: \"" . this->fromName . "\" <" . this->from . ">". this->newline;
+        if this->replyTo {
+            let email .= "Reply-To: \"" . this->replyToName . "\" <" . this->replyTo . ">" . this->newline;
+        }
+        
+        let email .= this->setRecipients();
+        
+        if headers != null { let email .= headers . this->newline; }
```
sergeyklay: Use common code style please

phalcon/mailer.zep
```diff
@@ -0,0 +1,624 @@
...
+    * $objMail->addTo('prueba_to@address.com', 'pruba_name_to');
+    *
+    * // Send mail
+    * $objMail->send();
+    * </code>
+    *
+    **/
+
+class Mailer {
+    protected server {
```
Jurigag: I think we are keeping set, get on the same lines in other classes so:

```
protected server { set, get }
```

andresgutierrez:
We don't add changes to master, please submit your PR to 3.1.x.

Also, it would be better to have this functionality in the incubator until it gets enough mature for the core

michanismus:
Inventing the wheel again? Why you don't use SwiftMailer? http://swiftmailer.org 
Probably the best mailer out there...

PR #11897, opened 2016-06-16
Fix config.zep for php7

Jurigag:


Jurigag:
What a hell, what conflicts ?

sergeyklay:
Thanks

Issue #11427, opened 2016-02-17
[2.1.x] Model alias in getter function not accessible


michanismus:
Example: 
Defined relation with alias "roles" in Model named "User".

I defined a getter function named getRoles() where I want to access $this->roles results in error: Undefined property: User::$roles

The relation:

``` php
/**
 * @HasMany(
 *     "id",
 *     "Model\UserRole",
 *     "userId",
 *     {
 *         "alias": "roles"
 *     }
 * )
 */
```

The function:

``` php
    public function getRoles()
    {
        if ($this->getDirtyState() == self::DIRTY_STATE_PERSISTENT) {
            $roles = array();
            foreach ($this->roles as $role) {
                $roles[] = $role->getRole();
            }

            return $roles;
        }

        return $this->_roles;
    }
```

michanismus:
@sergeyklay I can use the alias anywhere else, but it is impossible to create/overwrite a getter function for the alias in the model. I use the latest 2.1.x branch. Error happens since your last 2.0.x merge.

michanismus:
Found the commit which changes the behaviour: 
https://github.com/phalcon/cphalcon/commit/6d17755858027929b6f98d925f3646731332329d

sergeyklay:
@nsossonko

bmoore:
I cannot find this commit in the 2.0.x history. Does this issue still exist with my changes?

Also, there is no restriction to create and utilize your own setter methods. This just provides magic so you can use $user->role = instead of $user->setRole($role);

bmoore:
I do see this commit on 2.1.x 

So this get function is in the user model? What is the visibility of the roles and _roles properties? Could you provide an expected and actual vardump ?

michanismus:
2.1.x branch and yes issue exists with your changes.
The problem is, I want to decide to use a setter or not.
And it breaks the getter function for aliases. Take a look at my function, the alias (not existing property) is not accessible inside the function anymore.

michanismus:
I have to recompile phalcon with your changes. Takes a couple of minutes, then I will provide a call stack...

michanismus:
Here is a call stack for the error:

```
ErrorException: Undefined property: Horus\AdminBundle\Model\User::$roles
 at Horus\Helper\ErrorHandler->handleError(User.php:206)
 at Horus\AdminBundle\Model\User->getRoles(User.php:206)
 at Horus\AdminBundle\Model\User->setRoles(User.php:219)
[...]

PR #11914, opened 2016-06-22
Fixed incorrect return value when set value to non-existing property

tembem:


Jurigag:
As you see tests are failing beacause of this change.

tembem:
Well, if this is the correct design from beginning, I will close the PR

Jurigag:
Actually the tests are failing beacause it's beacause of this change and it's logical, for me it could be really good change, beacause most of people NOW are using private/protected property to set anything even if they don't have setter which is so bad behaviour and should never happen. With this change they would need to create setter at least, and then why use protected/private property when you have setter. Just tests needs to be changed as well(model classes need setters).

But this change can break a lot of apps.

tembem:
@Jurigag couldn't agree more.

Any suggestions @sergeyklay @andresgutierrez ?

sergeyklay:
Refs #11286, #11427

Issue #11941, opened 2016-07-01
[BUG] \Phalcon\Session\Flash::getMessages();


tugrul:
## Test
```php
$flash = new \Phalcon\Session\Flash();
$flash->error('Error Message');
var_dump($flash->getMessages('success', false));
```
## Expected Result
```
array (size=0)
  empty
```
## Actual Result
```
array (size=1)
  'error' => 
    array (size=1)
      0 => string 'Error Message' (length=13)
```

stamster:
This issue is present only in 2.1.x branch?

sergeyklay:
Fixed in the `2.1.x` branch.

PR #11942, opened 2016-07-01
[BUG] \Phalcon\Session\Flash::getMessages() #11941

tugrul:
#11941

### DIFF
--- a/phalcon/flash/session.zep
+++ b/phalcon/flash/session.zep
@@ -68,7 +68,7 @@ class Session extends FlashBase implements FlashInterface, InjectionAwareInterfa
 		let messages = session->get("_flashMessages");
 
 		if typeof type == "string" {
-			if !isset(messages[type])) {
+			if !isset(messages[type]) {
 				return [];
 			}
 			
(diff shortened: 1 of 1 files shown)


phalcon/flash/session.zep
```diff
@@ -67,7 +67,11 @@ class Session extends FlashBase implements FlashInterface, InjectionAwareInterfa
 		let session = <SessionInterface> dependencyInjector->getShared("session");
 		let messages = session->get("_flashMessages");
 
-		if typeof type == "string" && isset(messages[type]) {
+		if typeof type == "string" {
+			if !isset(messages[type])) {
+				return [];
+			}
+			
 			if !fetch returnMessages, messages[type] {
```
Jurigag: Isn't this fetch supposed to do what this if above is doing ?

tugrul: @Jurigag I'm not familiar to code Zephir. I looked up behavior of `fetch` operator and you are right.

Following block possibly better:
```zephir
if typeof type == "string" {
    if fetch returnMessages, messages[type] {
        if remove === true {
            unset(messages[type]);
            session->set("_flashMessages", messages);
        }
        return returnMessages;
    }
    
    return [];
}
```

I'm going to commit this.

sergeyklay:
We don't add changes directly to stable branches or `master`. Could you please send this PR to the `2.1.x` branch?

sergeyklay:
Fixed in the `2.1.x` branch.

PR #11946, opened 2016-07-03
Fixed Phalcon\Session\Flash::getMessages

sergeyklay:
Refs https://github.com/phalcon/cphalcon/issues/11941

### DIFF
--- a/phalcon/flash/session.zep
+++ b/phalcon/flash/session.zep
@@ -3,7 +3,7 @@
  +------------------------------------------------------------------------+
  | Phalcon Framework                                                      |
  +------------------------------------------------------------------------+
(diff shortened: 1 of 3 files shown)


phalcon/flash/session.zep
```diff
@@ -44,14 +44,20 @@ class Session extends FlashBase implements FlashInterface
...
 				let returnMessages = [];
 			}
+
 			if remove === true {
 				unset(messages[type]);
 				session->set("_flashMessages", messages);
 			}
+
 			return returnMessages;
 		}
```
tugrul: Better implementation:
```zephir
if typeof type == "string" {
    if fetch returnMessages, messages[type] {
        if remove === true {
            unset(messages[type]);
            session->set("_flashMessages", messages);
        }
        return returnMessages;
    }

    return [];
}
```

Issue #11940, opened 2016-07-01
[NFW] getter for definitions from Phalcon\Mvc\Micro\LazyLoader


Mechzeit:
Currently when using a Micro application, you have no way of getting which controller is going to be executed with the action matched from the application handle() call - BEFORE execution.

eg. Event micro:beforeExecuteRoute the only way that I am aware of (due to no dispatcher) to get the current lined up controller & action to be executed is via 'getActiveHandler()' against the applicable, which returns an array with 2 keys 0 & 1.

```php
array (size=2)
  0 => 
    object(Phalcon\Mvc\Micro\LazyLoader)[118]
      protected '_handler' => null
      protected '_definition' => string 'ExampleController' (length=17)
  1 => string 'exampleAction' (length=13)
```

But there is no literal way to get the definition from the LazyLoader, based on https://github.com/phalcon/cphalcon/blob/cf3c1fe62184d36d24c543c0b4f9993ea8f2445b/phalcon/mvc/micro/lazyloader.zep

Alas, I propose that a simple method is introduced to the LazyLoader class to 'getDefinition'.

stamster:
So, basically you want to get definitions from Micro handler which uses Lazy loader?
I'll have to check out my project since I'm pretty much I did some crazy stuff at runtime to inject properties etc.

PR #11951, opened 2016-07-03
Micro LazyLoader now has a getter for the definition property

VeryStrongFingers:
As per issue #11940 I have added a getter to get definition, for the LazyLoader (Phalcon\Mvc\Micro\LazyLoader) class


sergeyklay:
Thanks

PR #11954, opened 2016-07-04
Update ini.zep

dred86:
i Try use for "_cast" return type "-> var" but this is the error:

```
Zephir\CompilerException: Returning type: bool but this type is not compatible with return-type hints declared in the method in /var/www/github/phalcon/config-ini/cphalcon/phalcon/config/adapter/ini.zep on line 141

                        return true;
        ---------------------------^
```

### DIFF
--- a/phalcon/config/adapter/ini.zep
+++ b/phalcon/config/adapter/ini.zep
@@ -62,7 +62,8 @@ class Ini extends Config
 	{
 		var iniConfig;
 
-		let iniConfig = parse_ini_file(filePath, true);
+		let iniConfig = parse_ini_file(filePath, true, INI_SCANNER_RAW);
+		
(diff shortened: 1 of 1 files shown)


phalcon/config/adapter/ini.zep
```diff
@@ -118,4 +121,45 @@ class Ini extends Config
 
 		return [key: this->_parseIniString(path, value)];
 	}
+	
+    /**
+     * We have to cast values manually because parse_ini_file() has a poor implementation.
+     *
+     * @param mixed ini The array casted by `parse_ini_file`
+     */
+    private function _cast(var ini) -> bool | null | double | int | string
```
Jurigag: Fix indentation in this method.

PR #11969, opened 2016-07-13
Improved Phalcon\Config\Adapter\Ini

dred86:


andresgutierrez:
Thanks

sergeyklay:
👍

dred86:
Thank you too

zytzagoo:
This breaks parsing option values that have php constants in them? (due to use of INI_SCANNER_RAW, in which case option values are not parsed according to http://php.net/manual/en/function.parse-ini-file.php)

This is a BC break (worked fine until 3.0) - might be a good idea to add it to the list of major incompatible changes somewhere

sergeyklay:
https://blog.phalconphp.com/post/phalcon-3-0-0-released

zytzagoo:
Hm, the blog post doesn't mention it (neither does the changelog)... 

Not really sure how to easily fix this yet, but I made a test which passes in 2.0.13 (and earlier) and fails in 3.0.0: https://github.com/zytzagoo/cphalcon/commit/cb5af1f7c9012734456381a2431e96e34b73a0bc

PR #11970, opened 2016-07-13
MySQL dialect has missing schema check on some describe methods.

davidsev:
In db/dialect/mysql.zep some of the describe methods query INFORMATION_SCHEMA without specifying a schema name, causing records from other schemas to be pulled through.

For example, calling viewExists("myview") will return true if "myview" exists on any database, as opposed to only looking at the current database as expected.

This pull request adds `TABLE_SCHEMA` checks to viewExists, listViews, describeReferences and tableOptions


Jurigag:
Update tests too, squash commits and update `changelog.md`

sergeyklay:
@davidsev  
The `2.1.x` branch is not longer supported. Could you please send this PR to the `3.0.x`?

sergeyklay:
Fixed in the `3.0.x` branch.
Thanks

PR #11971, opened 2016-07-13
Uniqueness validator with NULL values

basilfx:
This PR adds support for validating uniqueness of NULL values, including two tests.

Example usecase: model instances that are unique as long as they are not deleted using soft-delete behavior.

For the test cases, I added an extra field to the `robots` table, as the other fields were all `NOT NULL`. Fixed other test accordingly.


Jurigag:
Update changelog and squash commits if you can.

basilfx:
Squashed into one commit and updated changelog.

nsossonko:
This is great, thanks!

andresgutierrez:
Thanks

PR #11975, opened 2016-07-14
Fixed virtual foreign key check for single column key

ghost:
Hi,

I believe I have fixed the problem with virtual foreign key consisting of only single column. When allowNulls option was set to TRUE and such column/attribute was set to a valid value (was not NULL) no check for the existence of the related record was generated and it crashed on PDO Exception. Hence no Phalcon validation message was added to the model.

Unfortunately I do not have working copy of Zephyr compiler right now and cannot check the code or write proper test.
But my PR is a fix of a single typo so can anybody check it?

Thank you

### DIFF
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -1497,7 +1497,7 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 			position, bindParams, extraConditions, message, fields,
 			referencedFields, field, referencedModel, value, allowNulls;
 		int action, numberNull;
-		boolean error, validateWithNulls = false;
+		boolean error, validateWithNulls;
 
 		/**
 		 * Get the models manager
@@ -1513,6 +1513,7 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 			let error = false;
 			for relation in belongsTo {
 
+				let validateWithNulls = false;
 				let foreignKey = relation->getForeignKey();
 				if foreignKey !== false {
 
@@ -1566,7 +1567,7 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 
 						} else {
 
-							fetch value, this->{field};
+							fetch value, this->{fields};
 							let conditions[] = "[" . referencedFields . "] = ?0",
 								bindParams[] = value;
 
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -1566,7 +1566,7 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 
 						} else {
 
-							fetch value, this->{fields};
+							fetch value, this->{field};
 							let conditions[] = "[" . referencedFields . "] = ?0",
 								bindParams[] = value;
 
--- a/phalcon/mvc/model/manager.zep
+++ b/phalcon/mvc/model/manager.zep
@@ -1129,7 +1129,7 @@ class Manager implements ManagerInterface, InjectionAwareInterface, EventsAwareI
 		var key, value, findParams;
 
 		if typeof findParamsOne == "string" && typeof findParamsTwo == "string" {
-			return ["(" . findParamsOne . ") AND " . findParamsTwo];
+			return ["(" . findParamsOne . ") AND (" . findParamsTwo . ")"];
 		}
 
 		let findParams = [];
@@ -1141,7 +1141,7 @@ class Manager implements ManagerInterface, InjectionAwareInterface, EventsAwareI
 					if !isset findParams[0] {
 						let findParams[0] = value;
 					} else {
-						let findParams[0] = "(" . findParams[0] . ") AND " . value;
+						let findParams[0] = "(" . findParams[0] . ") AND (" . value . ")";
 					}
 					continue;
 				}
@@ -1162,7 +1162,7 @@ class Manager implements ManagerInterface, InjectionAwareInterface, EventsAwareI
 					if !isset findParams[0] {
 						let findParams[0] = value;
 					} else {
-						let findParams[0] = "(" . findParams[0] . ") AND " . value;
+						let findParams[0] = "(" . findParams[0] . ") AND (" . value . ")";
 					}
 					continue;
 				}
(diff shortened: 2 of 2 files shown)


Jurigag:
As you see there is seg fault in tests.

CrNix:
Yeah I see, I guess I rushed the fix...
But the problem remains. The condition
```
if !validateWithNulls && !referencedModel->count([join(" AND ", conditions), "bind": bindParams]) {
```
is evaluated incorrectly somehow and database exception is thrown incorrectly. I'll try to get the Zephyr compiler up and running and try to look at that...

Jurigag:
But what exactly pdo exception did you have ?

CrNix:
Exactly this:
```
PDOException: SQLSTATE[23503]: Foreign key violation: 7 ERROR:  insert or update on table <table-name> violates foreign key constrain
```
This only happens then foreign key is using single column, like this:
````
$this->belongsTo(
			'userId',
			'App\Model\User',
			'id',
			[
				'alias' => 'user',
				"foreignKey" => [
					"allowNulls" => true,
					"message"    => "User does not exist"
				]
			]
		);
```
Now, when I set the attribute of the foreign key column to some non-existent value, PDO Exception is thrown without Phalcon generating the existence check.
But, when I define the column as an array, Phalcon checks it and creates the correct validation message on the model
```
$this->belongsTo(
			'[userId]', // Note the array here
			'App\Model\User',
			['id'],
			[
				'alias' => 'user',
				"foreignKey" => [
					"allowNulls" => true,
					"message"    => "User does not exist"
				]
			]
		);
```

Jurigag:
Can you post some code to reproduce a problem ? Never had something like this.

CrNix:
So this is the minimal example I could get to work. Might need a little tweak to work for your case. Let's have these 3 tables (By the way, using PostgreSQL):
````
CREATE TABLE public.robot (
  id SERIAL,
  
  PRIMARY KEY (id)
);

CREATE TABLE public.head (
  id SERIAL,
  
  PRIMARY KEY (id)
);


CREATE TABLE public.part (
  id SERIAL,
  robot_id INT,
  head_id INT,
  
  PRIMARY KEY (id),
  FOREIGN KEY (robot_id) REFERENCES public.robot(id) ON DELETE CASCADE ON UPDATE CASCADE,
  FOREIGN KEY (head_id) REFERENCES head(id) ON DELETE CASCADE ON UPDATE CASCADE
);
````
And 3 models for these tables:

***Robot.php***
````
<?php
namespace App\Model;

class Robot extends \Phalcon\Mvc\Model
{

    protected $id;

	public function initialize()
	{
		$this->setSchema('public');
		$this->setSource('robot');
	}

    public function columnMap()
    {
        return [
        	'id' => 'id',
        ];
    }

	public function setId($value)
	{
		$this->id = $value;

		return $this;
	}

	public function getId()
	{
		return $this->id;
	}
}
````
***Head.php***
````
<?php
namespace App\Model;

class Head extends \Phalcon\Mvc\Model
{

    protected $id;

	public function initialize()
	{
		$this->setSchema('public');
		$this->setSource('head');
	}

    public function columnMap()
    {
        return [
        	'id' => 'id',
        ];
    }

	public function setId($value)
	{
		$this->id = $value;

		return $this;
	}

	public function getId()
	{
		return $this->id;
	}
}
````
and ***Part.php***
````
<?php
namespace App\Model;

class Part extends \Phalcon\Mvc\Model
{

    protected $id;
	protected $robotId;
	protected $headId;

	public function initialize()
	{
		$this->setSchema('public');
		$this->setSource('part');

		$this->belongsTo(
			'robotId',
			'\App\Model\Robot',
			'id',
			[
				'alias' => 'robot',
				"foreignKey" => [
                    "allowNulls" => true,
                    "message"    => "Robot does not exists"
                ]
			]
		);

		$this->belongsTo(
			'headId',
			'\App\Model\Head',
			'id',
			[
				'alias' => 'head',
				"foreignKey" => [
                    "allowNulls" => true,
                    "message"    => "Head does not exists"
                ]
			]
		);
	}

    public function columnMap()
    {
        return [
        	'id' => 'id',
        	'robot_id' => 'robotId',
        	'head_id' => 'headId',
        ];
    }

	public function setId($value)
	{
		$this->id = $value;

		return $this;
	}

	public function getId()
	{
		return $this->id;
	}

	public function setRobotId($value)
	{
		$this->robotId = $value;

		return $this;
	}

	public function getRobotId()
	{
		return $this->robotId;
	}

	public function setHeadId($value)
	{
		$this->headId = $value;

		return $this;
	}

	public function getHeadId()
	{
		return $this->headId;
	}
}
````
(Note: I tried to place tables into different schemas or different combinations of schemas and so far it does have no effect on the outcome)

And here comes the funny part. Let's have exactly these records in the database:
````
Robot:
1/   id = 1
 2/  id = 2
  
Head:
1/   id = 1
 2/  id = 2
  
Part:
1/   id = 1, robot_id = 1, head_id = NULL
````
The NULL in the part table is pretty important.
Now, let's have this controller:

````
public function testAction()
{
	$part = \App\Model\Part::findFirst();

	$part->setHeadId(9999); // Invalid ID

	if (!$part->save()) {
		return [
			'error' => $part->getMessages()[0]->getMessage()
		];
	}

	return [
		'error' => false
	];
}
````
When head_id is originally NULL, PDO Exception with foreign key violation message is thrown.
But when you use this call instead, everything is fine.
````
$part->setRobotId(9999);
````
And what is the most funny part, is that when you change the order of the calls to the belongsTo(...) methods and you register first the relation for the *head* table it works as well!

Can't get the Phalcon compilation to work so far and I am clueless what can cause this.

**Comment by CrNix** [comment]
Final comment on this issue. I believe that setting this variable to false
````
validateWithNulls = false;
````
at the top of the function body, should be set at every iteration of the for loop instead. Otherwise value from the previous relation iteration stays set and could cause that funny behaviour with orderings of calls to belongsTo(...) methods inside model's initialize hook.

Can file another PR. But I am still unable to compile .zep files so I cannot check it.

File: phalcon/mvc/model.zep [hunk 1]
```diff
@@ -1497,7 +1497,7 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 			position, bindParams, extraConditions, message, fields,
 			referencedFields, field, referencedModel, value, allowNulls;
 		int action, numberNull;
-		boolean error, validateWithNulls = false;
```
sergeyklay: Could you please return back default value

sergeyklay:
@CrNix 
The `2.1.x` branch is not longer supported. Could you please submit this PR to the `3.0.x` branch?

PR #11976, opened 2016-07-14
The Events Manager now only uses Spl Priority Queues internally (no arrays)

SidRoberts:


Jurigag:
What's exactly the reason behind this change ?

SidRoberts:
It seems a bit pointless having two systems to store event listeners (with lots of duplicated code) when we just use SplPriorityQueue and optionally ignore priorities. This gets rid of ~100 lines of code without affecting usage whatsoever and makes it easier to maintain in the future.

Jurigag:
Also update changelog maybe ?

PR #11977, opened 2016-07-14
Added ability to check whether element has been defined in a form

Izopi4a:
Lets say we have a form for adding a product. Now this product has a type, and based on the type some of the fields are used, other are not. What I do is create 1 form class no matter what the type is and currently we have to perform the same check twice :

```
if ($product_type === "whatever" ) {
//add the field in the form
}

.... 

and then the same in our view
if ($product_type === "whatever" ) {
//display the field
}
```

probably you are saying yo yourself that we have to do the same with this implementation, and you are right, but we can skip cases like this :

```
if ($product_type === "car"  || $product_type === "bike"  || $product_type === "boat")
```

after half of out fields in our form. Also the display of the form can be dictated from form class only.

### DIFF
--- a/phalcon/forms/form.zep
+++ b/phalcon/forms/form.zep
@@ -710,4 +710,14 @@ class Form extends Injectable implements \Countable, \Iterator
 	{
 		return isset this->_elementsIndexed[this->_position];
 	}
+
+	/**
+	 * checks whether element with such a name exists in form definition
+	 *
+	 * @param string $name
+	 */
+	public function hasField(string name) -> boolean
+	{
+		return isset this->_elements[name];
+	}
 }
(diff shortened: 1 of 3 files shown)


unit-tests/FormsTest.php
```diff
@@ -540,4 +540,13 @@ public function testElementAppendMessage()
 		$element = new Select('test-select');
 		$element->appendMessage(new Message(''));
 	}
+
+	public function testFormHasField()
```
Jurigag: There is no need for this method here, in future `unit-tests` folder will be removed in favor of `tests`

phalcon/forms/form.zep
```diff
@@ -710,4 +710,15 @@ class Form extends Injectable implements \Countable, \Iterator
...
 		return isset this->_elementsIndexed[this->_position];
 	}
+
+    /**
+     * checks whether element with such a name exists in form definition
+     *
+     * @param string $name
+     * @return bool
+     */
+	public function hasField(string name) -> boolean
```
sergeyklay: Could you please fix indents? Also `@param`, `@return` not needed?

Izopi4a:
@sergeyklay should be better now

Jurigag:
@Izopi4a remove also this not needed test

Izopi4a:
meh... this method already exists https://github.com/phalcon/cphalcon/blob/master/phalcon/forms/form.zep#L632

sorry, didnt expect that function name, saw it just now by accident.

Jurigag:
Hahahhaha, but there are tests ? You could add them anyway.

Izopi4a:
I will have to make sure they don't exist before create a PR again :D

Jurigag:
Also i am wondering about this here - https://github.com/phalcon/cphalcon/blob/master/phalcon/forms/form.zep#L645

Is it really so much slower to call has method ? @andresgutierrez

PR #11981, opened 2016-07-15
Add firing of 'afterConnect' event to PDO adapter.

dschissler:
The event still is not able to be listened to because the connect function is called in the adapter constructor and so the events manager is unable to catch this early event. This event is being fired in the correct place though so future commits should be able to resolve this.

### DIFF
--- a/phalcon/db/adapter/pdo.zep
+++ b/phalcon/db/adapter/pdo.zep
@@ -59,7 +59,6 @@ abstract class Pdo extends Adapter
 	 */
 	public function __construct(array! descriptor)
 	{
-		this->connect(descriptor);
 		parent::__construct(descriptor);
 	}
 
@@ -177,7 +176,9 @@ abstract class Pdo extends Adapter
 	 */
 	public function prepare(string! sqlStatement) -> <\PDOStatement>
 	{
-		return this->_pdo->prepare(sqlStatement);
+		var pdo;
+		let pdo = this->getInternalHandler();
+		return pdo->prepare(sqlStatement);
 	}
 
 	/**
@@ -312,7 +313,7 @@ abstract class Pdo extends Adapter
 			}
 		}
 
-		let pdo = <\Pdo> this->_pdo;
+		let pdo = this->getInternalHandler();
 		if typeof bindParams == "array" {
 			let statement = pdo->prepare(sqlStatement);
 			if typeof statement == "object" {
@@ -367,7 +368,7 @@ abstract class Pdo extends Adapter
 		 */
 		let affectedRows = 0;
 
-		let pdo = <\Pdo> this->_pdo;
+		let pdo = this->getInternalHandler();
 		if typeof bindParams == "array" {
 			let statement = pdo->prepare(sqlStatement);
 			if typeof statement == "object" {
@@ -446,7 +447,9 @@ abstract class Pdo extends Adapter
 	 */
 	public function escapeString(string str) -> string
 	{
-		return this->_pdo->quote(str);
+		var pdo;
+		let pdo = this->getInternalHandler();
+		return pdo->quote(str);
 	}
 
 	/**
@@ -513,7 +516,7 @@ abstract class Pdo extends Adapter
 	public function lastInsertId(sequenceName = null) -> int | boolean
 	{
 		var pdo;
-		let pdo = this->_pdo;
+		let pdo = this->getInternalHandler();
 		if typeof pdo != "object" {
 			return false;
 		}
@@ -527,7 +530,7 @@ abstract class Pdo extends Adapter
 	{
 		var pdo, transactionLevel, eventsManager, savepointName;
 
-		let pdo = this->_pdo;
+		let pdo = this->getInternalHandler();
 		if typeof pdo != "object" {
 			return false;
 		}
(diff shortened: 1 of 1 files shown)


phalcon/db/adapter/pdo.zep
```diff
@@ -85,9 +85,12 @@ abstract class Pdo extends Adapter
 	 */
 	public function connect(descriptor = null)
 	{
-		var username, password, dsnParts, dsnAttributes,
+		var eventsManager,
+			username, password, dsnParts, dsnAttributes,
 			persistent, options, key, value;
 
+		let eventsManager = <ManagerInterface> this->_eventsManager;
```
Jurigag: Okay, but on this point `this->_eventsManager` always will be null, beacause we can't set it before `connect` method is called beacause it's called in constructor.

dschissler: True but the behaviour and location is correct as far as just the firing of the event.  It also then sets up workarounds like the one Andres suggested such as extending the parent and child adapters to prevent the connect from occurring in the constructor.

dschissler: Then we can talk about not having the connect done in the constructor and then doing several checks in the many functions that assume that there will always be a `_pdo` object.

sergeyklay: How about extending this class from `\Phalcon\Di\Injectable`
in order to get `eventsManager` from `this->_dependencyInjector`?

Jurigag: This is another idea, but what if someone don't want to use eventsManager from di and create seperated ? Correct me if i'm wrong, we can basically have one eventsManager for whole application for watching all fire events and dispatch them to listeners, this will mean that we will have less memory usage but i think this will cost more time(beacause it will cost more time to find proper listener when we will have a lot of them) or we can have many eventsManagers which will cost more memory but will be faster ?

dschissler: I think that we should try to keep the current way of being able to use the normal EM or creating a new one.

dschissler: What if `setEventsManager` was removed from the adapter and it had to be passed into the constructor.  Its a small breakage but then all events would always work.  Also people would be able to instantly see the error message about the missing method and then to read the upgrade guide.  That kind of breakage on upgrade is very easy to fix.

dschissler: It could even log a message or throw an exception to let people know what to change.  A nice deprecation notice doesn't work because then sometimes events are available and sometimes not.

sergeyklay: the `\Phalcon\Db\Adapter\Pdo` class extends from `\Phalcon\Db\Adapter`, but `\Phalcon\Db\Adapter` implements `\Phalcon\Events\EventsAwareInterface`. So you can modify `\Phalcon\Events\EventsAwareInterface:: getEventsManager` as follows:

```zephir
public function getEventsManager() -> <ManagerInterface>
{
    var eventsManager;

    let eventsManager = this->_eventsManager;
    if typeof eventsManager != "object" {
        let eventsManager = new \Phalcon\Events\Manager();

        // Or 
        // var di = \Phalcon\Di::getDefault();
        // let eventsManager = di->getShared('eventsManager');

        let this->_eventsManager = eventsManager;
    }

    return eventsManager;
}
```

dschissler: The problem is not having a new events manager it is rather not being able to attach to it early enough.  So this doesn't help.

```php
    if typeof eventsManager != "object" {
        let eventsManager = new \Phalcon\Events\Manager();
        let this->_eventsManager = eventsManager;
    }
```

This approach would help but then it seems to remove flexibility from a project that uses more than one DB service of different setups and even DB types since a plugin would need to attach to all global `db` events.  In this case the user would need to check for stuff in the plugin that it shouldn't know about and it would be bad design.

```php
    if typeof eventsManager != "object" {
        var di = \Phalcon\Di::getDefault();
        let eventsManager = di->getShared('eventsManager');
        let this->_eventsManager = eventsManager;
    }
```

phalcon/db/adapter/pdo.zep
```diff
@@ -153,6 +155,15 @@ abstract class Pdo extends Adapter
 		 * Create the connection using PDO
 		 */
 		let this->_pdo = new \Pdo(this->_type . ":" . dsnAttributes, username, password, options);
+
+		error_log("connect");
```
sergeyklay: ???

phalcon/db/adapter/pdo.zep
```diff
@@ -745,7 +762,14 @@ abstract class Pdo extends Adapter
 	 */
 	public function getInternalHandler() -> <\Pdo>
 	{
-		return this->_pdo;
+		var pdo;
+		let pdo = this->_pdo;
+		if typeof pdo != "object" {
+			error_log(typeof pdo);
```
sergeyklay: ???

dschissler: Yes this is a working copy.  Its an easy way for me to get Travis to run tests.  There are complications and I'm not 100% sure if I have all bugs worked out of Phalcon yet and my plugin needs to be redesigned.

dschissler: Hey!  So this is only passing on PHP 7, which is my environment.  What could this be to have it fail on 5.4-5.6?

sergeyklay:
@dschissler As you can see the tests fails

Jurigag:
He knows but he don't care about it beacause he don't use postgresql...

Issue #11726, opened 2016-04-29
hasOne() relationship not working properly in 2.0.x


makerlabs:
Lets say we have something like this:
User model:

```php
use Phalcon\Mvc\Model\Validator\Email as Email;

class User extends \Phalcon\Mvc\Model
{
    public $id;

    public $email;

    public $password;

    public function initialize()
    {
        $this->hasOne('id', 'UserProfile', 'user_id', array('alias' => 'UserProfile'));
    }

    public function getSource()
    {
        return 'user';
    }

}
```

UserProfile model:

```php
class UserProfile extends \Phalcon\Mvc\Model
{
    public $user_id;

    public $fullname;

    public function initialize()
    {
        $this->belongsTo('user_id', 'User', 'id', array('alias' => 'User'));
    }

    public function getSource()
    {
        return 'user_profile';
    }

}
```

And we try to assign two different User object instances:

```php
$profile = new \UserProfile();

/**
 * Our first object without email and password values 
 */
$profile->user = new \User();

/**
 * Our second object with email and password values 
 */
$user = new \User();
$user->email = 'test@test.com';
$user->password = '1234567890';

$profile->user = $user;

/* We check the values just to be sure */
echo "email=".$profile->user->email."<br>";
echo "password=".$profile->user->password."<br>";
 
$profile->fullname = 'Jan Kowalski';

if (!$profile->save()) {
    foreach ($profile->getMessages() as $m) {
        echo $m->getMessage()."<br>";
    }

}
else {
    echo "ok!";
}
```

Our result:
```
email=test@test.com
password=1234567890
email is required
password is required
```

It seems like phalcon is trying to save the first User object rather then the second one.

Jurigag:
```'alias' => 'User'```

But you are using 

```$profile->user```

makerlabs:
@Jurigag I use it properly u use the lowercase lookup the documentation. Actually in this case i could omit the alias because there is no conflict in the relations.

Jurigag:
Well then there must be mistake in documentation. It should be actually used as lowercase.

makerlabs:
@Jurigag ok but this isn't the issue here the issue here is the object assign not the case of the variable...

Jurigag:
Well actually im checking the source and it's always setting to lower so there is no difference.

makerlabs:
Yeah your resolution is working fine which is strange because accodring to the code it should work the same. Maybe it's more like a zephir bug in this case?

[...]

Issue #10800, opened 2015-08-14
Phalcon Model Relation Alias Not Caching with CamelCase


virgofx:
When using an alias for a related model (defined either via annotations or directly on the model via `initialize()`)  that is camel cased ... the corresponding `getRelated()` works; however, it does not cache the query on subsequent requests. 

When using a non-camel cased alias, the query is properly cached.

I have tested and confirmed this in my application explicitly on the `hasMany()`.  It probably exists as an issue on the other relations as well.

Pseudo example:

``` php
// Initialize database to log queries for reference  (replace {{VARIABLES}} as necessary)
$this->di->setShared('db', function() {

    $databaseConnection = new \Phalcon\Db\Adapter\Pdo\Mysql([
        'host'     => '{{HOST}}',
        'username' => '{{USERNAME}}',
        'password' => '{{PASSWORD}}',
        'dbname'   => '{{DBNAME}}'
    ]);
    $eventsManager = $this->di->getShared('eventsManager');
    $logger = new \Phalcon\Logger\Adapter\File('{{LOGFILE}}');
    $profiler = new \Phalcon\Db\Profiler();

    $eventsManager->attach('db:beforeQuery', function($event, $connection) use ($logger, $profiler) {
        $profiler->startProfile($connection->getSQLStatement());
    });

    $eventsManager->attach('db:afterQuery', function($event, $connection) use ($logger, $profiler) {
        $profiler->stopProfile();
        $logger->log(\Phalcon\Logger::INFO, $connection->getSQLStatement());
    });
    $databaseConnection->setEventsManager($eventsManager);
}
```

``` php
class Robot extends \Phalcon\Mvc\Model {

    public function initialize() {
        $this->hasMany("id", "RobotsParts", "parts_id", [
            'alias' => 'robotParts'
        ]);
        $this->hasMany("id", "RobotsParts", "parts_id", [
            'alias' => 'robotpartsnocamel'
        ]);
    }

    public function testGetRelated() {
        return $this->robotParts;
    }

    public function testGetRelatedNoCamel() {
        return $this->robotpartsnocamel;
    }
}
```

``` php
for ($i = 0; $i < 10; $i++) {
    /// ...
    $robotModel->testGetRelated();
}
// Database Queries ~ 10  (Not Expected, Expected = 1)

for ($i = 0; $i < 10; $i++) {
    /// ...
    $robotModel->testGetRelatedNoCamel();
}
// Database Queries ~ 1 (As expected)
```

[...]

PR #11983, opened 2016-07-16
Proposed Fix for Issue #10800

rudiservo:
Currently if you call a relation with CamelCase it will make a new fetch every and not pull the one from memory, this hapens even for alias in CamelCase.

i.e. $parts->robot != $parts->Robot

Refs #10800, #11726

### DIFF
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -4246,6 +4246,12 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 		let relation = <RelationInterface> manager->getRelationByAlias(modelName, lowerProperty);
 		if typeof relation == "object" {
 
+			/*
+			 Not fetch a relation if it is on CamelCase
+			 */
+			if isset this->{lowerProperty} && typeof this->{lowerProperty} == "object" {
+				return this->{lowerProperty};
+			}
 			/**
 			 * Get the related records
 			 */
--- a/tests/unit/Mvc/ModelTest.php
+++ b/tests/unit/Mvc/ModelTest.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace Phalcon\Test\Unit\Mvc;
+
+use Phalcon\Mvc\Model\Criteria;
+use Phalcon\Test\Models\AlbumORama\Albums;
+use Phalcon\Test\Module\UnitTest;
+
+/**
+ * \Phalcon\Test\Unit\Mvc\Model\ManagerTest
+ * Tests the Phalcon\Mvc\Model\Manager component
+ *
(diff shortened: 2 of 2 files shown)


unit-tests/ModelsRelationsTest.php
```diff
@@ -374,6 +374,13 @@ protected function _testIssue938($di)
 		}
 	}
 
+	protected function _testIssue10800()
```
Jurigag: This method here is not needed. In future `unit-tests` folder will be removed in favor of `tests`

rudiservo: yes I realize that, but I was unsure when that might happen or if it would break any tests so I added to both unit-tests and tests, in the future I will make changes only to tests.

Jurigag: Just remove it now when you will have time.

rudiservo: removed

nsossonko:
This is very much needed, seems like a great fix! Thanks

andresgutierrez:
Thanks

virgofx:
@rudiservo @Jurigag Is there any way we can add a test that confirms that #10800 gets fixed as well as outlined in the issue? 

Essentially  -- make sure sure that only 1 database call happens on repeat getter for related item. The pseudo code in #10800 is pretty close.

Jurigag:
Before query and call how many requests happen if those queries are firing beforeQuery.

rudiservo:
@virgofx if you get a relation and change a value without doing save(); immediately after if you try and get the same relation with CamelCase, the value has to be the one you changed and not the one from db.

i.e.
$car->owner->name = 'NotAPerson';
$car->owner->name == $car->Owner->name;

Jurigag:
I mean count ofc :)

virgofx:
@rudiservo I'm referring specifically to Issue #10800 

1. Retrieve related item in application from database
2. Attempt to call getter multiple times (Ensure that database is only hit once on first retrieval)

Can we test for above ^^? Previously, with camel case aliases in the relation this is not the case. Because the cache would store lower and the subsequent getter would check the relation alias as is.

rudiservo:
@virgofx yes this is the fix for #10800, I also had this problem with my projects and it will run has expected.
and has I mention before if you change a properties of an object in a hasOne relation and not persist it, the object has to be the same with camelCase, the other way is to check how many times it will fire with log on beforeQuery on a iterator like for ($i = 0; $i < 10000; $i++) and check if you log 10000 times the same select query.
I have tested it and it works fine, only one query.

virgofx:
Ok cool, I'll try to test this week as well and close the issue once I've confirmed as well. Thanks

Issue #11987, opened 2016-07-19
[ModelManager] Merging find parameters


CrNix:
Hi,

when using getRelated() method in a model with non-trivial find paramaters, for example while using OR in a condition, it behaves strangely and bad query is generated for me.

I think is an error on these lines:
https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/model/manager.zep#L1130
https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/model/manager.zep#L1151

and I think that $value should be wrapped inside brackets as well. Because related records have this implicit equality condition on foreign key columns.

But when using OR keyword inside parameters when calling getRelated() instead of condition like this:
```
[
    'conditions' => '<foreign-key-cond> AND (<cond1> OR <cond2>)'
]
```
is generated only this
```
[
    'conditions' => '<foreign-key-cond> AND <cond1> OR <cond2>'
]
```
Which is obviously wrong. This can be fixed manually by including the brackets in the condition:

```
return $this->getRelated('relatedAlias', [
    'conditions' => '(<cond1> OR <cond2>)' // without brackets it doesn't work
]);
```
But from the users point of the view it should be better to do it automatically. Because users do not need to know about any of the implicit conditions.

Jurigag:
You can create PR with fix for it if you need it now.

CrNix:
OK, so here is the PR for review https://github.com/phalcon/cphalcon/pull/11990

sergeyklay:
Fixed in the `2.1.x` branch. 👍

PR #11990, opened 2016-07-19
Fix for issue #11987 - merging conditions in manager.zep _mergeFindParameters method

ghost:
Hi, 

here is the PR review for issue when merging conditions inside find parameters.
The problem was what brackets weren't generated around the second parameter of the AND expression, thus generating logically invalid conditions in some cases.

Refs #11987


Jurigag:
Update changelog too :)

CrNix:
I hope it's ok now :)

andresgutierrez:
Thanks

PR #11996, opened 2016-07-20
Compound relation fix #10368

### LINKED ISSUE #10368: [ORM] Relations to the same model with composite keys not working properly
Hi. I'm using Models in my project, and I have been struggling with one that references itself - a "parent" row. But this table has a composite key, which envolves an id and a language id. I defined the 2 relations as follows:

```php
$this->hasMany(array('id','language_id'),'Libraries\Models\Pages', array('parent_id','parent_language_id'),array('alias'=>'children'));

$this->belongsTo(array('parent_id','parent_language_id'),'Libraries\Models\Pages', array('id','language_id'),array('alias'=>'parent'));
```

However this does not work correctly. I have been over this multiple times and can't seem to find an error on my part. Did I miss something or is it actually a bug?

ysosiak:
Fixes issue #10368
**Issue description:** Relations are not working properly when using multiple columns to join models by different field names.
**Example:**
`$this->hasOne(['user_id', 'client_id'], 'SomeModel', ['user_id_2', 'client_id']);`

### DIFF
--- a/phalcon/mvc/model/manager.zep
+++ b/phalcon/mvc/model/manager.zep
@@ -1298,7 +1298,7 @@ class Manager implements ManagerInterface, InjectionAwareInterface, EventsAwareI
 			 * Compound relation
 			 */
 			let referencedFields = relation->getReferencedFields();
-			for refPosition, field in relation->getReferencedFields() {
+			for refPosition, field in relation->getFields() {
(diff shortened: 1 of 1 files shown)


sergeyklay:
@yura215 Could you please provide a small test case for this PR?

sergeyklay:
@yura215 ping

sergeyklay:
Thanks! Fixed in the `3.0.x` branch.

PR #12003, opened 2016-07-21
Renamed isSoapRequest to isSoap and isSecureRequest to isSecure

niden:
Keeping the interface/verbiage the same, renamed isSoapRequest to isSoap and isSecureRequest to isSecure. Left the existing functions as aliases (to be removed later on)


sergeyklay:
Thanks

PR #12016, opened 2016-07-25
3.0.0

andresgutierrez:


build/install
```diff
@@ -32,6 +32,11 @@ if [ "${PHP_FULL_VERSION:0:3}" == "5.3" ]; then
 	exit 1
 fi
 
+if [ "${PHP_FULL_VERSION:0:3}" == "5.4" ]; then
+	echo "php 5.4 is no longer supported"
```
sergeyklay: 👍

Jurigag: Is there any reason for not supporting 5.4 ? Like something that doesn't work with php 5.4 ?

Green-Cat: It has been EOLd nearly a year ago http://php.net/supported-versions.php

Issue #12005, opened 2016-07-22
[2.1.x] Uniqueness validator should convert values.


basilfx:
The Uniqueness validator executes a query on the database to see if it unique. But it uses the values from the model instance as-is. In my case, I convert high-level object to database-compatible values. For instance, I like to use `DateTime`, store 'structures' in JSON, and so on.

In the example below, the Uniqueness validator builds a query using `$this->deleted`, which is still a `DateTime` object because `beforeSave` is not yet invoked. This makes the query invalid, because PHP cannot convert a `DateTime` object into a string.

The simple fix would be to call `beforeSave()` before all of the validations run. This would ensure that a query is build using the values that would otherwise be submitted to the database. But I don't want that, because it could break other validators (e.g. ones that DO expect richer objects). After all, the validation classes are generic and not explicitly bound to the ORM. I could use them for non-ORM things.

Example model:

```php
class User extends Model {
    private $name;
    private $deleted;

    private function beforeSave()
    {
        if ($this->deleted !== null) {
            $this->deleted = $this->deleted->format("Y-m-d H:i:s");
        }
    }
    private function afterFetch()
    {
        if ($this->deleted !== null) {
            $this->deleted = new \DateTime($this->deleted);
        }
    }
    
    public function validation()
    {
        $validator = new Validation();
        $validator->add(
            ["name", "deleted"],
            new Uniqueness([
                "message" => "The name must be unique.",
            ])
        );
        
        return $this->validate($validator);
    }
}
```

Jurigag:
Wait a minute ! Isn't that's why we have `beforeValidationOnSave` ?!

basilfx:
That would not work if I want to have a custom validator that uses a `DateTime` while also use Uniqueness that requires a timestamp as string.

Jurigag:
Then just in this validator create datetime. Why beforeSave should be called if there is already beofreValidationOnSave ? Then both would to the same.....

basilfx:
I just found out that my fix wasn't actually working. Guess I had some source file not saved and thought it was working.

Let me rephrase completely, using an example from [this documentation page](https://docs.phalconphp.com/en/latest/reference/models.html).

```php
<?php

class Robots extends Model
{
    public $id;

    public $name;

    public $status;

    public function beforeSave()
    {
        // Convert the array into a string
        $this->status = join(',', $this->status);
    }

    public function afterFetch()
    {
        // Convert the string to an array
        $this->status = explode(',', $this->status);
[...]

PR #12017, opened 2016-07-25
Implement convert option to Uniqueness validator (#12005)

basilfx:
This PR proposes a fix for #12005.

I have only implemented this for the Uniqueness validator, because that is the only one that does a database lookup, which may need values to be converted to do the lookup.


Jurigag:
I guess send it to 3.0.x @sergeyklay ?

sergeyklay:
@basilfx The 2.1.x branch is not longer supported. Could you please send this PR to the 3.0.x?

basilfx:
Sure, I'll open a new PR.

PR #12030, opened 2016-07-26
Implement convert option to Uniqueness validator (#12005)

basilfx:
This PR proposes a fix for #12005, for Phalcon 3.0.x.

I have only implemented this for the Uniqueness validator, because that is the only one that does a database lookup, which may need values to be converted to do the lookup..


sergeyklay:
Thanks

Jurigag:
@basilfx also update docs repo when you will have time

PR #12039, opened 2016-07-27
Added totalCount config to Phalcon\Paginator\Adapter\QueryBuilder

Izopi4a:
I guess situations like this https://forum.phalconphp.com/discussion/12207/how-to-alter-count-logic-in-querybuilderpaginator in some high complex application,  are highly possible. 

With this patch developers can specify the exact number of total results ( by performing a query of their own before that )  without build-ed in counter being fired. 

Maybe another plus of this patch is sometimes people wants to count id, instead of *, with this patch they can do it the way they want without any dramatic changes to the framework itself.

//other ideas: 
It can be done by checking if the given value is_numeric, otherwise people can specify raw queries, or pass another builder, but i think it might end up being an overkill. 

example usage : 

```

paginator   = new \Phalcon\Paginator\Adapter\QueryBuilder(
   array(
         "builder"  => $builder,
          "limit" => 20,
          "page"  => $currentPage,
          "totalCount" => $counterQuery->execute()->total_rows
   )
);
```


sergeyklay:
@Izopi4a
1. As you can see tests fails
2. The 2.1.x branch is not longer supported. Could you please send this PR to the `3.0.x` ?

Izopi4a:
i messed up in the commits, i will create a new one with clear history :D

and sure i will put it in for 3.0

PR #12048, opened 2016-07-29
ExclusionIn and InclusionIn validators not checking option "strict"

jrobinsonuk:
ExclusionIn and InclusionIn validators not checking option "strict" properly. Fixed

Original placement of check is redundant


sergeyklay:
We don't add changes directly to stable branches or `master`. Could you please send this PR to the `3.0.x` branch?

Jurigag:
This is already fixed in 2.1.x/3.0.x

PR #12068, opened 2016-08-01
Fixed virtual foreign key check

ghost:
Hi,

I am reopening PR https://github.com/phalcon/cphalcon/pull/11975 for 3.0.x branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Fixed `Phalcon\Mvc\Model\Manager::getRelationRecords` to correct using multi relation column [#12035](https://github.com/phalcon/cphalcon/issues/12035)
 - Fixed `Phalcon\Acl\Resource`. Now it implements `Phalcon\Acl\ResourceInterface` [#11959](https://github.com/phalcon/cphalcon/issues/11959)
 - Fixed `save` method for all cache backends. Now it updates the `_lastKey` property correctly [#12050](https://github.com/phalcon/cphalcon/issues/12050)
+- Fixed virtual foreign key check when having multiple keys [#12071](https://github.com/phalcon/cphalcon/issues/12071)
 
 # [3.0.0](https://github.com/phalcon/cphalcon/releases/tag/v3.0.0) (2016-07-29)
 - PHP 5.3 and 5.4 are now fully deprecated
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -1493,7 +1493,7 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 			position, bindParams, extraConditions, message, fields,
 			referencedFields, field, referencedModel, value, allowNulls;
 		int action, numberNull;
-		boolean error, validateWithNulls = false;
+		boolean error, validateWithNulls;
 
 		/**
 		 * Get the models manager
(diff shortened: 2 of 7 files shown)


CrNix:
@sergeyklay I have returned the default value for _validateWithNulls_ variable, but it is set also in the for loop now.

Wouldn't it be more friendly to set it only once in the for loop? It's not used anywhere the outside of the for loop I guess, so assigning the value twice might not be that readable.

sergeyklay:
Your PR does not make sense right now because the `validateWithNulls` has been set previously, before the loop

CrNix:
I do not understand. That was originally the problem, when it was set before the loop.

Because when you have a model with more than one virtual foreign key, first foreign key check leaves the ```validateWithNulls``` set and might be incorrect for the next for loop iteration.

That is why I removed the default value and put the assignment inside the for loop.

Jurigag:
He just means if relation is null then if there is enabled notNullValidations it should pass.

sergeyklay:
Okay, could you please provide a small test which cover this case?

CrNix:
I'll definitely give it a try.

sergeyklay:
@CrNix I have investigated the code more details. Yes. You're right. It be more friendly to set it only once in the for loop. But in any case your PR need to be tested.

sergeyklay:
Also could you please find the issue or open new one with script to reproduce the problem and refer this PR to the issue

CrNix:
OK, so here is the issue with steps to reproduce this  #12071

CrNix:
Could you explain to me please or refer me to some description what is the difference between tests and unit-tests folders?

Also, where can I define the database tables for the test? Is it OK to add them to tests/_data/schemas/mysql/mysql.dump.sql?

sergeyklay:
> Could you explain to me please or refer me to some description what is the difference between tests and unit-tests folders?

Use always  the `tests` folder.  In the future all tests will be located in the the `tests` folder.

> Also, where can I define the database tables for the test? Is it OK to add them to tests/_data/schemas/mysql/mysql.dump.sql?

Yes. But don't forget to ammend `unit-tests/DbDescribeTest.php` ( for example https://github.com/phalcon/cphalcon/pull/12077/files). We have not had time to rewrite this test

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 - Fixed `Phalcon\Mvc\Model\Manager::getRelationRecords` to correct using multi relation column [#12035](https://github.com/phalcon/cphalcon/issues/12035)
 - Fixed `Phalcon\Acl\Resource`. Now it implements `Phalcon\Acl\ResourceInterface` [#11959](https://github.com/phalcon/cphalcon/issues/11959)
 - Fixed `save` method for all cache backends. Now it updates the `_lastKey` property correctly [#12050](https://github.com/phalcon/cphalcon/issues/12050)
+- Fixed virtual foreign key check when having multiple keys
```
sergeyklay: Refer to the issue please

sergeyklay:
@CrNix Add new tables here https://github.com/phalcon/cphalcon/blob/3.0.x/tests/unit/Db/Adapter/Pdo/MysqlTest.php#L58-L81

sergeyklay:
Thanks!

CrNix:
You are welcome

PR #12100, opened 2016-08-05
Add optional `mode` param to Phalcon\Config\Adapter\Ini ctor + test

zytzagoo:
`mode` defaults to INI_SCANNER_RAW as it was earlier


sergeyklay:
Thanks

PR #12101, opened 2016-08-06
Prepend Class Auto Loader

Fenikkusu:
The auto loader class currently appends itself when calling spl_autoloader_register. This can cause the Phalcon loader to not be the primary loader when working with other registered autoloaders. The best example is Composer. Composer, by default, prepends its autoloader, making it the first to run.

The applied changes makes use of the same concept to prepend the Phalcon loader. This, in combination with the recent 'registerFiles' method allows Phalcon to always be the primary loader.

For backwards compatability, the loader defaults to appending still, but can be converted to prepending by calling setPrepend(TRUE)

`$loader = new \Phalcon\Loader();`
`$loader->registerNamespaces(['SomeNamespace' => 'Some\Namespace'])`
`->setPrepend(TRUE)`
`->register();`

### DIFF
--- a/phalcon/loader.zep
+++ b/phalcon/loader.zep
@@ -66,6 +66,8 @@ class Loader implements EventsAwareInterface
 
 	protected _registered = false;
 
+	protected _prependLoader = false;
+
 	/**
 	 * Sets the events manager
 	 */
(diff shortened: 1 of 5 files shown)


Green-Cat:
We do not take pull requests to the master branch (it is the stable branch). 
Please resubmit your pull request to the 3.0.x branch.

Fenikkusu:
Updated/Rebased

https://github.com/phalcon/cphalcon/pull/12103

PR #12103, opened 2016-08-06
Prepend Class Auto Loader

Fenikkusu:
The auto loader class currently appends itself when calling spl_autoloader_register. This can cause the Phalcon loader to not be the primary loader when working with other registered autoloaders. The best example is Composer. Composer, by default, prepends its autoloader, making it the first to run.

The applied changes makes use of the same concept to prepend the Phalcon loader. This, in combination with the recent 'registerFiles' method allows Phalcon to always be the primary loader.

For backwards compatability, the loader defaults to appending still, but can be converted to prepending by calling setPrepend(TRUE)

$loader = new \Phalcon\Loader();
$loader->registerNamespaces(['SomeNamespace' => 'Some\Namespace'])
->setPrepend(TRUE)
->register();

### DIFF
--- a/tests/_data/vendor/Example/Test/AnotherTest.php
+++ b/tests/_data/vendor/Example/Test/AnotherTest.php
@@ -1,11 +1,4 @@
 <?php
-
-    /**
-     * Created by PhpStorm.
-     * User: fenikkusu
-     * Date: 8/5/16
-     * Time: 10:28 PM
-     */
     class AnotherTest {
 
     }
--- a/tests/_data/vendor/Example/Test/AnotherTest.php
+++ b/tests/_data/vendor/Example/Test/AnotherTest.php
@@ -1,0 +1,11 @@
+<?php
+
+    /**
+     * Created by PhpStorm.
+     * User: fenikkusu
+     * Date: 8/5/16
+     * Time: 10:28 PM
+     */
+    class AnotherTest {
+
+    }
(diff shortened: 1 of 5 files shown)


tests/_data/vendor/Example/Test/YetAnotherTest.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+    /**
+     * Created by PhpStorm.
+     * User: fenikkusu
+     * Date: 8/5/16
+     * Time: 10:28 PM
+     */
+    class YetAnotherTest {
+
```
sergeyklay: Could you please remove these useless comments

tests/_data/vendor/Example/Test/AnotherTest.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+    /**
+     * Created by PhpStorm.
+     * User: fenikkusu
+     * Date: 8/5/16
+     * Time: 10:28 PM
+     */
+    class AnotherTest {
```
sergeyklay: the same

Fenikkusu:
I've removed the auto-generated comments.

Fenikkusu:
I was curious as to why this was closed with no merger. Found the other PR doing the same thing in slightly different manner. Any particular reason that one was used? Given this had been waiting for 7 months prior, I'm slightly confused.

sergeyklay:
After deleting 3.0.x branch this issues was closed automatically because your PR is still directed to it.

Could you please submit your PR to the 3.2.x branch?

Fenikkusu:
I can, but didn't the PR from a month ago (#12580) render this null/void?

PR #12112, opened 2016-08-08
Filter Sanitize in Volt 

pedro151:
creation of the filter sanitize before volt compilation.


andresgutierrez:
Could you please include just the commits that belong to your PR?

sergeyklay:
Close in favor of #12133

PR #12133, opened 2016-08-12
filter sanitize before compiling volt

pedro151:


CameronHall:
@niden did you want me to deal with this one?

niden:
@CameronHall If you can yes please.

CameronHall:
I won't have be able to get this done _soon_. If you remind me what it's supposed to do, I can take a look at it and chip away.

sergeyklay:
If it is a fix for a real problem, let's apply it before beta release. But first I would like to see a script to reproduce the bug

niden:
The way I deciphered this since there was no description or rationale behind this, was to introduce a way to manipulate HTML after it has been compiled. The particular PR introduces this functionality and the supplied filter class strips whitespaces, line feeds etc., thus reducing the output on screen.

I do not think that this should be part of Phalcon. Services such as Varnish or Cloudflare can offer this functionality without having to change your code.

@pedro151 Thank you so much for the contribution, however I will close this. If later on the community is interested in something like this we can reintroduce this.

Issue #12120, opened 2016-08-10
Missing docs / API on some Validators


igorsantos07:
It seems there's a bunch of Validator classes that are not mentioned in [2.0](https://docs.phalconphp.com/en/2.0.0/reference/validation.html#validators) nor [3.0](https://docs.phalconphp.com/en/latest/reference/validation.html#validators) docs or [API listings](https://docs.phalconphp.com/en/2.0.0/api/index.html)... I found them because I thought they must exist and tried to auto-complete on PHPStorm using `phalcon-devtools`...

Examples:
- Date
- Alpha
- Alnum
- CreditCard
- Digit
- Identical
- File
- a bunch others

![screenshot_051](https://cloud.githubusercontent.com/assets/532299/17568371/da0b8b4c-5f19-11e6-8c58-33e90da7a929.jpg)

sergeyklay:
The `Phalcon\Mvc\Model\Validation` is now deprecated in favor of `Phalcon\Validation`
The functionality of both components is merged into one, allowing us to reduce the codebase while offering the same functionality as before.

**Old way:**
```php
namespace Invo\Models;

use Phalcon\Mvc\Model;
use Phalcon\Mvc\Model\Validator\Email as EmailValidator;
use Phalcon\Mvc\Model\Validator\Uniqueness as UniquenessValidator;

class Users extends Model
{
    public function validation()
    {
        $this->validate(
            new EmailValidator(
                [
                    'field' => 'email',
                ]
            )
        );

        $this->validate(
            new UniquenessValidator(
                [
                    'field'   => 'username',
                    'message' => 'Sorry, That username is already taken',
                ]
            )
        );

        if ($this->validationHasFailed() == true) {
            return false;
        }
    }
}
```

**New way:**
```php
namespace Invo\Models;

use Phalcon\Mvc\Model;
use Phalcon\Validation;
use Phalcon\Validation\Validator\Email as EmailValidator;
use Phalcon\Validation\Validator\Uniqueness as UniquenessValidator;

class Users extends Model
{
    public function validation()
    {
        $validator = new Validation();

        $validator->add(
            'email', //your field name
            new EmailValidator([
                'model' => $this,
                'message' => 'Please enter a correct email address'
            ])
        );

        $validator->add(
            'username',
            new UniquenessValidator([
                'model' => $this,
                'message' => 'Sorry, That username is already taken',
            ])
        );

        return $this->validate($validator);
    }
}
```

[...]

PR #12145, opened 2016-08-15
Pointing where doc issues should be sent to

igorsantos07:
This was surfaced by #12120

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,4 +1,5 @@
 > Questions should go to https://forum.phalconphp.com
+> Documentation issues should go to https://github.com/phalcon/docs/issues
 
 ### Expected and Actual Behavior
 


andresgutierrez:
Thanks, we don't add changes directly to master, could you please submit this to 3.0.x?

https://github.com/phalcon/cphalcon/blob/master/CONTRIBUTING.md#pull-request-checklist

igorsantos07:
Actually @andresgutierrez, I think this is a special case. Template files are only considered from the default branch - it seems to be `master` in this case, right?

From the [docs](https://help.github.com/articles/creating-an-issue-template-for-your-repository/):
> Tip: You must create your template in your repository's default branch. Templates created in other branches are not available for collaborators to use.

igorsantos07:
Bonus: what about a pull request template? something like "hey way, are you PRing to the correct branch? we don't use master for this type of work", maybe with about the same text from the PR checklist you linked.

andresgutierrez:
Changes will be available when we merge 3.0.x into master

igorsantos07:
Well, so until 3.0.x gets a release there will be no update to the issue template? The issue template has _nothing_ related to the actual project code or state or release, it's just a repository configuration.
It doesn't make much sense to have this type of fix waiting weeks on 3.0.x...

andresgutierrez:
> Well, so until 3.0.x gets a release there will be no update to the issue template? 

Exactly this

igorsantos07:
Cool. What's would be of work trying to progress without a lot of bureaucracy in front of it, right?
What would be of logic if there was no one thinking about stuff before repeating what has been said by others, right?
Obviously there's a whole lot of sense of leaving a small fix sitting for weeks, together with completely unrelated stuff. Oh well.

igorsantos07:
Funny to actually notice the very issue I'm trying to point out here, @andresgutierrez, is exactly what just happened. There **is** a PR template stating stuff should not be sent to master. But thanks to this crazy policy of tying repository settings to Phalcon versions, I didn't see that.

PR #12146, opened 2016-08-15
Issue and PR templates improvements

igorsantos07:
This was surfaced by #12120 and is the correct version of #12145.

I've synchronized what's in the PR template with the contributing guide.

IMHO these files should be cherry-picked into `master`, as this is not related to the next Phalcon release and would help other developers to send better issues/PRs into the repository. Bear in mind that those changes don't get applied to the repository until they land into `master`.


sergeyklay:
Thanks

Issue #12170, opened 2016-08-26
[3.0.0]Snapshot data removed after saving to cache adapter


Jurigag:
### Expected and Actual Behavior

The script should return:

```
array(11) { ["cedula"]=> string(1) "1" ["tipo_documento_id"]=> string(1) "3" ["nombres"]=> string(15) "HUANG ZHENGQUIN" ["telefono"]=> string(9) "191821112" ["direccion"]=> string(16) "CRA 25 CALLE 100" ["email"]=> string(9) "@yahoo.es" ["fecha_nacimiento"]=> string(10) "2011-02-03" ["ciudad_id"]=> string(6) "127591" ["creado_at"]=> string(10) "2011-05-18" ["cupo"]=> string(7) "6930.00" ["estado"]=> string(1) "I" } array(11) { ["cedula"]=> string(1) "1" ["tipo_documento_id"]=> string(1) "3" ["nombres"]=> string(15) "HUANG ZHENGQUIN" ["telefono"]=> string(9) "191821112" ["direccion"]=> string(16) "CRA 25 CALLE 100" ["email"]=> string(9) "@yahoo.es" ["fecha_nacimiento"]=> string(10) "2011-02-03" ["ciudad_id"]=> string(6) "127591" ["creado_at"]=> string(10) "2011-05-18" ["cupo"]=> string(7) "6930.00" ["estado"]=> string(1) "I" }
```

But it's returning:

```
array(11) { ["cedula"]=> string(1) "1" ["tipo_documento_id"]=> string(1) "3" ["nombres"]=> string(15) "HUANG ZHENGQUIN" ["telefono"]=> string(9) "191821112" ["direccion"]=> string(16) "CRA 25 CALLE 100" ["email"]=> string(9) "@yahoo.es" ["fecha_nacimiento"]=> string(10) "2011-02-03" ["ciudad_id"]=> string(6) "127591" ["creado_at"]=> string(10) "2011-05-18" ["cupo"]=> string(7) "6930.00" ["estado"]=> string(1) "I" } NULL
```

```php
$di = new Phalcon\DI\FactoryDefault();


$di->setShared(
    'db',
    function () {
        $connection = new Phalcon\Db\Adapter\Pdo\Mysql(
            [
                'host' => 'localhost',
                'port' => 3306,
                'dbname' => 'phalcon_test',
                'username' => 'root',
                'charset' => 'UTF8',
            ]
        );

        return $connection;
    }
);
$di->set(
    'cache',
[...]

PR #12171, opened 2016-08-26
Fix #12000, #12170

### LINKED ISSUE #12000: Model snapshots doesn't work with cache
```php
public function initialize()
{
    $this->keepSnapshots(true);
}
```
works fine but when using it with the caching technique from this page: https://docs.phalconphp.com/en/latest/reference/models-cache.html#caching-resultsets
it produces an error  when calling the method hasChanged.
I guess the cache is not taking the snapshot when it restores the data.
The user has to manually call `$model->setSnapshotData($model->toArray())`, but this is a workaround that should not be required.

Jurigag:
Refs: #12000, #12170


sergeyklay:
Thanks

PR #12188, opened 2016-09-01
if no header defined should return null

WhatABigBottleOfPureWater:
isset($req->getHeaders()[ 'abc' ]) To null !== $req->getHeader('abc')


sergeyklay:

PR #12189, opened 2016-09-01
if no header defined should return null

WhatABigBottleOfPureWater:
isset($req->getHeaders()[ 'abc' ]) To null !== $req->getHeader('abc')

### DIFF
--- a/phalcon/http/request.zep
+++ b/phalcon/http/request.zep
@@ -260,7 +260,7 @@ class Request implements RequestInterface, InjectionAwareInterface
 	/**
 	 * Gets HTTP header from request data
 	 */
-	public final function getHeader(string! header) -> string
+	public final function getHeader(string! header) -> string | null
 	{
 		var value, name;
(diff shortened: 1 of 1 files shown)


tkaven:
@racklin what's wrong with AppVeyor? ..

sergeyklay:
How about:
```php
if (!$req->getHeader('abc')) {
    // ...
}
```

Right now it works so. 
Also I think that introducing `hasHeader` is better than changing the current API.

tkaven:
@sergeyklay  yes! introducing hasHeader is better than changing the current API :)  :+1:

sergeyklay:
@tkaven Could you please rebase onto 3.1.x branch

sergeyklay:
@tkaven ping

niden:
Closing in favor of https://github.com/phalcon/cphalcon/pull/13570

@tkaven Thank you very much for the contribution

PR #12206, opened 2016-09-06
Fixed PHP 7 Catchable fatal error: session_regenerate_id(): Failed to create(read) session ID: user (path: ) 

yitype:
- read not returning always a string
- destroy wrongly returning false when session doesn't exist. 

**read**

If read returns anything else than a string it will fail with PHP Catchable fatal error: session_regenerate_id(): Failed to create(read) session ID: user. PHP 7 requires that reading always returns a string, regardless the existence of data.

**destroy**

When session_regenerate_id() is called right after session_start() php tries to destroy a session that doesn't exist yet. destroy() to wrongly return false when the session doesn't exist yet. PHP 7 will then fail with Warning: session_regenerate_id(): Session object destruction failed. ID: user (path: ) .

``` php
// with PHP 7.0.10 and Phalcon 3.0.1
use Phalcon\Session\Adapter\Redis as SessionAdapterRedis;

$session = new SessionAdapterRedis([
    'host'     => '127.0.0.1',
    'port'     => 6379,
]);

$session->start();

$oldId = $session->getId();

$session->regenerateId();

var_dump($oldId);
var_dump($session->getId());
```


sergeyklay:
@zhao5908 Could you please update CHANGELOG.md?

zhao5908:
Thanks @sergeyklay

sergeyklay:
Fixed in the `3.0.x` branch.

thilinah:
I'm was having this issue with 3.0.x and fixed it via extending redis.zep in php. Please check https://github.com/phalcon/cphalcon/pull/12331

PR #12207, opened 2016-09-06
Fix Confirmation validator replace pairs ':with' value.

jaredwarren:
The wrong variable was being used as the key for looking up the 'labelWith' value. 
Also, there was a missing single-quote in the class example.


sergeyklay:
@jaredwarren Could you please update CHANGELOG.md

Jurigag:
Thanks, when changing all validators i had to miss it.

sergeyklay:
Thanks

PR #12216, opened 2016-09-12
Update `_connect` function in redis.zep

stramunin:
Use `select` internal function only when the index is greater than zero. Because, new connections always use DB 0 and the index is zero-based. It is important when use twemproxy, some people have a troubles with it. twemproxy (https://github.com/twitter/twemproxy) doesn't implement select function. I found simple solution. What do you say, comrades?


sergeyklay:
@metaliciouz Could you please update `CHANGELOG.md`?

metaliciouz:
Updated.

sergeyklay:
Thanks

PR #12218, opened 2016-09-13
Update cookie.zep

AWS-alchemy:
If set to '/', the cookie will be available within the entire domain, but if i need, that cookie will only be available current directory?

see official API Description 

``` php
bool setcookie ( string $name [, string $value = "" [, int $expire = 0 [, string $path = "" [, string $domain = "" [, bool $secure = false [, bool $httponly = false ]]]]]] )
```


sergeyklay:
@ch3k1 It would be fine if you able to create a new issue with minimal script to reproduce and refer this PR to the issue.

Issue #12099, opened 2016-08-05
Phalcon\Tag::resetInput() is not work


Studentsov:
Code:
```php
$_POST = ['a' => '1', 'b' => '2'];
\Phalcon\Tag::resetInput();
var_dump($_POST);
```
output:
`array(2) { ["a"]=> string(1) "1" ["b"]=> string(1) "2" } `

Phalcon 3.0.0
OS Windows 10, PHP 7.0.8 x64
OS CentOS 7, PHP 7.0.9 x64

This problem breaks clean forms.
I found an old message: http://stackoverflow.com/questions/37365130/how-to-clear-phalcon-form-element-values, so I think the problem is not new.

sergeyklay:
@Studentsov 
Why did you decide that the `Phalcon\Tag::resetInput` should reset `$_POST`? Could you please explain a bit more?

As correctly noted @SidRoberts in #12227 the `Phalcon\Tag::resetInput` shouldn't reset `$_POST`. It should only reset parameters within the `Phalcon\Tag` object. 

I agree that we have variances in the documentation about this method, and it was used in an unsuccessful attempt to clear `$_POST`. But I'm really not sure that we need a method to do `$_POST = [];`

To clear the form you should use `Phalcon\Forms\Form::clear`. It clears every element in the form to its default value. I'll check `Phalcon\Forms\Form::clear` ASAP, but anyway this is a **different issue**, not related with cleaning `$_POST` through  `Phalcon\Tag::resetInput`.

Studentsov:
Implementation of `Phalcon\Forms\Form::clear` depends on the state of the `Phalcon\Tag`

Jurigag:
Where - https://github.com/phalcon/cphalcon/blob/master/phalcon/forms/form.zep#L663

Studentsov:
https://github.com/phalcon/cphalcon/blob/master/phalcon/forms/element.zep#L507

Jurigag:
So what ? Still it shouldn't reset `$_POST`.

sergeyklay:
@Studentsov But `Phalcon\Forms\Element::clear` not doing anything with `$_POST`

Studentsov:
https://github.com/phalcon/cphalcon/blob/master/phalcon/tag.zep#L258
https://github.com/phalcon/cphalcon/blob/master/phalcon/tag.zep#L321
just try to clean form

Jurigag:
[...]

PR #12227, opened 2016-09-16
Fixed Phalcon\Tag::resetInput for proper use without attempts to clear $_POST

sergeyklay:
Refs: #12099


SidRoberts:
I don't think resetInput() should overwrite $_POST - it should only reset parameters within the Tag object. 

My $0.02 :)

sergeyklay:
😐

sergeyklay:
@SidRoberts Good catch!

Issue #12233, opened 2016-09-19
Is that a new behavior to Phalcon\Tag::prependTitle or it is a bug?


borisdelev:
> **In Phalcon v2.x when i use tag:prependTitle the title is prepended**

```php
Phalcon\Tag::setTitle("PAGE");
Phalcon\Tag::prependTitle("test1");
Phalcon\Tag::prependTitle("test2");

// Title now is "test2 - test1 - PAGE" (depends of separator)
```
> **In Phalcon v3.x when i use tag:prependTitle the title replaced**

```php
Phalcon\Tag::setTitle("PAGE");
Phalcon\Tag::prependTitle("test1");
Phalcon\Tag::prependTitle("test2");

// Title now is "test2 - PAGE" (depends of separator)
```
### So which behavior is right?
Sorry for my English.
Thanks!

### Details

* Phalcon version: 3.x
* PHP Version: 7
* Operating System: Windows 10
* Server: Apache

butamuh4o:
Yep, sure. :) thanks

butamuh4o:
Finally. I did it :) Good luck all, and thanks for your patience.

sergeyklay:
Fixed in the `3.0.x` branch.

PR #12236, opened 2016-09-19
Fix appendTitle, prependTitle. Update unit test for append, prepend and setTitle.

borisdelev:
Update tests for prependTitle and appendTitle functions.
For details:  #12233 

Update test for setTitle and getTitle: Make sure that getTitle(false) did not return <title> tag element

### DIFF
--- a/phalcon/tag.zep
+++ b/phalcon/tag.zep
@@ -1061,46 +1061,24 @@ class Tag
 	/**
 	 * Appends a text to current document title
 	 */
-	public static function appendTitle(var title) -> void
+	public static function appendTitle(string title) -> void
 	{
-		var append = [];
 		if self::_documentAppendTitle !== null {
-			let append = self::_documentAppendTitle ;
-		}
-		
-		if typeof title == array {
-			let append = title;
+			let self::_documentAppendTitle = self::_documentAppendTitle . self::_documentTitleSeparator . title ;
 		} else {
-			if typeof title == string {
-				let append[] = title;
-			}
-		}
-		
-		if !empty append {
-			let self::_documentAppendTitle = append ;
+			let self::_documentAppendTitle = title ;
 		}
 	}
 
 	/**
 	 * Prepends a text to current document title
 	 */
-	public static function prependTitle(var title) -> void
+	public static function prependTitle(string title) -> void
 	{
-		var prepend = [];
 		if self::_documentPrependTitle !== null {
-			let prepend = self::_documentPrependTitle ;
-		}
-		
-		if typeof title == array {
-			let prepend = title;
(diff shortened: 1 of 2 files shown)


butamuh4o:
Hope i make it right now. Sorry for my mistakes :/

sergeyklay (approved):

virgofx:
Unit tests look fine ... can we get can update though for the actual Tag implementation so this will pass?

butamuh4o:
@virgofx i will try to make this changes too...

sergeyklay:
@butamuh4o As you can see tests fails

butamuh4o:
@sergeyklay I want to ask you to help me with something about a change i was trying to do. Can u check that and give me some details where and what is wrong with syntax? Thank you!!!

```
	/**
	 * Prepends a text to current document title
	 */
	public static function prependTitle(var title) -> void
	{
		var prepend = [];
		if self::_documentPrependTitle !== null {
			let prepend = self::_documentPrependTitle ;
		}
		
		if typeof title == array {
			let prepend = title;
		} else {
			if typeof title == string {
				array_unshift(prepend, title) ;
			}
		}
		
		if !empty prepend {
			let self::_documentPrependTitle = prepend ;
		}
	}
```

phalcon/tag.zep
```diff
@@ -1061,17 +1061,51 @@ class Tag
...
-	public static function prependTitle(string title) -> void
+	public static function prependTitle(var title) -> void
 	{
-		let self::_documentPrependTitle = title;
+		var prepend = [];
+		if self::_documentPrependTitle !== null {
+			let prepend = self::_documentPrependTitle ;
+		}
+		
+		if typeof title == "array" {
```
sergeyklay: This is new feature request?

phalcon/tag.zep
```diff
@@ -1061,17 +1061,51 @@ class Tag
...
+			if typeof title == "string" {
+				array_unshift(prepend, title) ;
+			}
+		}
+		
+		if empty prepend {
+			let prepend = null ;
+		}
+		
+		let self::_documentPrependTitle = prepend ;
```
sergeyklay: So now `_documentPrependTitle` is array or null and this breaking BC

sergeyklay (changes requested):

butamuh4o:
No it is not change request. Just think it is better to have that option. Will revert that idea. Sorry

sergeyklay:
It is good idea, but for 3.1.x not for 3.0.x

sergeyklay:
@butamuh4o Also could you update CHANGELOG.md?

butamuh4o:
So i will make a clear pull request and update changelog. Just to get to the office. Thank you!

PR #12237, opened 2016-09-20
Fix appendTitle, prependTitle. Update unit test for append, prepend and setTitle

borisdelev:
Update prependTitle and appendTitle to work correctly. 

Update tests for prependTitle and appendTitle functions.
For details: #12233

Update test for setTitle and getTitle: Make sure that getTitle(false) did not return tag element.


butamuh4o:
It is same as https://github.com/phalcon/cphalcon/pull/12236. Closed by phone mistake.

sergeyklay (approved):

sergeyklay:
Thanks

PR #12240, opened 2016-09-21
Added "DISTINCT" type for Text::random()

tmihalik:


sergeyklay:
@tmihalik Could you please provide a small test which covers new feature. Also please update change log

PR #12243, opened 2016-09-21
Added two new filters for use in the Filter->sanitization() method.  …

jabarihunt:
Added two new filters for use in the Filter->_sanitization()_ method...

The first is FILTER_BOOLEAN, which returns the boolean value (true or false) if the value is an actual boolean (not 0, 1, 'true' as a string, etc) or it returns null if the value is not a true boolean.

The second is FILTER_BOOLEAN_CAST, which returns whatever the boolean would be of the casted value.  For example, 1 would return true and and empty string would return false.

**_NOTE:_** FILTER_BOOLEAN is a special case that required a break in the sanitization switch statement since it may return null.  This was accounted for in related loops.

### DIFF
--- a/phalcon/filter.zep
+++ b/phalcon/filter.zep
@@ -35,33 +35,38 @@ use Phalcon\Filter\Exception;
  *	$filter->sanitize("hello<<", "string"); // returns "hello"
  *	$filter->sanitize("!100a019", "int"); // returns "100019"
  *	$filter->sanitize("!100a019.01a", "float"); // returns "100019.01"
+ *	$filter->sanitize(1, "boolean"); // returns TRUE
  *</code>
  */
 class Filter implements FilterInterface
 {
-	const FILTER_EMAIL      = "email";
+	const FILTER_EMAIL        = "email";
 
-	const FILTER_ABSINT     = "absint";
+	const FILTER_ABSINT       = "absint";
 
-	const FILTER_INT        = "int";
+	const FILTER_INT          = "int";
 
-	const FILTER_INT_CAST   = "int!";
+	const FILTER_INT_CAST     = "int!";
 
-	const FILTER_STRING     = "string";
+	const FILTER_STRING       = "string";
 
-	const FILTER_FLOAT      = "float";
+	const FILTER_FLOAT        = "float";
 
-	const FILTER_FLOAT_CAST = "float!";
+	const FILTER_FLOAT_CAST   = "float!";
 
-	const FILTER_ALPHANUM   = "alphanum";
+	const FILTER_ALPHANUM     = "alphanum";
 
-	const FILTER_TRIM       = "trim";
+	const FILTER_TRIM         = "trim";
 
-	const FILTER_STRIPTAGS  = "striptags";
+	const FILTER_STRIPTAGS    = "striptags";
 
-	const FILTER_LOWER      = "lower";
+	const FILTER_LOWER        = "lower";
 
-	const FILTER_UPPER      = "upper";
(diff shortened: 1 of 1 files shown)


phalcon/filter.zep
```diff
@@ -209,6 +218,17 @@ class Filter implements FilterInterface
 				}
 				return strtoupper(value);
 
+			case Filter::FILTER_BOOLEAN:
+				/**
+				 * Must break, as it may return a null value
+				 */
+				return is_bool($value) ? $value : null;
```
sergeyklay: `$value` or `value` ?

phalcon/filter.zep
```diff
@@ -103,6 +108,10 @@ class Filter implements FilterInterface
 					} else {
 						let value = this->_sanitize(value, filter);
 					}
+					/**
+					 * Account for any filters that may return a null value
+					 */
+					if (value === null) {break;}
```
sergeyklay: parentheses not needed here

sergeyklay (changes requested):

phalcon/filter.zep
jabarihunt: Not sure what you are asking here.  Would you rather I use an if statement instead of the ternary operator?

phalcon/filter.zep
jabarihunt: I changed the syntax in the last commit.

phalcon/filter.zep
sergeyklay: I meant that you should use `value` without `$`

phalcon/filter.zep
```diff
@@ -209,6 +220,17 @@ class Filter implements FilterInterface
...
+				/**
+				 * Must break, as it may return a null value
+				 */
+				return is_bool($value) ? $value : null;
+				break;
+				
+			case Filter::FILTER_BOOLEAN_CAST:
+
+				return boolval($value);
+
```
sergeyklay:
```
return boolval(value);
```

andrew-demb:
These changes it's new features. By semver it must be in new minor version, not 3.0.x. Maybe need to create branch 3.1.x and move PR to new branch?

Jurigag:
Well most of PRs need to be redone on 3.1.x.

Jurigag (commented):
This is new feature and need 3.1.0 label.

phalcon/filter.zep
jabarihunt: Corrected!  Sorry, I was stuck in PHP mode.

sergeyklay:
@jabarihunt Could you please rebase onto 3.0.x and update CHANGELOG.md. Also squash commits so it would contain only 1 commit.

phalcon/filter.zep
jabarihunt: Corrected.

phalcon/filter.zep
```diff
@@ -111,9 +111,13 @@ class Filter implements FilterInterface
 					/**
 					 * Account for any filters that may return a null value
 					 */
+<<<<<<< HEAD
```
sergeyklay: Could you please resolve conflicts?

sergeyklay:
@jabarihunt 
Thank you for contributing.
Could you please cover new feature by small unit test?

Jurigag:
Also squash please.

sergeyklay:
@jabarihunt Could you please rebase onto 3.1.x ?

sergeyklay:
@jabarihunt Could you please rebase onto 3.2.x ?

PR #12256, opened 2016-09-26
Fix #12254

### LINKED ISSUE #12254: $validation->setFilters("name", "trim"); seems like not working
here is my code:

```php
use Phalcon\Validation;
use Phalcon\Mvc\Model;
use Phalcon\Validation\Validator\PresenceOf;

class Category extends Model
{
    public function validation()
    {
        $validation = new Validation();
        $validation->add(['title', 'name'], new PresenceOf([
            'message' => [
                'title' => 'title is required',
                'name' => 'name is required'
            ]
        ]));

        $validation->setFilters("name", "trim");
        $validation->setFilters("title", "trim");

        return $this->validate($validation);
    }
```

my input is some blank sapce like title: '    ',  name: '  ',
then i set filters by using 
> $validation->setFilters("name", "trim");

but it sames not working. the validation still return true;
and if i use $this->title = trim($this->title);  before validation, the validation will return false.

Jurigag:


phalcon/validation.zep
```diff
@@ -494,30 +495,30 @@ class Validation extends Injectable implements ValidationInterface
...
-					let value = data->{field};
+
+			let value = null;
+			if typeof data == "array" {
+				if isset data[field] {
+					let value = data[field];
+				}
+			} else {
+				if typeof data == "object" {
+					if isset data->{field} {
```
sergeyklay:
```php
$data = new stdClass;
$field = "foo";
$data->{$field} = null;
var_dump(isset($data->foo)); // false
var_dump(property_exists($data, "foo")); // true
```

sergeyklay (changes requested):
Also could you please rebase. Fixed Appveyor build: https://github.com/phalcon/cphalcon/pull/12257

sergeyklay (approved):

Issue #12165, opened 2016-08-25
Element clear not working


Semihal:
Does not work in the form of a $elemnt->clear() while transfer in the form of any entity.

```php
        $confirmPassword = new Password('confirmPassword', [
            'placeholder'   => '***********',
        ]);

        $confirmPassword->addValidator(
            new Confirmation([
                'with'  => 'password',
                'message'   => 'Пароли должны совподать'
            ])
        );

        $confirmPassword->clear();
```
The field is not cleared. If only to refresh the page.

### Details
* Phalcon version: 3.0.0
* PHP Version: 7.0.8
* Operating System: Win10
* Installation type: installing dll
* Zephir version (if any): ---
* Server: Apache
* Other related info (Database, table schema): Mysql5.7

sergeyklay:
@Semihal Could you please take a look at #12280? In general this is what you need. Right?

sergeyklay:
Fixed in `3.0.x` branch.
Could you please check `3.0.x` branch?

```sh
git clone git@github.com:phalcon/cphalcon.git
cd cphalcon
git checkout 3.0.x

zephir fullclean
zephir build
```

Semihal:
Yes, thank you

Issue #11978, opened 2016-07-15
Form->clear() doesn't reset element's value when using Form->isValid()


cirot87:
Hi,  i have a login form with email and password elements. I need to clear the password value when validation fails.
In my /Phalcon/Forms/Form i have this:
```php
$email = new Text('email', [
    'placeholder' => DI::getDefault()->get("translation")->query("Insert your Email")
]);
$email->addValidators( [
    new PresenceOf( [
        'message'       => DI::getDefault()->get("translation")->query("The field is required"),
        'cancelOnFail'  => true
    ]),
    new Email( [
        'message' => DI::getDefault()->get("translation")->query("Invalid email address")
    ])
]);

$password = new Password('password', [
    'placeholder' => DI::getDefault()->get("translation")->query("Insert your Password")
]);
$password->addValidators([
    new PresenceOf( [
        'message' => DI::getDefault()->get("translation")->query("The field is required"),
        'cancelOnFail' => true
    ]),
    new StringLength([
        'min'            => 7,
        'messageMinimum' => DI::getDefault()->get("translation")->query("The text is too short")
    ])
]);
$password->setDefault("");
$password->clear();

$this->add($email);
$this->add($password); 
```

in my view i have a simple render for both, and after the post i check the value with the isValid() function.
The problem is that when i have a validation error the password won't reset.

I try this for debugging:
```php
Tag::setDefault("password", "cir");
$form->get('password')->clear();
$form->get('password')->setDefault('ci');
$form->clear(['password']);

echo "<br>Tag -------> ".Tag::getValue('password');
echo "<br>Form ------> ".$form->getValue('password');
echo "<br>Element ---> ".$form->get('password')->getValue();
```
        
but none of those reset my value on the form.

The problem is that when using isValid() which calls the bind() function,.
The bind() set automatically all post parameters into _data (a protected variable of the form) 
`let this->_data = data;`
and  this clear function never change the parameters inserted into this variable.
```zephir
public function clear(var fields = null) -> <Form>
{
    var elements, element;

    let elements = this->_elements;
    if typeof elements == "array" {
        for element in elements {
            if typeof fields != "array" {
                element->clear();
            } else {
                if in_array(element->getName(), fields) {
                    element->clear();
                }
            }
        }
    }
    return this;
}
```

My solution
```php
public function clear($fields = null)
{
    if (is_null($fields)) {
        $this->_data = [];
    } else {
        if (is_array($fields)) {
            foreach ($fields as $field) {
[...]

Issue #10044, opened 2015-04-15
[2.0.0] Form element clear do not works as in 1.3.4


DestinyMKas:
Hello,

1.3.4 version had form with password field. I don't want to render password, so I used $element->clear(). In 2.0.0 it not works.
I tried to do $element->setDefault('')->clear(), but it didn't worked too.

As a workaround I did:
$this->getEntity()->password = '';
But this method has some issues.

DestinyMKas:
#10069 pull request initiated

DestinyMKas:
Wrong comment

sergeyklay:
Fixed in `3.0.x` branch.
Could you please check `3.0.x` branch?

```sh
git clone git@github.com:phalcon/cphalcon.git
cd cphalcon
git checkout 3.0.x

zephir fullclean
zephir build
```

PR #12280, opened 2016-10-01
Test clearing form

sergeyklay:
Refs:
- #12099
- #12165
- #11978
- #10044
- http://stackoverflow.com/questions/37365130/how-to-clear-phalcon-form-element-values


Studentsov:
Please, check `$form->render('field')`

sergeyklay:
@Studentsov Could you please take a look?

scrnjakovic:
Phalcon 3.4.0

If you set the form itself as it's entity, calling `clear()` method won't have any effect as the form will still be populated with $_POST values. Is this intended behavior or bug?

```php
$form->setEntity($form);

if ($form->isValid($_POST)) {
    echo $form->getValue('password'); // secret
    echo $form->get('password')->getValue(); // secret
    $form->clear();
    echo $form->getValue('password'); // secret
    echo $form->get('password')->getValue(); // secret
}
```

PR #12285, opened 2016-10-02
Fix mysql and postgresql missing schema

Jurigag:


phalcon/db/dialect/postgresql.zep
```diff
@@ -611,7 +611,7 @@ class Postgresql extends Dialect
 		if schema {
 			let sql .= "tc.table_schema = '" . schema . "' AND tc.table_name='" . table . "'";
 		} else {
-			let sql .= "tc.table_name='" . table . "'";
+			let sql .= "tc.table_schema = = 'public' AND tc.table_name='" . table . "'";
```
Green-Cat: Should be 
> tc.table_schema = 'public'

Jurigag: Yea a typo

sergeyklay:
Thanks

PR #12287, opened 2016-10-03
Form Elements no longer implement ElementInterface twice.

SidRoberts:
[Phalcon\Forms\Element already implements ElementInterface](https://github.com/phalcon/cphalcon/blob/3.0.x/phalcon/forms/element.zep#L34)


sergeyklay:
Thanks

PR #12292, opened 2016-10-05
3.1.x check if before error,before throwable works in php5

Jurigag:


virgofx:
The dispatcher changes will not be compatible with this. Don't merge.

Jurigag:
Compiling fine, tests for php 5.5 and 5.6 passing, but better keep talking, yea.

virgofx:
Introducing throwables will not be done on a small use case like this. It requires change globally as I have said before, including documentation changes as well

Jurigag:
I will include documentation changes as well, what globally ? When 3.1.x repo will be here i will create final PR. But now there must be multicatching in zephir fixed because right now it doesn't work for some reason.

virgofx:
There needs to be a strategy involved in handling the transition to PHP7 with respect to exceptions globally in the code base, documentation, backwards compatibility, and other concerns.  It will involve more discussion in future.

So for now, this is a bigger project than a simple hook for dispatcher.

Jurigag:
What you mean globally in the code ? In php7 all exceptions implements throwable, nothing fancy here. Just write code which will be breaking this stuff after i will do PR to 3.1.x and then we can start discussion.

PR #12317, opened 2016-10-14
[RFC] Converting Active Record to the Data Mapper pattern

SidRoberts:
So I thought I'd have a go at converting Phalcon's Active Record models to the Data Mapper pattern (see [Doctrine](http://www.doctrine-project.org/projects/orm.html)) and I want to hear people's thoughts and opinions before I take it any further.

I've avoided doing to much work just in case it ends up being a dead-end so no code has been refactored and methods in the models manager now call protected methods in Phalcon\Mvc\Model, etc; - it's a bit of a mess. Obviously this will/should all change.

Put succinctly, methods have been moved to the models manager and (currently) operate exactly as they did before:

``` php
// Before
$person = People::findFirst($params);

// After
$person = $modelsManager->findFirst(
    People::class,
    $params
);
```

``` php
// Before
$person = People::count($params);

// After
$person = $modelsManager->count(
    People::class,
    $params
);
```

``` php
// Before
$person->delete();

// After
$modelsManager->delete($person);
```

``` php
// Before
$person->save($params);

// After
$modelsManager->save($person, $params);
```

Phalcon\Mvc\Model is still far too complicated, in my opinion, so there's still a lot of work to do. :stuck_out_tongue: 

A few things that I'd like to do is remove the ability to assign properties during create/update/save:

``` php
$robot->save(
    [
        "name" => "Bender",
    ]
);

$modelsManager->save(
    $robot,
    [
        "name" => "Bender",
    ]
);
```

I know people like it because it acts a little shortcut but it doesn't make sense in the data mapper context and looks a little clumsy. A method called `save()` should save and do nothing else.

I'd also like to decide between only supporting [public properties or getters/setters](https://docs.phalconphp.com/en/latest/reference/models.html#public-properties-vs-setters-getters) but have no personal preference of either. Alternatively (or additionally), make an effort to vastly simplify how properties are get/set internally.

(It goes without saying but is obviously aimed at version 4 - not version 3)

Thoughts?

### DIFF
--- a/phalcon/mvc/model/manager.zep
+++ b/phalcon/mvc/model/manager.zep
@@ -231,7 +231,7 @@ class Manager implements ManagerInterface, InjectionAwareInterface, EventsAwareI
 		/**
 		 * Update the model as initialized, this avoid cyclic initializations
 		 */
-		let this->_initialized[className] = model;
+		let this->_initialized[className] = true;
 
 		/**
 		 * Call the 'initialize' method if it's implemented
@@ -275,7 +275,7 @@ class Manager implements ManagerInterface, InjectionAwareInterface, EventsAwareI
 	/**
 	 * Loads a model throwing an exception if it doesn't exist
 	 */
-	public function load(string! modelName, boolean newInstance = false) -> <ModelInterface>
+	public function load(string! modelName) -> <ModelInterface>
 	{
 		var model, colonPos, namespaceName, namespaceAlias, className;
 
@@ -299,19 +299,11 @@ class Manager implements ManagerInterface, InjectionAwareInterface, EventsAwareI
 		}
 
 		/**
-		 * Check if a model with the same is already loaded
-		 */
-		if !newInstance {
-			if fetch model, this->_initialized[strtolower(modelName)] {
-				model->reset();
-				return model;
-			}
-		}
-
-		/**
 		 * Load it using an autoloader
 		 */
-		return new {modelName}(null, this->_dependencyInjector, this);
+		let model = new {modelName}(null, this->_dependencyInjector, this);
+
+		return model;
 	}
 
 	/**
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -2,3 +2,4 @@
 - MVC Application and Router now must have a URI to process [#12380](https://github.com/phalcon/cphalcon/pull/12380)
 - Response headers and cookies are no longer prematurely sent [#12378](https://github.com/phalcon/cphalcon/pull/12378)
 - You can no longer assign data to models whilst saving them [#12317](https://github.com/phalcon/cphalcon/issues/12317)
+- `ModelManager::load()` no longer reuses already initialized models [#12317](https://github.com/phalcon/cphalcon/issues/12317)
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -3,3 +3,4 @@
 - Response headers and cookies are no longer prematurely sent [#12378](https://github.com/phalcon/cphalcon/pull/12378)
 - You can no longer assign data to models whilst saving them [#12317](https://github.com/phalcon/cphalcon/issues/12317)
 - `ModelManager::load()` no longer reuses already initialized models [#12317](https://github.com/phalcon/cphalcon/issues/12317)
+- Removed `Model::reset()` [#12317](https://github.com/phalcon/cphalcon/issues/12317)
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -1,3 +1,4 @@
 # [4.0.0](https://github.com/phalcon/cphalcon/releases/tag/v4.0.0) (2017-XX-XX)
 - MVC Application and Router now must have a URI to process [#12380](https://github.com/phalcon/cphalcon/pull/12380)
 - Response headers and cookies are no longer prematurely sent [#12378](https://github.com/phalcon/cphalcon/pull/12378)
+- You can no longer assign data to models whilst saving them [#12317](https://github.com/phalcon/cphalcon/issues/12317)
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -4769,13 +4769,4 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 		    globals_set("orm.disable_assign_setters", disableAssignSetters);
 		}
 	}
-
-	/**
-	 * Reset a model instance data
-	 */
-	public function reset()
-	{
-		let this->_uniqueParams = null;
-		let this->_snapshot = null;
-	}
 }
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -2991,20 +2991,14 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 	 * $robot->save();
 	 *</code>
 	 *
-	 * @param array data
-	 * @param array whiteList
 	 * @return boolean
 	 */
-	public function save(var data = null, var whiteList = null) -> boolean
+	public function save() -> boolean
 	{
 		var metaData, related, schema, writeConnection, readConnection,
 			source, table, identityField, exists, success;
 
 		let metaData = this->getModelsMetaData();
-
-		if typeof data == "array" && count(data) > 0 {
-			this->assign(data, null, whiteList);
-		}
 
 		/**
 		 * Create/Get the current database connection
@@ -3150,16 +3144,18 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 	 * // Passing an array to create
 	 * $robot = new Robots();
 	 *
-	 * $robot->create(
+	 * $robot->assign(
 	 *     [
 	 *         "type" => "mechanical",
 	 *         "name" => "Astro Boy",
 	 *         "year" => 1952,
 	 *     ]
 	 * );
+	 *
+	 * $robot->create();
 	 *</code>
 	 */
-	public function create(var data = null, var whiteList = null) -> boolean
+	public function create() -> boolean
 	{
 		var metaData;
 
@@ -3179,7 +3175,7 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 		/**
 		 * Using save() anyways
 		 */
-		return this->save(data, whiteList);
+		return this->save();
 	}
 
 	/**
@@ -3195,7 +3191,7 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 	 * $robot->update();
 	 *</code>
 	 */
-	public function update(var data = null, var whiteList = null) -> boolean
+	public function update() -> boolean
 	{
 		var metaData;
 
@@ -3222,7 +3218,7 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 		/**
 		 * Call save() anyways
 		 */
-		return this->save(data, whiteList);
+		return this->save();
 	}
 
 	/**
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -2903,7 +2903,7 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 						/**
 						 * Create a new instance of the intermediate model
 						 */
-						let intermediateModel = manager->load(intermediateModelName, true);
+						let intermediateModel = manager->load(intermediateModelName);
 
 						/**
 						 * Write value in the intermediate model
--- a/phalcon/mvc/model/managerinterface.zep
+++ b/phalcon/mvc/model/managerinterface.zep
@@ -103,7 +103,7 @@ interface ManagerInterface
 	/**
 	 * Loads a model throwing an exception if it doesn't exist
 	 */
-	public function load(string modelName, boolean newInstance = false) -> <ModelInterface>;
+	public function load(string modelName) -> <ModelInterface>;
 
 	/**
 	 * Setup a 1-1 relation between two models
--- a/phalcon/mvc/model/query.zep
+++ b/phalcon/mvc/model/query.zep
@@ -2982,11 +2982,13 @@ class Query implements QueryInterface, InjectionAwareInterface
 		 */
 		let insertModel = clone manager->load(modelName);
 
+		insertModel->assign(insertValues);
+
 		/**
 		 * Call 'create' to ensure that an insert is performed
 		 * Return the insert status
 		 */
-		return new Status(insertModel->create(insertValues), insertModel);
+		return new Status(insertModel->create(), insertModel);
 	}
 
 	/**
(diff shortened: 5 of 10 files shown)


Jurigag:
Intorduce pls repository class imho as well, because using modelsManager is dirty. So we can still jus t do `$personRepostory->find();` for example.

Also this change is of course for 4.0.0. I thought @andresgutierrez was planning this for 4.0.0

sergeyklay (approved):
Looks great to me

phalcon/mvc/model/manager.zep
```diff
@@ -1805,4 +1811,888 @@ class Manager implements ManagerInterface, InjectionAwareInterface, EventsAwareI
...
 		Query::clean();
 	}
+
+
+
+
+
+
+
+
```
Green-Cat: Too much spacing.

phalcon/mvc/model/manager.zep
```diff
@@ -1805,4 +1811,888 @@ class Manager implements ManagerInterface, InjectionAwareInterface, EventsAwareI
...
+				resultset->setHydrateMode(hydration);
+			}
+		}
+
+		return resultset;
+	}
+
+
+	/**
+	 * Allows to query the first record that match the specified conditions
```
Green-Cat: Allows to query for the first record that matches the specified conditions

phalcon/mvc/model/manager.zep
SidRoberts: I was going for speed. :wink:

Jurigag: No need for speed, it will be at least few months until we create 4.0.x repo i guess, and few years to release :D

SidRoberts:
@Jurigag, it'd be tedious to have to create repository classes for each model. We could set the model class name in the constructor though:

```zephir
namespace Phalcon\Mvc\Model;

class Repository
{
	public function __construct(string! modelClass);

	public function find(array params = []) -> <ResultsetInterface>;

	public function findFirst(array params = []) -> <ModelInterface>;

	public function count(array parameters = []) -> int;

	public function sum(array parameters = []);

	public function maximum(array parameters = []);

	public function minimum(array parameters = []);

	public function average(array parameters = []);
}
```

It would reduce the complexity of the Models Manager a bit but would repositories add complexity for  users/developers?:

```php
$robots = $modelsManager->findFirst(
    Robots::class,
    $params
);

// OR ...

$robotsRepository = $modelsManager->getRepository(
    Robots::class
);

$robots = $robotsRepository->findFirst($params);
```

phalcon/mvc/model/manager.zep
SidRoberts: Fixed in 77acbfe.

Jurigag:
I mean if you don't want - you don't need to create repository. You can use as well just `$modelsManager->findFirst(Robots::class, $params)` i guess maybe. Repository will be just some kind of shorthand and some place where you can add addtional methods for finding stuff. I know developer can always create his own method but in most Data Mappers there is Repository class introduced.

Jurigag (changes requested):
`Phalcon/Mvc/Model/Repository` class, add option to use custom Repository class for models.

phalcon/mvc/model/manager.zep
```diff
@@ -1805,4 +1811,855 @@ class Manager implements ManagerInterface, InjectionAwareInterface, EventsAwareI
...
+		 * Fire the start event
+		 */
+		model->fireEvent("prepareSave");
+
+		/**
+		 * Save related records in belongsTo relationships
+		 */
+		let related = model->_related;
+		if typeof related == "array" {
+			if model->_preSaveRelatedRecords(writeConnection, related) === false {
```
Jurigag: There should be something like `Phalcon\Mvc\Model\Worker` or `Phalcon\Mvc\Model\UnitOfWork` which will do this.

phalcon/mvc/model/manager.zep
```diff
@@ -1805,4 +1811,855 @@ class Manager implements ManagerInterface, InjectionAwareInterface, EventsAwareI
...
+			}
+
+			return false;
+		}
+
+		/**
+		 * Depending if the record exists we do an update or an insert operation
+		 */
+		if exists {
+			let success = model->_doLowUpdate(metaData, writeConnection, table);
```
Jurigag: There should be something like `Phalcon\Mvc\Model\Worker` or `Phalcon\Mvc\Model\UnitOfWork` which will do this.

phalcon/mvc/model/manager.zep
```diff
@@ -1805,4 +1811,855 @@ class Manager implements ManagerInterface, InjectionAwareInterface, EventsAwareI
...
+			/**
+			 * Rollbacks the implicit transaction if the master save has failed
+			 */
+			if success === false {
+				writeConnection->rollback(false);
+			} else {
+				/**
+				 * Save the post-related records
+				 */
+				let success = model->_postSaveRelatedRecords(writeConnection, related);
```
Jurigag: There should be something like `Phalcon\Mvc\Model\Worker` or `Phalcon\Mvc\Model\UnitOfWork` which will do this.

phalcon/mvc/model/manager.zep
```diff
@@ -1805,4 +1811,855 @@ class Manager implements ManagerInterface, InjectionAwareInterface, EventsAwareI
...
+		 * _postSave() invokes after* events if the operation was successful
+		 */
+		if globals_get("orm.events") {
+			let success = model->_postSave(success, exists);
+		}
+
+		if success === false {
+			model->_cancelOperation();
+		} else {
+			model->fireEvent("afterSave");
```
Jurigag: There should be something like `Phalcon\Mvc\Model\Worker` or `Phalcon\Mvc\Model\UnitOfWork` which will do this.

phalcon/mvc/model/manager.zep
```diff
@@ -1805,4 +1811,855 @@ class Manager implements ManagerInterface, InjectionAwareInterface, EventsAwareI
...
+
+		/**
+		 * _postSave() invokes after* events if the operation was successful
+		 */
+		if globals_get("orm.events") {
+			let success = model->_postSave(success, exists);
+		}
+
+		if success === false {
+			model->_cancelOperation();
```
Jurigag: There should be something like `Phalcon\Mvc\Model\Worker` or `Phalcon\Mvc\Model\UnitOfWork` which will do this.

Jurigag (commented):
Worker class for model methods:
- _preSaveRelatedRecords
- _exists
- _preSave
- _doLowUpdate
- _doLowInsert
- _postSaveRelatedRecords
- _postSave
- _cancelOperation
- fireEvent
- _checkForeignKeysReverseRestrict
- _checkForeignKeysReverseCascade

So we can remove all this not needed logic in model. It should impelemnt `EvensAwareInterface` so we can attach some listener to it. Not sure about which events exactly - i guess preSave, postSave, onCancelOperation, onUpdate, onInsert ? Also as well existing model events so we can watch all model events fired for all models in one listener.

Jurigag (changes requested):
Test cases for class implementing only ModelInterface not extending Model,  in future for any class without any interface or extending class.

SidRoberts:
What uses could a custom repository have? In my head, I imagine Repository only has [the methods listed above](https://github.com/phalcon/cphalcon/pull/12317#issuecomment-253839204) (as well as support for findFirstBy..., findBy..., countBy..., etc;).

It might be a good idea to allow the developer to overwrite the repository class via the DI:

```php
$di->set("modelRepository", "Phalcon\\Mvc\\Model\\Repository");

// OR...

$di->set("Phalcon\\Mvc\\Model\\Repository", "MyCustomRepositoryClass");
```

```zephir
let repository = <RepositoryInterface> dependencyInjector->get(
    "modelRepository",
    [modelClass, this]
);

// OR...

let repository = <RepositoryInterface> dependencyInjector->get(
    "Phalcon\\Mvc\\Model\\Repository",
    [modelClass, this]
);
```

Jurigag:
Repository also should provide query builder property(`Phalcon\Mvc\Model\Query\Builder`) and query method(`Phalcon\Mvc\Model`Criteria`). With builder we can do extensive queries. So there is example:

```php
class PersonRepository extends Repository
{
    public function findWithProfiles()
    {
        return $this->queryBuilder->from('Person')
            ->columns('*')
            ->leftJoin('Profile')
            ->getQuery()
            ->execute();
    }
}
```

Jurigag:
@SidRoberts 

Also this is really great idea. But this definitely need a hard work and implement new classes. It's not only change a few we have right now = and we have data mapper. I just add my ideas what i think should be here and how it should look. How it will be implemented and what's your idea it's up to you and someone who will merge this.

SidRoberts:
It's been a long day and after spending so much time looking at code, I can't see straight anymore. I need a good night's sleep before I can do any more serious coding.  :stuck_out_tongue_closed_eyes: I agree with your ideas about custom repositories though.

How about copying how the Models Manager deals with custom model sources and schemas (`Model::setSource()` / `Model::setSchema()`) and managing custom repositories like this:

```php
class Robots extends \Phalcon\Mvc\Model
{
	public function initialize()
	{
		$this->setRepository(
			RobotsRepository::class
		);
	}
}
```

I think I'll use this PR as a playground for ideas and redo the entire thing when the 4.0.x branch is created and everyone's opinions are taken into account.

Jurigag:
Yea This is good way BUT i'm not sure about $this->setRepository. DataMapper is all about NOT EXTENDING any class i think.

I think in result and in final form the only needed thing for our models in phalcon 4.0.0 should be a) implement interface b) use trait c) don't use anything. Im just not sure about solution. For NOW your solution is fine, but i think we should avoid ` extends \Phalcon\Mvc\Model` and in final form it should aim to not having ` extends \Phalcon\Mvc\Model` and somethign like `use Model` as a trait(as far as i know traits are just classes internally ? Correct me if im wrong. Or just only some interface for needed things to work. And repository would be just `public static function getRepository()`

SidRoberts:
I'm glad you're here - you're right and I'm thinking of solutions that require the least amount of effort to implement. :wink: 

I've never used traits but I seem to remember that it's akin to copying and pasting chunks of code, as far as the compiler is concerned.

We'll figure it out tomorrow.

Jurigag:
But im not sure, maybe there should be like totally nothing, this is up to people with more knowledge, don't if DataMapper pattern allows anything like trait or something like this. Well in doctrine there are annotations, and something is parsing them, so well - this is just some kind of "workaround" in doctrine to not having trait or something like this.

SidRoberts:
That's kind of the problem. Doctrine stores a lot of metadata in annotations and I don't think annotations will be considered fast enough. We could have metadata stored somewhere else (YAML, XML, ...) but that'd be messy. Or we could just have a few getters in Model.

Jurigag:
Or just trait and still use $this->xyz but without extends.

Jurigag:
Well it's not so easy, but maybe this is a FINALLY time to introduce traits to zephir ? @andresgutierrez @sergeyklay we are waiting already 2 years for it https://github.com/phalcon/zephir/issues/504

Jurigag:
Well i will add trait support for PHP(only atm) in zephir soon so DataMapper will work as trait in php classes.

phalcon/mvc/model/repository.zep
```diff
@@ -0,0 +1,596 @@
...
+ */
+class Repository implements RepositoryInterface
+{
+
+	/**
+	 * @var string
+	 */
+	protected _modelClass;
+
+	protected _modelsManager;
```
Jurigag: Oh okay, there is already `_modelsManager`, forgot what i wrote about `createBuilder` in repository then :D

phalcon/mvc/model/manager.zep
```diff
@@ -1805,4 +1810,459 @@ class Manager implements ManagerInterface, InjectionAwareInterface, EventsAwareI
...
+			model->fireEvent("afterSave");
+		}
+
+		return success;
+	}
+
+	/**
+	 * Executes internal events after save a record
+	 */
+	protected function _postSave(<ModelInterface model, boolean success, boolean exists) -> boolean
```
Jurigag: Don't agree with having this in manager imho. As i wrote already we should introduce some new class.

Jurigag:
I don't like making this manager such a big class now, we definitely need new class what i already wrote imho.

phalcon/mvc/model/manager.zep
SidRoberts: But UnitOfWork is effectively a transaction - it only makes sense to use it if multiple models are being saved at once. I want to get all the pre/post protected methods into Manager just for now so they're all together and we can move again them later.

SidRoberts:
I've reverted the last few commits. There was a bug somewhere in there anyway. :stuck_out_tongue:

phalcon/mvc/model/manager.zep
```diff
@@ -2249,4 +2235,157 @@ class Manager implements ManagerInterface, InjectionAwareInterface, EventsAwareI
 
 		return success;
 	}
+
+	/**
+	 * Checks if a record already exists or not
+	 */
+	public function exists(<ModelInterface> model) -> boolean
```
Jurigag: Well this can be in manager maybe, but i think the rest of `_anyMethod` from model should go to new class.

allflame:
@Jurigag on a side note: why not use doctrine as is?

Jurigag:
Because doctrine is PHP and phalcon is c and is all about performance and doctrine is slowest ORM available for php ?

allflame:
If you look into the comparison you will see a strong correlation between UoW/Datamapper (slower) - ActiveRecord (faster).
I do understand the benefits of writing Zephir code for the Phalcon ORM and maybe this is a good idea.

Jurigag:
Well this is why we have zephir and we cant create something which will be still fast enough and will get all advantages of Datamapper - but it should be a real Datamapper or be really similar to it.

sergeyklay:
abandoned project https://github.com/lynx/lynx-legacy-zephir

Jurigag:
@SidRoberts also if you will return to this it could be nice to add some event like notFound or something if result of `findFirst` don't exist in database

SidRoberts:
I'm currently working on moving the relevant unit tests to Codeception and tidying them up a bit (#12348), refactoring model classes (#12328) and doing anything else to reduce the possibility of messy merge conflicts and introducing regressions.

SidRoberts:
How about introducing this slowly over the course of version 3, keeping BC along the way and then drop BC or deprecate in version 4? Or would that be too complicated?

Jurigag:
Too complicated in my opinion. Better just create 4.0.x branch and merge it while it's be done and that's it.

tempcke:
I've been doing a lot of reading lately about data mapper vs active record and the more I read the more I like the data mapper pattern.  Though I do not like doctrine due to all the dang @annotations.  Is it your goal to replace active record with data mapper pattern or will both be available ?  Will the domain models be required to extend a base class?

virgofx:
Just a note theres no way the active record can be replaced in Phalcon as this is one of the core features of the framework. Any new ORM will need to be an additional option that doesn't impact existing active record.

SidRoberts:
I completely messed up the history and there were too many merge conflicts to deal with so I'm recreating these commits manually. The original branch is available at https://github.com/SidRoberts/cphalcon/commits/v4-model-data-mapper-original

sergeyklay:
@SidRoberts Could you please check tests

sergeyklay:
Thank you

MufidJamaluddin:
I think the better option is move pagination method to this repository or make new class (PagingRepository)

PR #12331, opened 2016-10-17
Fix session_regenerate_id issue when using redis as the session handler

thilinah:
Fix session_regenerate_id issue when using redis as the session handler in php7. No session handler should return null for read method. it should return an empty string is the key is not found

### DIFF
--- a/phalcon/session/adapter/redis.zep
+++ b/phalcon/session/adapter/redis.zep
@@ -118,7 +118,8 @@ class Redis extends Adapter
 	 */
 	public function read(sessionId) -> string
 	{
-		return (string) this->_redis->get(sessionId, this->_lifetime);
+	    let res = (string) this->_redis->get(sessionId, this->_lifetime);
+        return res == null ? "" : res;
 	}
 
 	/**
--- a/phalcon/session/adapter/redis.zep
+++ b/phalcon/session/adapter/redis.zep
@@ -119,7 +119,7 @@ class Redis extends Adapter
 	public function read(sessionId) -> string
 	{
 	    let res = (string) this->_redis->get(sessionId, this->_lifetime);
-        return res == null ? "" : res;
+	    return res == null ? "" : res;
 	}
 
 	/**


Jurigag:
Some test if it's fixed could be good.

phalcon/session/adapter/redis.zep
```diff
@@ -118,7 +118,8 @@ class Redis extends Adapter
 	 */
 	public function read(sessionId) -> string
 	{
-		return (string) this->_redis->get(sessionId, this->_lifetime);
+	    let res = (string) this->_redis->get(sessionId, this->_lifetime);
+	    return res == null ? "" : res;
```
sergeyklay: after `let res = (string) ....` the `res` wont be a null.

sergeyklay (changes requested):

phalcon/session/adapter/redis.zep
sergeyklay: also you missed `var res`

SidRoberts: You also need to instantiate the 'res' variable before this line:

    var res;

thilinah:
There is a problem with current implementation. The line
```
return (string) this->_redis->get(sessionId, this->_lifetime);
```

is returning null when the session key is not in store (when using redis session handler)

This causes a fatal error when calling session_regenerate_id

I was using phalcon with php7 and redis session handler and had to extend it and use the extended php session handler with following method. 

```
    public function read($sessionId)
    {
        $val = $this->_redis->get($sessionId, $this->_lifetime);
        return empty($val) ? "" : $val;
    }
```
May be someone can apply a fix based on this

Jurigag:
How this is returning null when it should be casted to `(string)` ?

thilinah:
It could be a problem with zephir in php7. Some how I'm certain redis.zep read function returning null when the sessino key is not found.

branch: 3.0.x
php version: 7

screen shot 2016-10-17 at 12 23 17

Jurigag:
Why you didn't added `(string)` here..... ?

sergeyklay:
Fixed in the `3.0.x` branch.

```sh
git clone git@github.com:phalcon/cphalcon.git
cd cphalcon
git checkout 3.0.x

zephir fullclean
zephir build
```

sergeyklay:
```php
// always will return string, not null
return (string) $this->_redis->get($sessionId, $this->_lifetime);

// just try
var_dump(null);
var_dump((string) null);
```

sergeyklay:
@thilinah
http://php.net/manual/en/language.types.type-juggling.php#language.types.typecasting

thilinah:
@sergeyklay when the null casted to string you might not be getting a null in zephir. But when the redis.zep is extended and called in php7 it is retuning null to php7. Please check the screenshot attached.

@Jurigag the code in screen shot is php. $this->_redis->get calls the get function in phalcon redis adapter compiled from branch 3.0.x. But as you can see it is retuning null. (even though you have casted the result from redis store to string). This breaks "session_regenerate_id" in php7.

I extend the pahlcon redis session handler and apply above to fix the issue.

sergeyklay:
@thilinah Please see at `public function read(sessionId) -> string` .
The `-> string` will prevent to return non string

sergeyklay:
Provide please small script to return non-string, real null.

PR #12335, opened 2016-10-18
3.0.1

DK189:
My project build from Phalcon 3.0.1 ( base on PHP 7.0.11 ).

i detected error in Phalcon\Http\Request !!

When i type `http://mywebsite/controller/action`**/?**`param=value`, it work.
But i type `http://mywebsite/controller/action`**?**`param=value`, it not working right and show the index!! ( IndexController::indexAction )

two url different at **?** !!

i fix it by insert **/** to any link, but i not wanted that!!
**I hope there are better solutions. Thanks for reading!!**

_Sorry, the ability to write my english is poor !!_


Jurigag:
We don't accept PRs to master branch. Please send it to good branch.

PR #12337, opened 2016-10-18
Enabled Validation tests

sergeyklay:
- [x] MySQL
- [x] PostgreSQL
- [x] SQLite


tests/_data/models/Subscriptores.php
```diff
@@ -0,0 +1,77 @@
...
+
+    public function validation()
+    {
+        $validator = new Validation();
+        $validator
+            ->add('created_at', new PresenceOf())
+
+            ->add('email', new StringLength(['min' => '7', 'max' => '50']))
+            ->add('email', new Email())
+            ->add('email', new Uniqueness(['model' => $this]))
```
sergeyklay: does not work?

Jurigag: Work but is not needed in this case. Just doing nothing here.

PR #12343, opened 2016-10-21
[WIP][dont merge] Fix build with image magick 7

Izopi4a:
If you are using latest image magic 7 build fails because  `setImageOpacity` doesnt exists

build fails with 

```

Zephir\CompilerException: Class 'Imagick' does not implement method: 'setImageOpacity' in /home/izopi4a/cphalcon/phalcon/image/adapter/imagick.zep on line 362

          watermark->setImageOpacity(opacity);
        -------------------------------------^

```

https://github.com/mkoppanen/imagick/blob/bd9548c426391878905b2a6bc36accc4c79f7ef7/php_imagick_defs.h#L428

i used this to replace it

http://stackoverflow.com/questions/3538851/php-imagick-setimageopacity-destroys-transparency-and-does-nothing

### DIFF
--- a/phalcon/image/adapter/imagick.zep
+++ b/phalcon/image/adapter/imagick.zep
@@ -359,7 +359,7 @@ class Imagick extends Adapter implements AdapterInterface
 			watermark = new \Imagick();
 
 		watermark->readImageBlob(image->render());
-		watermark->setImageOpacity(opacity);
+		watermark->evaluateImage(\Imagick::EVALUATE_MULTIPLY, opacity, \Imagick::CHANNEL_ALPHA);
 
 		this->_image->setIteratorIndex(0);
 


Jurigag:
But this is gonna broke for php5 ? Doesn't it ?

Izopi4a:
the stackoverflow post is from 2010 so it should work

Izopi4a:
actually ... was wrong, this method doesnt exists in image magick 7 ..  otherwise it exists always

Jurigag:
Well. Maybe if else. I guess we can't in minor version deprecate only some related extension version or something ? Anyway i doubt that everyone is going to update imagick.

Izopi4a:
well since the guy says that this method is not ok, and since it has been removed, probably this new implementation is better ... dont know really, i am not an image magick expert. I do compile it from souce and upgrade it very often cuz of the CVEs that it has from time to time, and I use it for processing many many things :D

But i don't know, as you decide i will update the PR accordingly

sergeyklay:
> Class '\Imagick' does not have a constant called: 'EVALUATE_MULTIPLY' in C:\projects\phalcon\phalcon\image\adapter\imagick.zep on line 362

Izopi4a:
ok i will dive deep into it to see whats the best solution there and let you know when its ok

Izopi4a:
ok its just typo in php imagick.

https://github.com/mkoppanen/imagick/issues/180#issuecomment-255714954

Issue #12355, opened 2016-10-24
[3.0.1] The argument is not initialized or iterable() in phalcon/mvc/view.zep


ss56806298:
After I updated my phalcon from 2.0.X to 3.0.1,the error occured.And when i turned the version back,it runned normally.I'm not sured where the error happened.

```php
$di = new FactoryDefault();

$di->set('view', function() use ($config) {

    $view = new View();

    return $view;
}, true);
```
### Details

* Phalcon version:3.0.1
* PHP Version: 5.6.25
* Operating System: WINDOWS
* Installation type: DLL
* Server:  Apache
* Other related info (Database, table schema): MYSQL 5.7.14

Jurigag:
Just only this code is causing a problem ? Works fine for me.

stamster:
Just get rid of Windows, it'll work fine on GNU.

sergeyklay:
@ss56806298 Could you please provide detailed error log?

ss56806298:
@stamster 
It's convenient to code on Windows.The error only occured when i updated the phalcon,so i guess the latest version may not suit for Windows.
@sergeyklay 
#0 [internal function]: Phalcon\Mvc\View->_engineRender(Array, 'Login/check', true, true, NULL)
 #1 [internal function]: Phalcon\Mvc\View->render('Login', 'check', Array) 
#2 F:\wamp\www\original_server\index.php(22): Phalcon\Mvc\Application->handle() 
#3 {main}The argument is not initialized or iterable()

Jurigag:
`Login/check'` ? I don't see this in your code. Post whole script to reproduce.

SidRoberts:
It could be that [View::getViewsDirs()](https://github.com/phalcon/cphalcon/blob/v3.0.1/phalcon/mvc/view.zep#L1407) isn't returning an array, [as is required in View::_engineRender()](https://github.com/phalcon/cphalcon/blob/v3.0.1/phalcon/mvc/view.zep#L596). Perhaps if View::setViewsDir() isn't used?

@ss56806298 **if you haven't used View::setViewsDir()**, could you try building Phalcon (using Zephir) from https://github.com/SidRoberts/cphalcon/tree/possible-fix-12355 to see if that fixes the problem?

Jurigag:
If there would be script fully reproduce a problem then it would be good.

ss56806298:
@SidRoberts  @Jurigag 
I alreay solved the problem.Here is the  the structure of my code.

[...]

PR #12362, opened 2016-10-26
View::_viewsDirs is initialized as an array

SidRoberts:
See https://github.com/phalcon/cphalcon/issues/12355#issuecomment-256279771


sergeyklay:
Thanks

PR #12366, opened 2016-10-27
Refactored ACL funcAccess

SidRoberts:


phalcon/acl/adapter/memory.zep
```diff
@@ -678,78 +679,94 @@ class Memory extends Adapter
...
+						let userParametersSizeShouldBe--;
+
+						continue;
 					}
-					// We dont have enough parameters
-					else {
-						throw new Exception("You didn't provide all necessary parameters for defined function when check ".roleName." can ".access." ".resourceName);
+
+					// This is some user defined class, check if his parameter is instance of it
+					if isset parameters[parameterToCheck] && typeof parameters[parameterToCheck] == "object" && !reflectionClass->isInstance(parameters[parameterToCheck]) {
```
Jurigag: Well i know i created this but right now im thinking about need of this. Maybe we should just stick to php which will throw error anyway if wrong type will be passed and remove this complicated condition from here ?

phalcon/acl/adapter/memory.zep
```diff
@@ -678,78 +679,94 @@ class Memory extends Adapter
...
+				trigger_error(
+					"Number of parameters in array is higher than the number of parameters in defined function when check " . roleName . " can " . access . " " . resourceName . ". Remember that more parameters than defined in function will be ignored.",
+					E_USER_WARNING
+				);
 			}
+
+			// We dont have any parameters so check default action
+			if count(parametersForFunction) == 0 {
+				if numberOfRequiredParameters > 0 {
+					trigger_error(
```
Jurigag: Well when i was creating it i added this, but should we inform developer that we will use this check ? Or maybe remove this trigger_error ? @sergeyklay

Jurigag:
Also i could try to implement some cache mechanism for reflection api used here for 3.1.x.

sergeyklay:
I'll take a look in few days

sergeyklay:
@Jurigag ?

sergeyklay:
Thanks

PR #12369, opened 2016-10-29
tests/ is now PSR2

SidRoberts:
tests/_bootstrap.php has been restructured slightly but its mostly php-cs-fixer with a few manual corrections.

tests/integration and tests/unit are completely PSR2. tests/_data is largely compliant - existing errors are mostly relating to the lack of namespaces.


Jurigag:
Add in PR template that tests should follow PSR2 rules.

sergeyklay:
Thanks

PR #12373, opened 2016-10-30
Added the ability to truncate DB tables

SidRoberts:
I've purposely left `truncateTable()` out of DialectInterface to maintain BC.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # [3.1.0](https://github.com/phalcon/cphalcon/releases/tag/v3.1.0) (2016-XX-XX)``
 - Added `Phalcon\Validation\Validator\Callback`, `Phalcon\Validation::getData`
+- Added the ability to truncate database tables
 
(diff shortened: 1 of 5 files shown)


sergeyklay:
@SidRoberts Could you please update `CHANGELOG.md`

SidRoberts:
CHANGELOG updated.

sergeyklay:
Thanks

lisek2000 (commented):
- [ ] Hello,

what about calling truncate for table referenced by foreign keys? Current implementation will fail as e.g. mysql will not allow to truncate such table.  You can disable checking by SET FOREIGN_KEY_CHECKS = 0;  but I don't know if it's not too risky for data integrity.

sergeyklay:
@lisek2000 As you can see these are only the dialect generators. They do not do any work, they just create the SQL that can be executed later.

lisek2000:
@sergeyklay my bad. It's ok then.

PR #12374, opened 2016-10-30
cleans up some docblock verbiage

arielallon:


sergeyklay (changes requested):

arielallon:
yep, i'll clean this up.
got a little messy when i saw after the fact that PRs to master directly would be rejected

arielallon:
cleaned up in https://github.com/phalcon/cphalcon/pull/12383
thanks.

PR #12375, opened 2016-10-30
[4.0.0] CLI parameters now work like MVC parameters

SidRoberts:
Taking this as an example:

``` bash
php cli.php greeting morning Sid Roberts
```

Before:

``` php
class GreetingTask extends \Phalcon\Cli\Task
{
    public function morningAction($params)
    {
        echo "Good morning " . $params[0] . " " . $params[1];
    }
}
```

Now:

``` php
class GreetingTask extends \Phalcon\Cli\Task
{
    public function morningAction($firstName, $lastName)
    {
        echo "Good morning " . $firstName . " " . $lastName;
    }
}
```

This, of course, makes `Dispatcher::callActionMethod()` redundant.


niden:
Implemented in https://github.com/phalcon/cphalcon/pull/13630

Issue #12098, opened 2016-08-05
Remove deprecated code


sergeyklay:
- [x] Remove `Phalcon\Mvc\Model\Criteria::order` in favor of `Phalcon\Mvc\Model\Criteria::orderBy`
- [ ] Remove `Phalcon\Mvc\Model\Validator\*` in favor of `Phalcon\Validation\Validator\*`
- [ ] Remove `Phalcon\Cache\Backend\Memcache` in favor of `Phalcon\Cache\Backend\Libmemcached`
- [ ] Remove `Phalcon\Session\Adapter\Memcache` in favor of `Phalcon\Session\Adapter\Libmemcached`
- [x] Remove `Phalcon\Debug::getMajorVersion` due to the fact that we never use this method
- [ ] Remove `Phalcon\Mvc\Collection::validationHasFailed`
- [ ] Remove calling `Phalcon\Mvc\Collection::validate` with object of type `Phalcon\Mvc\Model\ValidatorInterface`
- [ ] Refactor `*\Apc` classes to drop `APC` in favor of `APCu`
  The `apc` extension is considered unmaintained and dead.
  See: http://php.net/manual/en/intro.apc.php
- [ ]  Remove `Phalcon\Mvc\Micro\Lazyloader::__call` in favor of `Phalcon\Mvc\Micro\Lazyloader::callMethod`
- [x] Remove `Phalcon\Cli\Console::addModules` in favor of `Phalcon\Cli\Console::registerModules`
- [ ] The `xcache` support should be deprecated as `xcache` is not compatible with the often by default enabled opcache extension. The `xcache` will not available for PHP 7
- [x] Remove `Phalcon\Dispatcher::setModelBinding` in favor of `Phalcon\Dispatcher::setModelBinder`

Jurigag:
I just wonder why we didn't did it in 3.0.0 already.

sergeyklay:
We just deprecated it in 3.0.0, but not removed

Jurigag:
```
- [] Remove `Phalcon\Mvc\Collection::validationHasFailed`,
- [] Remove calling `Phalcon\Mvc\Collection::validate()` with object of type `Phalcon\Mvc\Model\ValidatorInterface`
```

Jurigag:
Remove `Phalcon\Mvc\Micro\Lazyloader::__call` in favor of `Phalcon\Mvc\Micro\Lazyloader::callMethod`

ruudboon:
[...]

PR #12376, opened 2016-10-30
[4.0.0] Collections now use the Validation component

SidRoberts:
Refs: https://github.com/phalcon/cphalcon/issues/12098


Jurigag:
Adding this already to 3.x.0 won't be hard and still BC. Only instanceof checks and that's it. Question is - are we want do it ?

Jurigag:
Well im adding this for 3.1.0 with BC. So in your PR there will be only need to remove old ability and and old classess.

SidRoberts:
I had to remove some files from ext/ because, previously, AppVeyor reached 100 errors and would stop compiling.

sergeyklay:
Let me deal with this a bit later. I'll try to sort out tomorrow.

sergeyklay:
@SidRoberts Regenerated the build and updated `ext/` dir. Could you please rebase

niden:
Thank you!

PR #12378, opened 2016-10-30
[4.0.0] Response headers and cookies are no longer prematurely sent

SidRoberts:
Refs: #12908


sergeyklay:
Could you please explain a bit more - why?

SidRoberts:
The Application class is effectively responsible for turning HTTP requests into HTTP responses. It shouldn't assume that the response will actually be sent - which is particularly true in testing.

Application requests should be handled like this:

```php
$response = $application->handle($uri);

$response->send();
```

Not this:

```php
echo $application->handle($uri)->getContent();
```

sergeyklay:
Thank you

PR #12379, opened 2016-10-30
[4.0.0] Tag no longer uses static methods

SidRoberts:


yankos:
Hm... Is it good?
Tag as class with static method is not ideal. But this PR do not bring something new.
I think much better if Tag will be as service.

Jurigag:
Yea, don't like creating new tag object in each method.

sergeyklay:
@Jurigag Just inject Tag to the Di and use `$this->tag`, `$di->get('tag')`. And yes, I agree with @yankos about service

Jurigag:
Yea i know, but it should happen then in this PR everywhere i think. As well this would require Tag in factorydefault perhaps.

yankos:
@Jurigag of course Tag must be in factorydefault.

So, maybe it's time to refactor Tag as service for 4 version of Phalcon? 
I can to do this.

SidRoberts:
Static methods are glorified functions with namespaces. Using them in this way means we can't override them with a custom tag class and they're hard dependencies. It defeats the whole point of object-oriented programming. For all intents and purposes, they're effectively evil. :p

This PR is still a work in progress and I plan to get the 'tag' service from the DI instead of just creating a new instance every time. Unfortunately the PR doesn't pass in its current form so I'm focusing on that first.

phalcon/assets/manager.zep
```diff
@@ -717,12 +717,14 @@ class Manager
...
 	public function outputInline(<Collection> collection, type) -> string
 	{
-		var output, html, codes, filters, filter, code, attributes, content, join, joinedContent;
+		var output, html, codes, filters, filter, code, attributes, content, join, joinedContent, tag;
 
 		let output = "",
 			html = "",
 			joinedContent = "";
 
+		let tag = new Tag();
```
sergeyklay: @SidRoberts Can we seek it in the Di first?

phalcon/assets/manager.zep
```diff
@@ -777,15 +779,17 @@ class Manager
...
+		var collection, tag;
 
 		if !collectionName {
 			let collection = this->getCss();
 		} else {
 			let collection = this->get(collectionName);
 		}
 
-		return this->output(collection, ["Phalcon\\Tag", "stylesheetLink"], "css");
+		let tag = new Tag();
```
sergeyklay: The same

phalcon/assets/manager.zep
```diff
@@ -813,15 +817,17 @@ class Manager
...
+		var collection, tag;
 
 		if !collectionName {
 			let collection = this->getJs();
 		} else {
 			let collection = this->get(collectionName);
 		}
 
-		return this->output(collection, ["Phalcon\\Tag", "javascriptInclude"], "js");
+		let tag = new Tag();
```
sergeyklay: and here

sergeyklay (changes requested):

niden:
Addressed in https://github.com/phalcon/cphalcon/pull/13656

PR #12383, opened 2016-10-31
fixing verbiage in the docblocks of some methods

arielallon:
cleaned up version of PR https://github.com/phalcon/cphalcon/pull/12374


sergeyklay:
Thanks

Issue #12370, opened 2016-10-29
[BUG] Assets Collection does not join when no filter is set.


elburgl69:
Phalcon\Assets\Collection->Join(true) does not join resources when there is no Filter set on the collection.

Expected behavior:
When issueing:
        Phalcon\Assets\Collection->Join(true)
the added resources should be joined. It outputs however the (not merged) individual added resources.

When you assign a filter to the collection, it does Join the resources.

Even when you add only resources that should not be filtered (fe Collection::addJS($file, false, false), Join() does not function when no filter is assigned and does join when a filter is assigned to the collection!

From a design perspective, when would expect Join to function independently from any applied filter, for example to easily test in which order added files are merged in development or test environments.

gguridi:
I think this issue has already been fixed.

PR #12384, opened 2016-10-31
Fix for issue #12370

elburgl69:
In addition commented out a obsoleet if statement.


sergeyklay:

PR #12386, opened 2016-11-01
Fix for issue #12370

elburgl69:
Fixed issue #12370

PR #12388, opened 2016-11-02
[3.0.x] Fix query keys method with prefix in file cache

Fishdrowned:
The passed prefix should not be compared to the filename directly, it should be parsed into filename, just like it did in save/get/delete


sergeyklay (approved):
Could you please update `CHANGELOG.md`

Fishdrowned:
Change log updated.

Commits squashed, thank you @Jurigag for the instruction in #12386

sergeyklay:
Thanks

PR #12389, opened 2016-11-02
[3.1.x] Fix query keys method with prefix in file cache

Fishdrowned:
The passed prefix should not be compared to the filename directly, it should be parsed into filename, just like it did in save/get/delete

(cherry picked from commit b45d6b4ded6ffb84967c044933f9850385c0b9b8)


sergeyklay:
Thanks

PR #12406, opened 2016-11-17
Phalcon\Cache\Frontend\Igbinary

dred86:
declare _frontendOptions


sergeyklay (changes requested):
See https://github.com/phalcon/cphalcon/pull/12407#pullrequestreview-9012646

sergeyklay:
How about removing __construct at all and use parent contructor instead ?

PR #12407, opened 2016-11-17
Phalcon\Cache\Frontend\Msgpack

dred86:
declare _frontendOptions


sergeyklay (changes requested):
Could you please remove 

```
let this->_frontendOptions = frontendOptions;
```

from `__construct` and use parent::__construct

PR #12410, opened 2016-11-19
Escaping DB tables is now done solely by DB dialects

### LINKED ISSUE #12408: Method escapeIdentifier of \Phalcon\Db\Adapter\Pdo\Mysql does not actually escape anything
Method \Phalcon\Db\Adapter\Pdo\Mysql::escapeIdentifier does not actually escape supplied identifier. Method only adds ` before and after identifier.
From its name I assumed that a value passed through this method would be safe to use in a SQL-statement, but this opens an attack vector for SQL-injection.

Below code uses hard-coded identifier and is thus safe, but if identifier is dynamic and possibly of user origin, an attack could be possible.

https://github.com/phalcon/cphalcon/blob/v3.0.1/phalcon/db/adapter/pdo/mysql.zep#L66

Prerequisites:
```sql
CREATE TABLE `test`.`as``d` (
    `id` INTEGER UNSIGNED NOT NULL AUTO_INCREMENT,
  PRIMARY KEY (`id`)
);
```
```php
$db = new \Phalcon\Db\Adapter\Pdo\Mysql(['host' => '127.0.0.1', 'username' => 'root']);
$sql = "SELECT * FROM {$db->escapeIdentifier(['test', 'as`d'])}";
var_dump($sql);
var_dump(count($db->fetchAll($sql)));
```
Expected:
```php
string(28) "SELECT * FROM `test`.`as``d`"
int(0)
```
Actual:
```php
PDOException : SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '`' at line 1
 
string(27) "SELECT * FROM `test`.`as`d`"
```
### Details
* Phalcon version: 3.0.1
* PHP Version: PHP 5.5.24 MSVC11 TS
* Operating System: Windows 7
* Installation type: xampp
* Server: Apache and CLI

SidRoberts:
Fixes #12408

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - Fixed `Phalcon\Acl\Adapter\Memory::isAllowed` to call closures when using wildcard [#12333](https://github.com/phalcon/cphalcon/issues/12333)
 - Fixed `Phalcon\Validation\Validator\File` array to string conversion in `minResolution` and `maxResolution` [#12349](https://github.com/phalcon/cphalcon/issues/12349)
 - Fixed `Phalcon\Cache\Backend\File::queryKeys()` to compare the filename against parsed prefix
+- Database identifiers are now properly escaped [#12410](https://github.com/phalcon/cphalcon/issues/12410)
 
 # [3.0.1](https://github.com/phalcon/cphalcon/releases/tag/v3.0.1) (2016-08-24)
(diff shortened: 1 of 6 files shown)


sergeyklay:
@SidRoberts Could you please update `CHANGELOG.md`

sergeyklay:
Thanks

PR #12411, opened 2016-11-19
Removed Db\Column::_schemaName as its not used or accessible

SidRoberts:


sergeyklay:
@SidRoberts Could you please explain a bit more?

SidRoberts:
It can't be used anywhere as there is no setter / variable assignment - it will **always** be `null`. The getter isn't used anywhere either so it's effectively dead code. It's also important to note that there isn't a property for the table name so knowing the schema name is sort of pointless.

sergeyklay:
@SidRoberts is it `v4.x`-related changes. Right?

SidRoberts:
It depends on your mood. :stuck_out_tongue:  I consider `_schemaName` and `getSchemaName()` to be an accidental addition and I think it should be in 3.1.x but we are technically changing the interface (albeit removing a method, not adding).

sergeyklay:
In general I'm not sure about this PR. I couldn't find a convincing reason - what problem it solves. 
Thank you for contributing

PR #12414, opened 2016-11-21
ModelManager::isVisibleModelProperty() now caches model properties properly

SidRoberts:
Before, the `_modelVisbility` property wasn't actually used.


sergeyklay:
Thanks

PR #12429, opened 2016-11-26
Fixed call functions to null variable in Phalcon\Forms\Form

Studentsov:
By default _elements value isn't defined that can cause errors.
For example:
```
<?php

use Phalcon\Forms\Form;

class MyForm extends Form
{
    public function initialize()
    {
        
    }
}

$form = new MyForm();

foreach ($form as $element) {
    //...
}
```
There will be an error in the Phalcon\Forms\Form::rewind method: Warning: array_values() expects parameter 1 to be array, null given.
My patch will correct it.


sergeyklay:
@Studentsov Could you please add test for `rewind` and update `CHANGELOG.md`?

sergeyklay:
Thanks

PR #12438, opened 2016-11-27
Cache\Backend\Redis::get method bugfix

mbrostami:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/12437

Small description of change: When you put a key with empty value in redis cache backend, get method returns false for that key.

Thanks

When there is a key with value of empty string (""), get method MUST return empty string (""), but it didn't.

### DIFF
--- a/phalcon/cache/backend/redis.zep
+++ b/phalcon/cache/backend/redis.zep
@@ -158,7 +158,7 @@ class Redis extends Backend implements BackendInterface
 		let this->_lastKey = lastKey;
 		let cachedContent = redis->get(lastKey);
 
-		if !cachedContent {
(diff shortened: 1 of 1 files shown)


sergeyklay:
@mbrostami Thank you for pointing the problem. But actually your PR does not solve it. Because frontend's `unserialize` will return `FALSE` for empty string. This is why I would like to see the tests for these changes. 

I hope I managed to solve the problem in https://github.com/phalcon/cphalcon/pull/12440

mbrostami:
Of course you are right, I just tried RedisCache with Phalcon\Cache\Frontend\None. :)

PR #12445, opened 2016-11-30
[WIP] Added Phalcon\Validation for Phalcon\Mvc\Collection

Jurigag:
Hello!

* Type:  new feature

- [x] Extensive Uniqueness tests

Small description of change: This add option to use `Phalcon\Validation` for `Phalcon\Mvc\Collection` while keeping BC for old syntax. Refs #12376

Thanks

### DIFF
--- a/phalcon/validation/validator/uniqueness.zep
+++ b/phalcon/validation/validator/uniqueness.zep
@@ -23,7 +23,10 @@ use Phalcon\Validation;
 use Phalcon\Validation\CombinedFieldsValidator;
 use Phalcon\Validation\Exception;
 use Phalcon\Validation\Message;
+use Phalcon\Mvc\ModelInterface;
+use Phalcon\Mvc\CollectionInterface;
 use Phalcon\Mvc\Model;
+use Phalcon\Mvc\Collection;
 
 /**
  * Phalcon\Validation\Validator\Uniqueness
@@ -139,16 +142,7 @@ class Uniqueness extends CombinedFieldsValidator
 
 	protected function isUniqueness(<Validation> validation, var field) -> boolean
 	{
-		var value, values, convert, record, attribute, except,
-			index, params, metaData, primaryField, className, singleField,
-			fieldExcept, singleExcept, notInValues, exceptConditions;
-
-		let exceptConditions = [];
-		let index  = 0;
-		let params = [
-			"conditions": [],
-			"bind": []
-		];
+		var values, convert, record, params, className, isModel, singleField;
 
 		if typeof field != "array" {
 			let singleField = field;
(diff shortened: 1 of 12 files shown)


Jurigag:
Is there any mongo expert ? Can you check if uniqueness validator logic is correct ?

Jurigag:
Anyone knows what's wrong here - https://ci.appveyor.com/project/phalcon/cphalcon/build/1.0.911#L183 ?

phalcon/validation/validator/uniqueness.zep
```diff
@@ -185,52 +186,97 @@ class Uniqueness extends CombinedFieldsValidator
...
 				elseif count(field) == 1 {
-					let params["conditions"][] = attribute . " <> ?" . index;
-					let params["bind"][] = except;
-					let index++;
+					if isModel {
+						let params["conditions"][] = attribute . " <> ?" . index;
+						let params["bind"][] = except;
+						let index++;
+					} else {
+						array arrayValue = ["$nin": except];
```
Jurigag: $ne here

phalcon/validation/validator/uniqueness.zep
```diff
@@ -185,52 +186,97 @@ class Uniqueness extends CombinedFieldsValidator
...
+				} else {
+					let params[] = [attribute: null];
+				}
 			}
 
 			if except {
 				if typeof except == "array" && count(field) > 1 {
-					if isset except[singleField] {
-						let fieldExcept = except[singleField];
+					if isset except[attribute] {
```
Jurigag: Not needed change here.

sergeyklay (approved):

Jurigag (commented):
Add extensive Uniquness test for Collection like already existing one `Phalcon\Test\Unit\Validation\Validator\UniquenessTest` for Model to make sure it works fine.

phalcon/validation/validator/uniqueness.zep
```diff
@@ -185,52 +186,97 @@ class Uniqueness extends CombinedFieldsValidator
 				}
 			}
 
-			let attribute = this->getColumnNameReal(record, this->getOption("attribute", singleField)),
-				except = this->getOption("except");
+			let isModel = record instanceof Model;
```
virgofx: You're in loop here. Let's get the instance type of the record immediately. So move this to the top.

Then, to avoid all the if/elseif confusion, let's separate the logic for the new collection parts. This way we can actually leave the existing Model code AS IS.
```
    if (record instanceof ...\Model) {
        $this->_isUniquenessModel(validation, field, record);
    }
    elseif (record instanceof ...\Collection) {
        $this->_isUniquenessCollection(validation, field, record);
    }
```

So move existing `isUniqueness()` to `_isUniquenessModel` and then create new one for Collection accordingly.

sergeyklay:

Jurigag:
Yea will do with the rest of fixes and changes when will sit to it.

phalcon/validation/validator/uniqueness.zep
sergeyklay: I would like to check for `ModelInterface` here instead `Model`

phalcon/validation/validator/uniqueness.zep
```diff
@@ -185,52 +186,97 @@ class Uniqueness extends CombinedFieldsValidator
...
+
+			if !(isModel || record instanceof Collection) {
```
sergeyklay: Could you please use `CollectionInterface` here. Also it seems to me the parentheses not needed

phalcon/validation/validator/uniqueness.zep
```diff
@@ -185,52 +186,97 @@ class Uniqueness extends CombinedFieldsValidator
...
+				throw new Exception("The uniqueness validator works only with Phalcon\\Mvc\\Model or Phalcon\\Mvc\\Collection");
+			}
+
+			if isModel {
+				let attribute = this->getColumnNameReal(record, this->getOption("attribute", singleField));
```
sergeyklay: I would like to avoid constructions like `this->method(this->method())`.
Could you please use local variable to avoid performance regression

sergeyklay (changes requested):
In general, now `Phalcon\Validation\Validator\Uniqueness::isUniqueness` has 173 lines and this is very much. It seems to me that now the `Phalcon\Validation\Validator\Uniqueness::isUniqueness` has too much responsibility and I don't like that it can do anything (except coffee); I would like to perform this checks with the help of short methods to reduce cyclomatic complexity and add an opportunity to test/mock these methods independently

phalcon/validation/validator/uniqueness.zep
Jurigag: Yea i will. Keep in ming it was original code i think like this.

phalcon/validation/validator/uniqueness.zep
Jurigag: Then how to check if record is not instanceof collection without parentheses ?

Jurigag: Not sure if parser issue. Well i mean if we could do something `!instanceof` could be nice.

Jurigag:
This is ready to merge i think.

sergeyklay:
Thanks

PR #12451, opened 2016-12-02
feat(Micro): Response handler

Doehl:
* Type: new feature

**Small description of change:**

**What:**
Adds the possibility to create a response handler, that will handle the
response instead of the default hard coded micro handler.

**Why:**
Before the reponse format had to be coded into every route.
Now is it possible to create a custom rosponse handler, that will format
the data in custom created function.

**Changes to phalcon/mvc/micro.zep**
* Added **protected** class variable **_responseHandler**
_The custom response handler_

* Added **public** function **response** 
_Appends a response handler_

* Change to **public** function  **handle** 
_Check if a response handler has been appended._
_If **yes**, run the response handler._
_If **no**, run the same code as before this PR._

**How to use**
```php
$app->response(function () use ($app) {

    // Gets the client response, that has been returned from executed route
    $return = $app->getReturnedValue();

    // Put controller return array item data
    $return = array('data' => $return);

    // Set reponse Content-Type to json
    header('Content-Type: application/json');

    // Print json response string
    echo json_encode($return);
});
```


Jurigag:
New feature so 3.1.x branch please. Also change changelog.md and write some test.

Jurigag (changes requested):
New feature so 3.1.x branch please. Also change changelog.md and write some test.

sergeyklay:
@Doehl We use the 3.0.x only for bugfixes. Could you please submit PR to the 3.1.x branch

Doehl:
I will create another PR for the correct branch.

PR #12452, opened 2016-12-03
[NFR] Micro: Response handler

Doehl:
* Type: new feature

## Small description of change:

**What:**
Adds the possibility to create a response handler, that will handle the
response instead of the default hard coded micro handler.

**Why:**
Before the reponse format had to be coded into every route.
Now is it possible to create a custom rosponse handler, that will format
the data in custom created function.

**Changes to phalcon/mvc/micro.zep**
* Added **protected** class variable **_responseHandler**
_The custom response handler_

* Added **public** function **response** 
_Appends a response handler_

* Change to **public** function  **handle** 
_Check if a response handler has been appended._
_If **yes**, run the response handler._
_If **no**, run the same code as before this PR._

## Unit test
Adds a response handler, and check if it has been called.

## How to use
```php
$app->setResponseHandler(function () use ($app) {

    // Gets the client response, that has been returned from executed route
    $return = $app->getReturnedValue();

    // Put controller return array item data
    $return = array('data' => $return);

    // Set reponse Content-Type to json
    header('Content-Type: application/json');

    // Print json response string
    echo json_encode($return);
});
```

### DIFF
--- a/phalcon/mvc/micro.zep
+++ b/phalcon/mvc/micro.zep
@@ -80,6 +80,8 @@ class Micro extends Injectable implements \ArrayAccess
 
 	protected _modelBinder;
 
+	protected _responseHandler;
+
 	protected _afterBindingHandlers;
 
 	/**
@@ -980,24 +982,39 @@ class Micro extends Injectable implements \ArrayAccess
 			}
 		}
 
-		/**
-		 * Check if the returned value is a string and take it as response body
-		 */
-		if typeof returnedValue == "string" {
-			let response = <ResponseInterface> dependencyInjector->getShared("response");
-			response->setContent(returnedValue);
-			response->send();
-		}
 
 		/**
-		 * Check if the returned object is already a response
-		 */
-		if typeof returnedValue == "object" {
-			if returnedValue instanceof ResponseInterface {
-				/**
-				 * Automatically send the response
-				 */
-				returnedValue->send();
+		* Check if a response handler is defined, else use default response handler
+		*/
+		if this->_responseHandler {
+
+			if !is_callable(this->_responseHandler) {
+				throw new Exception("Response handler is not callable or is not defined");
+			}
+
+			let returnedValue = call_user_func(this->_responseHandler);
+
+		} else {
+
+			/**
+			* Check if the returned value is a string and take it as response body
+			*/
+			if typeof returnedValue == "string" {
+				let response = <ResponseInterface> dependencyInjector->getShared("response");
+				response->setContent(returnedValue);
+				response->send();
+			}
+
+			/**
+			* Check if the returned object is already a response
+			*/
+			if typeof returnedValue == "object" {
+				if returnedValue instanceof ResponseInterface {
+					/**
+					* Automatically send the response
+					*/
+					returnedValue->send();
+				}
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


Jurigag:
Squash commits please.

Doehl:
The commits has been squashed.

sergeyklay:
@Doehl Could you please rebase

Doehl:
Rebase done

phalcon/mvc/micro.zep
```diff
@@ -1152,6 +1169,18 @@ class Micro extends Injectable implements \ArrayAccess
 	}
 
 	/**
+	 * Appends a custom 'reponse' handler to be called insted of the default reponse handler
+	 *
+	 * @param callable handler
+	 * @return \Phalcon\Mvc\Micro
+	 */
+	public function response(handler) -> <Micro>
```
sergeyklay: I would like that this method was named `setResponseHandler`

sergeyklay (changes requested):

phalcon/mvc/micro.zep
```diff
@@ -1152,6 +1169,18 @@ class Micro extends Injectable implements \ArrayAccess
 	}
 
 	/**
+	 * Appends a custom 'reponse' handler to be called insted of the default reponse handler
+	 *
+	 * @param callable handler
+	 * @return \Phalcon\Mvc\Micro
```
sergeyklay: the `@return` not needed here

phalcon/mvc/micro.zep
```diff
@@ -980,24 +982,39 @@ class Micro extends Injectable implements \ArrayAccess
...
-				returnedValue->send();
+		* Check if a response handler is defined, else use default response handler
+		*/
+		if this->_responseHandler {
+
+			if !is_callable(this->_responseHandler) {
+				throw new Exception("Response handler is not callable or is not defined");
+			}
+
+			call_user_func(this->_responseHandler);
```
sergeyklay: How about `let returnedValue = call_user_func(this->_responseHandler);`

phalcon/mvc/micro.zep
Doehl: I did try to keep it the same as "after", "before" and "finish" functions.
But if that is how we should name function, I will of course change it.

phalcon/mvc/micro.zep
Doehl: Remove it I will.

Doehl:
All the changes has been made.

sergeyklay (approved):

sergeyklay:
@virgofx Could you please take a look

sergeyklay:
Cc: @phalcon/core-team

niden:
@Doehl Would you be so kind as to rebase this to the `3.4.x` branch and also issue one more for the `4.0.x` branch? 

Thanks for the work!

Doehl:
Hi niden,

Sorry for the late response.
I do not use the phalcon framework at the moment, and does no longer have access to the setup on where I had the phalcon environment running. Because of change on job.

I would love to help, but do no longer have access to an environment where I can easily help with this PR any longer.

niden:
@Doehl Not a problem. I can just "move" your work to a different branch so that you can get credit for this work :)

Thank you again and whenever you have the time we would welcome your contributions.

gspgsp:
2018/12/28 10:37:48 [error] 14931#0: *18 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Uncaught BadMethodCallException: Wrong number of parameters in /www/wechat-api/app/bootstrap/bootstrap.php:13

my bootstrap.php code is:
<?php
use Phalcon\Mvc\Micro;
use Quickapi\App\Service\Helper;

$micro = new Micro($di);
Helper::mountRoutes($micro, $routes); #$routes is in the same document
$micro->handle(); #this is line 13 , i don't konw why ,befor in php5.6  phalcon version 3, there is no problem ,today  i use php7.2 and the phalcon version is also update to 4.0

niden:
@gspgsp Please have a look at this blog post for a list of things that changed:

https://blog.phalconphp.com/post/upgrading-to-v4-alpha-1

Something you are using in your bootstrap file has changed and now you need to adjust your parameters.

Issue #12465, opened 2016-12-09
Form setValidation is not work correctly 


mbrostami:
When I want to use setValidation for form, it doesn't validate elements!

CustomForm.php
```php
<?php

use Phalcon\Forms\Form;
use Phalcon\Forms\Element\Text;

class CustomForm extends Form
{
    public function initialize($entity = null, $options = array())
    {
        $this->add(new Text("name"));
    }
}
```
CustomValidation.php
```php
<?php

use Phalcon\Validation;
use Phalcon\Validation\Validator\PresenceOf;

class CustomValidation extends Validation
{
    public function initialize()
    {
        $this->add('name', new PresenceOf([
            'message' => 'Name is required'
        ]));
    }
}
```
Controller
```php
class IndexController extends Controller
{
    public function testAction()
    {
        $customForm = new CustomForm();
        $customForm->setValidation(new CustomValidation());
        $customForm->isValid($this->request->getPost()); // BUG: if you don't post name, this line returns true! 
    }
}
```

Jurigag:
What a hell is `setValidation` method ?

Jurigag:
https://github.com/phalcon/cphalcon/blob/master/phalcon/forms/form.zep#L256

This is current implementation of isValid method. It's just expecting any validators added to elements, if you didnt added anything then nothing will be validated.

mbrostami:
So why there is a PUBLIC setter/getter for validation in form.zep!!!

Jurigag:
I don't know.

sergeyklay:
Fixed in the `3.0.x` branch. Thanks you

Issue #11500, opened 2016-03-06
[2.1.x] $form->isValid() doesn't merge validators (separate Validation class)


NZX-DeSiGN:
Hi there,

I use the version of 2.1.0RC1 Phalcon and I think there is a bug in the function isValid class Phalcon\Forms\Form. When a separate Validation class is used andbinded with a form, the method isValid() overwrite validators declared in the initialize() method of the Validation.
Sorry for my English and see below the context, bug and correction (if I'm not wrong).

**Validation class**

``` php
<?php

namespace App\Validations;

use Phalcon\Validation;
use Phalcon\Validation\Validator\StringLength;

class SaveValidation extends Validation
{
    public function initialize()
    {
        $this->add('name', new StringLength(array(
            'min'            => 3,
            'messageMinimum' => "Name too short (3 chars min)"
        )));
    }
}
```

**Controller class**

``` php
<?php
namespace App\Admin\Controllers;

use App\Models\Category;
use App\Validations\SaveValidation;

class CategoryController extends ControllerBase
{
    public function newAction()
    {
        // Create new form
        $form = new NewForm();

        // IF is POST
        if ($this->request->isPost()) {
            // Create the new category
            $category = new Category();

            // Updates entity attributes
            $form->bind($_POST, $category);

            // Set form validation
            $form->setValidation( new SaveValidation() );

            // IF the form is valid, save the category
            if ($form->isValid($_POST, $category)) {
                // Save in DB
                if ($category->save()) {
                    $this->flashSession->notice("Success");
                }
            }
        }
    }
}
```

**Phalcon\Forms\Form class with the bug and correction**

``` php
/**
 * Phalcon\Forms\Form
 */
class Form extends Injectable implements \Countable, \Iterator
{
    public function isValid(var data = null, var entity = null) -> boolean
    {         
        /*  [....]   */

        /**
         * Prepare the validators
         */
        let preparedValidators = [];

        for validator in validators {
            let preparedValidators[] = [name, validator];
        }

        let validation = this->getValidation();
        if typeof validation == "object" {
            if validation instanceof Validation {
                /**
                 * Set the validators to the validation
                 */
                validation->setValidators(preparedValidators);  // Replace this line (326) by the next line
                validation->rules(preparedValidators);          // Replace with rules method to add and not override validators
            }
        } else {
            /**
             * Create an implicit validation
             */
            let validation = new Validation(preparedValidators);
        }

        /*  [....]   */
    }
}
```

[...]

Issue #11135, opened 2015-11-14
Form validation for field not possible


KennethVeipert:
Hi!
If i try to get the messages for a single field i get nothing, but if i get the messages of the form for the field with "$form->getMessagesFor('xyz')" it works correctly.

Code to reproduce:

``` php
// Form field
            $element = $this->add(new Element\Mail('mail'));
            $element->addValidator(
                new Validator\PresenceOf(array(
                    'message' => $this->getDi()->get('translator')->query('front.form.error.PresenceOf'),
                ))
            );

// Controller for example
            $mail = $form->get('mail');
            echo '<pre>';
            var_dump($mail->getMessages()); // Wrong - No messages
            var_dump($form->getMessagesFor('mail')); // Correct - PresenceOf Message
            echo '</pre>';
```

sergeyklay:
Fixed in the `3.0.x` branch. Thanks you

Issue #3167, opened 2014-12-28
Phalcon\Forms\Element hasMessages() always returns false


ckip:
Should these 2 lines of code return the same value ?

1) $form->hasMessagesFor('name');
2) $form->get('name')->hasMessages();

Before validation both these lines return false. But when the form validation fails, the first line returns true (which is good) while the second line still return false.
##

sergeyklay:
Fixed in the `3.0.x` branch. Thanks you

PR #12466, opened 2016-12-09
Form validation some fixes

mbrostami:
Hello!

* Type: bug fix
* Link to issues (this will fix all of the below issues):
https://github.com/phalcon/cphalcon/issues/12465
https://github.com/phalcon/cphalcon/issues/11500
https://github.com/phalcon/cphalcon/issues/11135
https://github.com/phalcon/cphalcon/issues/3167

Small description of change:
setValidation for form is not work correctly as I explained at #12465.
Elements messages doesn't set if custom validation fails.

Thanks

### DIFF
--- a/phalcon/forms/form.zep
+++ b/phalcon/forms/form.zep
@@ -290,17 +290,19 @@ class Form extends Injectable implements \Countable, \Iterator
 			}
 		}
 
-		let notFailed = true,
-			messages = [];
+		let validationStatus = true;
+
+        let validation = this->getValidation();
+
+        if !(validation instanceof Validation) {
+            // Create an implicit validation
+            validation = new Validation();
+        }
 
 		for element in this->_elements {
 
 			let validators = element->getValidators();
-			if typeof validators != "array" {
-				continue;
-			}
-
-			if count(validators) == 0 {
+			if typeof validators != "array" || count(validators) == 0 {
 				continue;
 			}
 
@@ -309,25 +311,11 @@ class Form extends Injectable implements \Countable, \Iterator
 			 */
 			let name = element->getName();
 
-			/**
-			 * Prepare the validators
-			 */
-			let preparedValidators = [];
-
+            /**
+            * Append (not overriding) element validators to validation class
+            */
 			for validator in validators {
-				let preparedValidators[] = [name, validator];
-			}
-
-			let validation = this->getValidation();
-
-			if typeof validation == "object" {
-				if validation instanceof Validation {
-					// Set the validators to be validated
-					validation->setValidators(preparedValidators);
-				}
-			} else {
-				// Create an implicit validation
-				let validation = new Validation(preparedValidators);
+			    validation->add(name, validator);
 			}
 
 			/**
@@ -341,21 +329,24 @@ class Form extends Injectable implements \Countable, \Iterator
 			if typeof filters == "array" {
 				validation->setFilters(name, filters);
 			}
-
-			/**
-			 * Perform the validation
-			 */
-			let elementMessages = validation->validate(data, entity);
-			if count(elementMessages) {
-				let messages[name] = elementMessages;
-				let notFailed = false;
-			}
 		}
+
+        /**
+        * Perform the validation
+        */
+        let messages = validation->validate(data, entity);
+        if count(messages) {
+            // Add validation messages to relevant elements
+            for elementMessage in messages {
+                this->get(elementMessage->getField())->appendMessage(elementMessage);
+            }
+            validationStatus = false;
+        }
 
 		/**
 		 * If the validation fails update the messages
 		 */
-		if !notFailed {
+		if !validationStatus {
 			let this->_messages = messages;
 		}
 
@@ -369,7 +360,7 @@ class Form extends Injectable implements \Countable, \Iterator
 		/**
 		 * Return the validation status
 		 */
-		return notFailed;
+		return validationStatus;
 	}
 
 	/**
@@ -386,6 +377,10 @@ class Form extends Injectable implements \Countable, \Iterator
 			}
 			return messages;
 		}
+
+        if messages instanceof Group {
+            return messages;
+        }
 
 		let group = new Group();
 
--- a/phalcon/forms/form.zep
+++ b/phalcon/forms/form.zep
@@ -20,6 +20,7 @@
 namespace Phalcon\Forms;
 
 use Phalcon\Validation;
+use Phalcon\ValidationInterface;
 use Phalcon\DiInterface;
 use Phalcon\FilterInterface;
 use Phalcon\Di\Injectable;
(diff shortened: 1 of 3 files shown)


phalcon/forms/form.zep
```diff
@@ -290,17 +290,19 @@ class Form extends Injectable implements \Countable, \Iterator
 			}
 		}
 
-		let notFailed = true,
-			messages = [];
+		let validationStatus = true;
```
sergeyklay: In Zephir, all variables used in a given scope must be declared. This process gives important information to the compiler to perform optimizations and validations. This is why tests fails.

phalcon/forms/form.zep
```diff
@@ -290,17 +290,19 @@ class Form extends Injectable implements \Countable, \Iterator
...
+
+        let validation = this->getValidation();
+
+        if !(validation instanceof Validation) {
+            // Create an implicit validation
+            validation = new Validation();
```
sergeyklay: `let` statement is used to mutate variables, properties and arrays. Variables are by default immutable and this instruction makes them mutable:
```zephir
let validation = new Validation();
```

phalcon/forms/form.zep
```diff
@@ -290,17 +290,19 @@ class Form extends Injectable implements \Countable, \Iterator
 			}
 		}
 
-		let notFailed = true,
-			messages = [];
+		let validationStatus = true;
+
+        let validation = this->getValidation();
+
+        if !(validation instanceof Validation) {
```
sergeyklay: here does not need braces

phalcon/forms/form.zep
```diff
@@ -341,21 +329,24 @@ class Form extends Injectable implements \Countable, \Iterator
...
+        /**
+        * Perform the validation
+        */
+        let messages = validation->validate(data, entity);
+        if count(messages) {
+            // Add validation messages to relevant elements
+            for elementMessage in messages {
+                this->get(elementMessage->getField())->appendMessage(elementMessage);
+            }
+            validationStatus = false;
```
sergeyklay:
```zephir
let validationStatus = false;
```

sergeyklay (changes requested):

phalcon/forms/form.zep
```diff
@@ -290,17 +290,19 @@ class Form extends Injectable implements \Countable, \Iterator
...
 
-		let notFailed = true,
-			messages = [];
+		let validationStatus = true;
 
+        let validation = this->getValidation();
+
+        if !(validation instanceof Validation) {
+            // Create an implicit validation
+            let validation = new Validation();
```
sergeyklay: Well, I think I misled you. The exclamation point has a higher priority. So could you please use:
```zephir
if !(validation instanceof Validation) {
```

But actually I think that more flexibility was to use the interface but not concrete class

mbrostami: Yeah! You're right. Thanks for comments.

phalcon/forms/form.zep
```diff
@@ -369,51 +362,30 @@ class Form extends Injectable implements \Countable, \Iterator
...
 	{
-		var messages, group, elementMessages;
+		var messages;
 
 		let messages = this->_messages;
-		if byItemName {
-			if typeof messages != "array" {
-				return new Group();
-			}
-			return messages;
```
mbrostami: I think, returning multiple type of value is not a good idea (see line 378 old - return type).

sergeyklay:
@mbrostami Could you please split this PR into parts: one that can be merged to the 3.0.x branch without backward compatibility breaking and another one that need to be sent to the `4.0.x` branch. Also I would like to see a more detailed explanation of your changes.

Thanks

phalcon/forms/form.zep
```diff
@@ -712,7 +684,11 @@ class Form extends Injectable implements \Countable, \Iterator
 	public function rewind() -> void
 	{
 		let this->_position = 0;
-		let this->_elementsIndexed = array_values(this->_elements);
```
mbrostami: If you create a form without any element, and iterate that, php warning will be shown, because this->_elements is null. I added some code to prevent that. https://github.com/phalcon/cphalcon/blob/3.0.x/tests/unit/Forms/FormTest.php#L77

mbrostami:
I think using _messages as a single type (Group|null) would be better, and will avoid of complexity.
In old code, you can see _messages as an array and somewhere else as a Group object!  
Other issue was about a bug for ``getMessagesFor`` method, which didn't return correct value(#11135).   So inside isValid method, after validation, setting error messages to each element fixes this issue and also after that you can get each element messages from ```this->get(elementName)->getMessages()``` and doesn't need to iterate ``_messages``.(would be cleaner code I think). 
Other issue is injecting CustomValidation to a Form by ``setValidation`` method, it didn't work, now is fixed. (#12465 I don't know if it was a bug or not!).

sergeyklay (commented):
Cc: @niden @andresgutierrez Could you please take a look

sergeyklay:
@mbrostami  Could you please rebase, update CHANGELOG.md and cover issues by small test you successfully solved

mbrostami:
@sergeyklay Sure.

mbrostami:
@sergeyklay can you please check appveyor error?  Is it my bad!?

Jurigag:
`Inner Exception: Unable to connect to the remote server` no

mbrostami:
This issue (https://github.com/phalcon/cphalcon/issues/12395) will also be fixed by this PR.

phalcon/forms/form.zep
```diff
@@ -369,51 +362,30 @@ class Form extends Injectable implements \Countable, \Iterator
...
-		let messages = this->_messages;
-		if fetch elementMessages, messages[name] {
-			return elementMessages;
-		}
-
-		let group = new Group(),
-			this->_messages[name] = group;
-
-		return group;
+        return this->get(name)->getMessages();
```
sergeyklay: Well, in general looks good to me. But it bothers me that previously this method always returned the `Group` but now it can throw Exception. I think this behaviour should be the same now.

sergeyklay (changes requested):
Could you please take a look at comment for `getMessagesFor` and update CHANGELOG.md by adding last issue.

Thanks

phalcon/forms/form.zep
mbrostami: Right, It should be backward compatible, I missed that.

mbrostami:
@sergeyklay Ready.

sergeyklay:
@mbrostami  Could you please rebase onto `3.0.x`

sergeyklay (approved):

mbrostami:
@sergeyklay Done

Jurigag:
Squash please.

sergeyklay:
screen shot 2016-12-20 at 5 03 55 pm

sergeyklay:
Thanks

PR #12490, opened 2016-12-21
Phalcon\Crypt safe base64 padding

tugrul:
Hello!

* Type: new feature
* Link to issue: https://github.com/joaquimserafim/base64-url/blob/master/index.js

Small description of change:

Added autopadding feature for following methods

- `Phalcon\Crypt::encryptBase64`
- `Phalcon\Crypt::decryptBase64`

Equal character is unsafe for url and base64 use for padding this character. base64 able to automatically padded by character length. I inspired this feature from above link.

Thanks,
Tuğrul Topuz

### DIFF
--- a/phalcon/crypt.zep
+++ b/phalcon/crypt.zep
@@ -372,7 +372,7 @@ class Crypt implements CryptInterface
 	public function encryptBase64(string! text, key = null, boolean! safe = false) -> string
 	{
 		if safe == true {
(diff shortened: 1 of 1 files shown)


sergeyklay:
We don't add changes directly to stable branches or `master`. Could you please send this PR to the `3.1.x` branch? Note: `3.1.x` not `3.0.x`

sergeyklay:
Thank you

Issue #12148, opened 2016-08-16
Phalcon\Forms\Element::setLabel('0') render error


pwoszczyk:
Phalcon\Forms\Element::setLabel($label) render incorrect if $label = '0'.

Example:
$field = new Numeric('value');
$field->setLabel('0');
$field->label() // <label for="value">value</label>
- Phalcon version: 2.0.13
- PHP Version: 5.6.11
- Operating System: Windows 7
- Installation type: precompiled dll
- Server: Apache

pwoszczyk:
Same on:
- Phalcon version: 3.0.1
- PHP Version: 7.0.8
- Operating System: Ubuntu 16.04
- Installation type: packagecloud.io
- Server: Apache

gguridi:
I think it's fair to be able to associate '0' as label, in this case it shouldn't validate as empty label.

I tried to use empty label as condition because zephir documentation says that ‘Empty’ means the expression is null, is an empty string or an empty array. But that's not true and '0'also validates as false. So the only way to add '0' as a label is by explicitly check if it's numeric as condition.

Added a pull request with the change.

sergeyklay:
Fixed in the `3.0.x` branch.

PR #12515, opened 2016-12-29
Added a number (0) to be a label in a form field. Null, empty string …

gguridi:
…or empty array will keep invalidating a label and using the value instead.

Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/12148

Small description of change:
Can't associate a 0 with the element label. Tried to use empty label as condition because zephir documentation says that ‘Empty’ means the expression is null, is an empty string or an empty array. But that's not true and '0'also validates as false. So the only way to add '0' as a label is by explicitly check if it's numeric as condition.

Thanks

---
This change is [Reviewable](https://reviewable.io/reviews/phalcon/cphalcon/12515)


Jurigag (commented):
If it's bug fix then it should be pushed to 3.0.x branch. Also update changelog.md

sergeyklay:
@gguridi Could you please provide a test which proves that the problem is solved

gguridi:
@sergeyklay Added a test to check the label being 0.

sergeyklay:
Thanks

PR #12517, opened 2016-12-29
Added DI services loading from php/yaml file for readability

gguridi:
Hello!

* Type: new feature
* Link to issue:

Small description of change:

Dependency injection is the container of any instance declaration in the system. Theoretically any instance should be obtained through dependency injection, which means that in a normal website/application we can end up with lots of classes set in the di as a services. The current system to add services into the dependency injection object require the di itself, setting them one by one.

This pull request is to enable the load of the services from yaml or php files. This is not a big feature and won't solve any real problem, but I just though it would make things more readable as we can have one file with the declaration of the dependencies without having to call di methods, parameters, etc.

This just improves readability, nothing else.

Thanks

---
This change is [Reviewable](https://reviewable.io/reviews/phalcon/cphalcon/12517)

### DIFF
--- a/phalcon/di.zep
+++ b/phalcon/di.zep
@@ -26,6 +26,7 @@ use Phalcon\Di\ServiceInterface;
 use Phalcon\Di\Exception;
 use Phalcon\Events\ManagerInterface;
 use Phalcon\Di\InjectionAwareInterface;
+use Phalcon\Config;
 
 /**
  * Phalcon\Di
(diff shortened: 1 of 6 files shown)


Jurigag:
Update changelog.md please too

Jurigag (changes requested):
Update changelog.md please too

gguridi:
Added changes into the changelog.

phalcon/di.zep
```diff
@@ -429,4 +430,101 @@ class Di implements DiInterface
...
+	 * Loads services from a config object.
+	 */
+	protected function loadFromConfig(<Config> config)
+	{
+		var services, name, service, shared;
+
+		let services = config->toArray();
+		for name, service in services
+		{
+			let shared = isset service["shared"] ? service["shared"] : true;
```
Fenikkusu: You should probably change the default to false. The default for set's $shared variable is false. You are forcing all items to be shared items by default when loaded from the config. This is counter to current setups.

Fenikkusu (changes requested):

sergeyklay (approved):

Jurigag (approved):

Jurigag:
Can you please rebase so it could be merged?

sergeyklay:
Implemented in the `3.2.x` branch. Thank you for contributing.

Issue #12392, opened 2016-11-03
Security tokens error


emiliodeg:
Well I think this is an error

```php
echo $this->security->getSessionToken();
echo $this->security->getToken();
echo $this->security->getSessionToken();
echo $this->security->getToken();

//expeted output
RjlLK2k0bEZSd2hadThscnp3bWMwUT09 
bXdDMmx6cWZyMTRXdWF3SVdXc3VVUT09 
RjlLK2k0bEZSd2hadThscnp3bWMwUT09 
bXdDMmx6cWZyMTRXdWF3SVdXc3VVUT09 

//real output
RjlLK2k0bEZSd2hadThscnp3bWMwUT09 
bXdDMmx6cWZyMTRXdWF3SVdXc3VVUT09 
bXdDMmx6cWZyMTRXdWF3SVdXc3VVUT09 <<--- here getToken() rewrite getSessionToken()
bXdDMmx6cWZyMTRXdWF3SVdXc3VVUT09 

```
getToken function https://github.com/phalcon/cphalcon/blob/master/phalcon/security.zep#L354 
getToken() should not change the value of getSessionToken() until the next request

I think this error generates the misunderstanding that makes programmers confuse us also removes the possibility of using the Validator Identical in our forms assigning the accepted value getSessionToken()

What do you think?

mixinix:
I have same problem with this too. v3.0.1

I'm beginner, CSRF looks like easy in Phalcon, but not easy.

temuri416:
Looks like you've come across the same issue:

https://forum.phalconphp.com/discussion/8093/csrf-problems-on-206#C21945

emiliodeg:
@temuri416 yes you are right, here some other examples
```php
//some action view in volt 
{{ dump(security.checkToken(null, null, false)) }}
{{ security.getSessionToken() }}
{{ security.getToken() }}//here sessionToken was rewrited
{{ security.getSessionToken() }}
{{ security.getToken() }}
{{ dump(security.checkToken(null, null, false)) }}

<form action="" method="post">
    <input type="text" name="{{ security.getTokenKey() }}" value="{{ security.getToken() }}">
    <input type="submit">
</form>

//ouput
boolean true
aVAza2FRMy9ydzVqdGgvdU9GQTMrQT09
YTBuUlRpMGtBTlYwc0ZVeHdJakc5UT09
YTBuUlRpMGtBTlYwc0ZVeHdJakc5UT09
[...]

PR #12518, opened 2016-12-29
Added retainer for current session token in order to not break checkToken functionality if getToken is called before.

gguridi:
Hello!

* Type: bug fix | new feature
* Link to issue: https://github.com/phalcon/cphalcon/issues/12392

Small description of change:

Security::checkToken functionality doesn't use the regenerated session but uses the one that was set at the beginning if it existed. This will prevent that if we call getToken and the session token is regenerated, the checkToken method will fail because the session token has changed.

I've also cast to string the userToken coming from the request::getPost method as it could be null and I was receiving an E_WARNING when passing it to hash_equals.

Thanks

---
This change is [Reviewable](https://reviewable.io/reviews/phalcon/cphalcon/12518)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # [3.0.4](https://github.com/phalcon/cphalcon/releases/tag/v3.0.4) (XXXX-XX-XX)
 - Fixed Isnull check is not correct when the model field defaults to an empty string. [#12507](https://github.com/phalcon/cphalcon/issues/12507)
+- Added a retainer for the current token to be used during the checkings, so when `Phalcon\Security::getToken` is called the token used for checkings don't change. [#12392](https://github.com/phalcon/cphalcon/issues/12392)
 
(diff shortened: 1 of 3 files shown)


Jurigag (changes requested):
Update `CHANGELOG.md` please too

sergeyklay (changes requested):
@gguridi We use the `3.0.x` branch for bug fixes that do not break backward compatibility. I changed the base branch to `3.0.x` from `3.1.x`. But now your PR  has conflicts that must be resolved. So could you please rebase the PR onto `3.0.x` branch?

sergeyklay:
@gguridi Could you please rebase again? This PR has conflicts (CHANGELOG.md) after merging https://github.com/phalcon/cphalcon/pull/12515

niden:
This has been implemented with PR https://github.com/phalcon/cphalcon/pull/13642

Issue #12106, opened 2016-08-07
Validator option to skip all validation if field is not present


codeaid:
## Problem

Currently there is no way to completely skip validation if a field has not been passed as part of an array into the `validate()` method. There was a similar issue submitted for this type of problem (#11597), however, the suggestion to use `allowEmpty` given in it doesn't really solve the problem. 

In my current scenario data is coming from an API endpoint, meaning not all fields are being submitted to PATCH endpoints. Users can submit `{ a: 1 }`, `{ b: 2 }` or `{ a: 1, b: 2 }`, which would all be equally valid in REST world. 

The problem is - if `allowEmpty` is enabled then consumers can still post `false`, `0`, `[]`, `null`, `""` etc. - anything that is an empty value, because the `preCheck()` method [has](https://github.com/phalcon/cphalcon/blob/master/phalcon/validation.zep#L573-L578) the following logic:

``` zephir
if validator->getOption("allowEmpty", false) {
    return empty this->getValue(singleField);
}
```

That means that additional sanitation needs to be performed in the method that eventually receives the validated data to make sure that invalid empty values are not present, simply because the data that results from the validator could look something like this (if `a` and `b` were posted as `false`):

``` php
[
    "a": false,
    "b": false,
]
```

In other words, it doesn't matter what other validators are attached to those fields, they'll happily "validate" if they're empty in any shape or form.
## Suggested solution

[Apigility](https://apigility.org/) solves this issue by having three options for each field:
- Required
- Allow empty
- Continue if empty

The first one is essentially the same as Phalcon's `PresenceOf` validator, and Apigility actually internally creates a very similar `NotEmpty` validator if `required` is set to `true`.

Allow empty is, yet again, an option that we luckily already have in Phalcon - it allows empty values to be passed in without performing any validation.

What we're missing though is implementation of the latter "continue if empty" option, which in case of Apigility completely ignores any validation on the field if the option is disabled and the value being validated is empty. In their [own words](https://apigility.org/documentation/content-validation/validating):

> If you want validators to run even if the data is not present or is empty, toggle the "Continue if Empty" setting.

My suggestion is to implement a similar option in Phalcon too, which would allow users of validator classes to skip validation on fields completely if they are not present in the data, which currently isn't possible without using the `allowEmpty` option (which in its turn allows invalid (empty) values to be accepted).

Rather than doing the following as a workaround:

``` php
$validator->add('a', new Validator\Between([
    "minimum"    => 1,
[...]

PR #12519, opened 2016-12-30
allowEmpty can be populated with the list of what we consider 'empty'. The validation only will stop if the empty validation returns true.

gguridi:
Hello!

* Type: bug fix | new feature
* Link to issue: https://github.com/phalcon/cphalcon/issues/12106

Small description of change:

Added the possibility to specify in allowEmpty option what we consider 'empty'. Fixed the behaviour so only in case it returns true we move to the next validation. Added recursivity in the preChecking method so we don't duplicate code. Updated changelog.

- allowEmtpy = true (this means that everything that evaluates to empty is ok)
- allowEmpty = ['', null] (this means that empty only means empty string and null, false and 0 won't evaluate as empty).

I've added a pull request with this change. The benefits are:
- Backwards compatible.
- Flexible, we can specify 'unknown', 'empty' or -1 as empty values in case we are using a shitty api.

An improvement also added in the pull request is that only if the allowEmpty validation is true the validation will stop there, in case of false it will continue with the next validators.

Thanks

---
This change is [Reviewable](https://reviewable.io/reviews/phalcon/cphalcon/12519)


sergeyklay:
Thanks

someson:
what if
```php
    // ...
    allowEmpty => ['', null, new RawValue('NULL')],
```
?

`new RawValue('NULL') ===  new RawValue('NULL')` → false

sergeyklay:
@someson could you please create a PR with test which fails?

someson:
@sergeyklay, yeah, I'll try :)

PR #12523, opened 2017-01-02
Added smallint support in Phalcon\Db

Studentsov:
This patch adds the capability to use the data type smallint, which is part of the SQL standard and is well supported by major DBMS. This data type uses 2 bytes instead of 4 bytes for type int, which will save memory if you don't need large numeric values.

### DIFF
--- a/phalcon/db/column.zep
+++ b/phalcon/db/column.zep
@@ -142,6 +142,11 @@ class Column implements ColumnInterface
 	const TYPE_TIMESTAMP = 17;
 
 	/**
+	 *  Small integer abstract type
+	 */
+	const TYPE_SMALLINTEGER = 18;
+
+	/**
 	 * Bind Type Null
 	 */
 	const BIND_PARAM_NULL = 0;
@@ -337,6 +342,7 @@ class Column implements ColumnInterface
 				case self::TYPE_DECIMAL:
 				case self::TYPE_DOUBLE:
 				case self::TYPE_BIGINTEGER:
+				case self::TYPE_SMALLINTEGER:
 					let this->_scale = scale;
 					break;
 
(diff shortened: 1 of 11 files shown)


phalcon/db/column.zep
```diff
@@ -142,6 +142,11 @@ class Column implements ColumnInterface
 	const TYPE_TIMESTAMP = 17;
 
 	/**
+	 *  Small integer abstract type
+	 */
+	const TYPE_SMALLINTEGER = 18;
```
sergeyklay: How about just `TYPE_SMALLINT`?

sergeyklay (changes requested):
@Studentsov Could you please cover new feature by unit tests?

phalcon/db/column.zep
Studentsov: We already have a type TYPE_BIGINTEGER. It would be very strange (though quite in the spirit of PHP 😆) to use the name TYPE_SMALLINT.

sergeyklay: agree

sergeyklay (approved):

Jurigag:
Can you add here tinyint support too?

Studentsov:
As far is I known, the one-byte integer type isn't provided by the SQL
standard. For example, it isn't in PostgreSQL.

Jurigag:
Oh okay, so it could be at least added to mysql. Not sure about Column class.

niden:
@Studentsov Could you be so kind as to rebase this for the `3.4.x` branch so that we can merge this?

Apologies for letting this linger for this long

niden:
Addressed in https://github.com/phalcon/cphalcon/pull/13562

PR #12547, opened 2017-01-13
Added way to get active key and active function from Phalcon\Acl\Adapter\Memory

Jurigag:
Hello!

* Type: new feature

Small description of change: added methods for getting active key and active function for latest acquired access in acl.

Thanks


Jurigag:
@sergeyklay could this be merged? Need this for some other PR.

sergeyklay:
This will work only for Memory Adapter right? So typical user, who does not read CHANGELOG.md,  will be very surprised why feature which will use these methods does not work with MyAwesomeAclAdapterFromIncubator. I'm not sure that some custom adapter (impements adapter pattern) should contain methods except those which are declared in the interface.

Jurigag:
Yea only for memory. Well then just incubator would need to add this too. This is why i added this to Memory class, not Adapter. I can do the same pr to incubator classes.

michanismus:
@Jurigag When it is only for Memory, why don't you extend the Memory adapter for your special purpose?

Jurigag:
@michanismus the thing is i needed this functionality for some other PR but i will postpone it to phalcon 4

sergeyklay:

niden:
This has been merged with cherry-pick

https://github.com/phalcon/cphalcon/commit/80ced232ff758fc92904679fff82f2349891e52f

@Jurigag thanks!

PR #12551, opened 2017-01-16
Allow other except fields than unique fields

Jurigag:
Hello!

* Type: bug fix
* Link to issue:

Small description of change: allow having other except fields than fields used for Uniqueness. Previously you could use only field for which uniqueness check was done in except option.

Thanks


sergeyklay:
Thanks

PR #12570, opened 2017-01-24
Fixes Binder internal cache

Jurigag:
Hello!

* Type: bug fix

Small description of change: info about classes for which where binding where done was created only when cache for binder was enabled which was causing internal cache not working properly. This is fixing this problem by removing not needed condition.

Thanks


sergeyklay:
Thanks

PR #12574, opened 2017-01-25
Update Phalcon\Mvc\Model return ResultsetInterface

ibpavlov:
Hello!

* Type: bug fix

ResultsetInterface is not imported and causes problems when generating Stubs for IDE. It is generating `@return` type of `Phalcon\Mvc\ResultsetInterface` and not `Phalcon\Mvc\Model\ResultsetInterface`

I think it should be imported in order to easily find the correct class.

Let me know if I am missing something

Thanks


sergeyklay:
Thanks

Issue #12613, opened 2017-02-10
\Phalcon\Config::offsetUnset property not unset


cottton:
`Phalcon\Config::offsetUnset` does not unset the propery - it sets it to `null`.
I saw here https://github.com/phalcon/cphalcon/blob/master/phalcon/config.zep#L147
that the unset has commented out. Was there a reason for that?

Problem: comparing 2 configs `::toArray` are not the same (just found that out while writing a unit test).

**Test:**

    $config = new \Phalcon\Config(['foo' => 'bar', 'bar' => 'baz']);
    $config->offsetSet('baz', 1);
    unset($config['baz']);
    $config->offsetUnset('foo');
    var_export($config->toArray());

**Expected:**

    array (
        'bar' => 'baz',
    ) 

**Actual:**

    array (
        'foo' => NULL,
        'bar' => 'baz',
        'baz' => NULL,
    )
 
_I know `isset(null)` it false but having a container that holds data i would expect that unset actually unsets the property._ 

### Details

Web framework delivered as a C-extension for PHP
phalcon => enabled
Author => Phalcon Team and contributors
Version => 3.0.1
Build Date => Oct  4 2016 18:23:02
Powered by Zephir => Version 0.9.4a-dev-7e304ba18c

PHP 7.0.11

Jurigag:
As i remember it's not possible to unset property from objects in zephir for some reason.

cottton:
I see. 
If anybody has the same problem then just overwrite `Phalcon\Config` with

    public function offsetUnset($index)
    {
        parent::offsetUnset($index);
        unset($this->{$index});
    }

Btw - is it possible to add that fix directly into phalcon and let PHP do the work?

Jurigag:
What you mean directly into phalcon? See above, it's not possible.

cottton:
I mean a `Config.php` in `scripts/Phalcon/` that overwrites the method.

Jurigag:
What you mean? It can't be in repository. Just add such an class to your app if you need fix.

PR #12615, opened 2017-02-11
refactor the phalcon/config.zep for issue #12613

qzfzz:
Hello!

* Type: code quality
* Link to issue:  https://github.com/phalcon/cphalcon/issues/12613
 #12613	

Small description of change:
The phalcon/config.zep has not fully implemented the ArrayAccess interface right now, so I refactor the phalcon/config.zep for fully implementing the interface!

Thanks

### DIFF
--- a/phalcon/config.zep
+++ b/phalcon/config.zep
@@ -106,11 +106,15 @@ class Config implements \ArrayAccess, \Countable
 	 * );
 	 *</code>
 	 */
-	public function offsetGet(var index) -> string
+	public function offsetGet(var index)
 	{
 		let index = strval(index);
 
-        return this->arrayConfig[index];
+        if isset this->arrayConfig[index]{
+		    return this->arrayConfig[index];
+		}
+		
+        return null;
 	}
 
 	/**
--- a/phalcon/config.zep
+++ b/phalcon/config.zep
@@ -49,7 +49,7 @@ use Phalcon\Config\Exception;
  */
 class Config implements \ArrayAccess, \Countable
 {
-	protected arrayConfig = [];
+	protected data = [];
 	
 	/**
 	 * Phalcon\Config constructor
@@ -59,7 +59,7 @@ class Config implements \ArrayAccess, \Countable
 		var key, value;
 
 		for key, value in arrayConfig {
-			let this->arrayConfig[key] = value;
+			this->offsetSet(key, value);
 		}
 	}
 
@@ -75,7 +75,8 @@ class Config implements \ArrayAccess, \Countable
 	public function offsetExists(var index) -> boolean
 	{
 		let index = strval(index);
-		return this->arrayConfig[index];
+		
+		return isset this->data[index];
 	}
 
 	/**
@@ -90,8 +91,8 @@ class Config implements \ArrayAccess, \Countable
 	{
 		let index = strval(index);
 
-        if isset this->arrayConfig[index]{
-        	    return this->arrayConfig[index];
+        if isset this->data[index]{
+        	    return this->data[index];
         }
         
 		return defaultValue;
@@ -110,8 +111,8 @@ class Config implements \ArrayAccess, \Countable
 	{
 		let index = strval(index);
 
-        if isset this->arrayConfig[index]{
-		    return this->arrayConfig[index];
+        if isset this->data[index]{
+		    return this->data[index];
 		}
 		
         return null;
@@ -131,9 +132,9 @@ class Config implements \ArrayAccess, \Countable
 		let index = strval(index);
 
 		if typeof value === "array" {
-			let this->arrayConfig[index] = value;
+		    let this->data[index]  = new self(value);
 		} else {
-			let this->arrayConfig[index] = value;
+			let this->data[index] = value;
 		}
 	}
 
@@ -148,7 +149,7 @@ class Config implements \ArrayAccess, \Countable
 	{
 		let index = strval(index);
 
-		unset this->arrayConfig[index];
+		unset this->data[index];
 	}
 
 	/**
@@ -185,7 +186,7 @@ class Config implements \ArrayAccess, \Countable
 		var key, value, arrayConfig;
 
 		let arrayConfig = [];
-		for key, value in this->arrayConfig {
+		for key, value in this->data {
 			if typeof value === "object" {
 				if method_exists(value, "toArray") {
 					let arrayConfig[key] = value->toArray();
@@ -214,7 +215,7 @@ class Config implements \ArrayAccess, \Countable
 	 */
 	public function count() -> int
 	{
-		return count( this->arrayConfig );
+		return count( this->data );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


Jurigag:
Update changelog, also we don't accept PRs to master branch.

phalcon/config.zep
```diff
@@ -58,7 +59,7 @@ class Config implements \ArrayAccess, \Countable
 		var key, value;
 
 		for key, value in arrayConfig {
-			this->offsetSet(key, value);
```
cottton: would suggest to leave `this->offsetSet(key, value);`

phalcon/config.zep
```diff
@@ -127,9 +131,9 @@ class Config implements \ArrayAccess, \Countable
 		let index = strval(index);
 
 		if typeof value === "array" {
-			let this->{index} = new self(value);
+			let this->arrayConfig[index] = value;
```
cottton: should be `let this->arrayConfig[index] = new self(value);`

phalcon/config.zep
```diff
@@ -49,7 +49,8 @@ use Phalcon\Config\Exception;
  */
 class Config implements \ArrayAccess, \Countable
 {
-
+	protected arrayConfig = [];
```
cottton: suggest protected data = [];

phalcon/config.zep
```diff
@@ -74,8 +75,7 @@ class Config implements \ArrayAccess, \Countable
 	public function offsetExists(var index) -> boolean
 	{
 		let index = strval(index);
-
-		return isset this->{index};
+		return this->arrayConfig[index];
```
cottton: return **isset** this->arrayConfig[index];

qzfzz:
Hello! Thank you for your patience. This is the first time I contributing source code for phalcon, and I have correct the bugs for refactor the phalcon/config.zep.

Jurigag:
Good but we don't accept PRs to master branch. Master is for latest stable version. Send your pr to 3.0.x and update changelog.

qzfzz:
Thank you for your attention, I will create new pr to 3.0.x soon!


原始邮件
发件人:Wojciech Ślawskinotifications@github.com
收件人:phalcon/cphalconcphalcon@noreply.github.com
抄送:qzfzzqzfzz@163.com; Authorauthor@noreply.github.com
发送时间:2017年2月11日(周六) 22:07
主题:Re: [phalcon/cphalcon] refactor the phalcon/config.zep for issue#12613 (#12615)


Good but we don't accept PRs to master branch. Master is for latest stable version. Send your pr to 3.0.x and update changelog.

cottton:
Would suggest another change btw.
Code: https://github.com/phalcon/cphalcon/blob/3.0.x/phalcon/config.zep#L188

     // updated: value instanceof Config &&
    if  value instanceof Config && method_exists(value, "toArray") {
        let arrayConfig[key] = value->toArray();
    } else {
        let arrayConfig[key] = value;
    }

**Reason**: should only convert to array if is instance of Config.
F.e. we got a class Car and Car has the method `::toArray`
then the current code would convert all the Car objects to array.
But user expected the **value** (object) car he previously set.

Jurigag:
Then just instanceof check will be enough.

cottton:
Indeed.

cottton:
Another problem may exists at https://github.com/phalcon/cphalcon/blob/3.0.x/phalcon/config.zep#L93
If value of `this->{index}` is `null` then `defaultValue` will be returned.
User may expected `null` (offset actually exists).

Dunno how to check if property exists in zephir but:

    public function get(var index, var defaultValue = null) -> var
	{
        let index = strval(index);

        if property_exists(this, index) // property_exists ?
            return this->{index}
		}

		return defaultValue;
	}

---------------------

The same problem at `::offsetExists` https://github.com/phalcon/cphalcon/blob/3.0.x/phalcon/config.zep#L74

If offset `index` exists and is `null` it should return true.
Otherwise the method would have be named `offsetIsset`

qzfzz:
Thank you so much for your advise.


Have a nice day!




原始邮件
发件人:cotttonnotifications@github.com
收件人:phalcon/cphalconcphalcon@noreply.github.com
抄送:qzfzzqzfzz@163.com; Authorauthor@noreply.github.com
发送时间:2017年2月13日(周一) 22:12
主题:Re: [phalcon/cphalcon] refactor the phalcon/config.zep for issue#12613 (#12615)


Another problem may exists at https://github.com/phalcon/cphalcon/blob/3.0.x/phalcon/config.zep#L93
 If value of this-{index} is null then defaultValue will be returned.
 User may expected null (offset actually exists).
Dunno how to check if property exists in zephir but:
public function get(var index, var defaultValue = null) - var { let index = strval(index); if property_exists(this, index) // property_exists ? return this-{index} } return defaultValue; }
The same problem at ::offsetExists https://github.com/phalcon/cphalcon/blob/3.0.x/phalcon/config.zep#L74
If offset index exists and is null it should return true.
 Otherwise the method would have be named offsetIsset

qzfzz:
Thank you so much, and I have fixed the bugs!


原始邮件
发件人:cotttonnotifications@github.com
收件人:phalcon/cphalconcphalcon@noreply.github.com
抄送:qzfzzqzfzz@163.com; Authorauthor@noreply.github.com
发送时间:2017年2月13日(周一) 22:12
主题:Re: [phalcon/cphalcon] refactor the phalcon/config.zep for issue#12613 (#12615)


Another problem may exists at https://github.com/phalcon/cphalcon/blob/3.0.x/phalcon/config.zep#L93
 If value of this-{index} is null then defaultValue will be returned.
 User may expected null (offset actually exists).
Dunno how to check if property exists in zephir but:
public function get(var index, var defaultValue = null) - var { let index = strval(index); if property_exists(this, index) // property_exists ? return this-{index} } return defaultValue; }
The same problem at ::offsetExists https://github.com/phalcon/cphalcon/blob/3.0.x/phalcon/config.zep#L74
If offset index exists and is null it should return true.
 Otherwise the method would have be named offsetIsset

cottton:
Question: 
`\Phalcon\Builder\Options` extends `Phalcon\Config`.
See https://github.com/phalcon/phalcon-devtools/blob/3.0.x/scripts/Phalcon/Builder/Options.php

If in class `Options` (or any other extending class) a value is set like `$this->{$key}` then the setter [(offsetSet)](https://github.com/qzfzz/cphalcon/blob/7d4c5cf0cee8eab7c0877b044e114712e9e68474/phalcon/config.zep#L130) gets called, right?

Jurigag:
No. offsetSet is method from ArrayAccess interface. It's making possible to do something like $this[$key] = 'xyz' and it will set property called $key with value 'xyz'.

http://php.net/manual/en/class.arrayaccess.php

cottton:
Ok. In the current fix we use `protected $data` that hold the data.
In PHP if we extend `Phalcon\Config` and use `$his->{$key} = $value;` - do we have an conflict?

Jurigag:
If someone will use `$this['data'] = $value` then yes.

Jurigag:
Well tbh i don't see offsetSet changed method here. Why is that?

Jurigag:
Oh i guess code is changed so there will be no conflict.

Jurigag:
Can you close this since it's to master repo?

PR #12629, opened 2017-02-16
PostgreSQL. Duplicate table: 7 ERROR: relation" PRIMARY "already exists"

jumper423:
CONSTRAINT must be unique

Hello!

* Type: bug fix
* Link to issue:

- [ ] I have read and understood the [Contributing Guidelines](https://github.com/phalcon/cphalcon/blob/master/CONTRIBUTING.md)?
- [ ] I have checked that another pull request for this purpose does not exist.

Small description of change:

Postgresql in CONSTRAINT must specify a unique name for this CONSTRAINT, otherwise it leads to mistakes.
> "ERROR: SQLSTATE [42P07]: Duplicate table: 7 ERROR: relation" PRIMARY "already exists"

Thanks

### DIFF
--- a/phalcon/db/dialect/postgresql.zep
+++ b/phalcon/db/dialect/postgresql.zep
@@ -287,7 +287,7 @@ class Postgresql extends Dialect
 	 */
 	public function addPrimaryKey(string! tableName, string! schemaName, <IndexInterface> index) -> string
 	{
-		return "ALTER TABLE " . this->prepareTable(tableName, schemaName) . " ADD CONSTRAINT \"PRIMARY\" PRIMARY KEY (" . this->getColumnList(index->getColumns()) . ")";
+		return "ALTER TABLE " . this->prepareTable(tableName, schemaName) . " ADD CONSTRAINT \"" . tableName . "_PRIMARY\" PRIMARY KEY (" . this->getColumnList(index->getColumns()) . ")";
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


jumper423:
It is possible to completely remove the use CONSTRAINT, I do not see it necessary.

sergeyklay:
> CONSTRAINT must be unique

So now do you provided this opportunity?

jumper423:
Yes. Now CONSTRAINT name part of the table name. The need to add more and field names, I doubt.

sergeyklay:
In general looks weird to me.

For example, when I call `dropPrimaryKey` I want to drop real PR, not `table_PK`, isn't it? Or for example, I want to call `addPrimaryKey` twice and the problem with the duplicates still present. So I don't understand what problem you are trying to solve.

Could you please elaborate on that?

jumper423:
Here is an example.
There are two migration to create tables. Each table must create a primary key.
But this can not be done. Because the error occurs.
"ERROR: SQLSTATE [42P07]: Duplicate table 7 ERROR: relation" core "already exists"
https://gist.github.com/jumper423/087d0abce5659aec22f1037ad08a6651

sergeyklay:
@jumper423 

Also please update `Phalcon\Db\Dialect\Postgresql::createTable` and `Phalcon\Db\Dialect\Postgresql::addIndex` by implement new logic.

And update CHANGELOG.md.

nsossonko:
This change is backwards incompatible it would seem and can cause issues for users currently using postgres since it will assume a changed primary key name that has not been that way in the past...I guess maybe there should be some sort of migration system (optional) to the new naming?

jumper423:
You can make imports. Ask for the name of the primary keys. Change them to a new standard. Do you think a good idea?

jumper423:
And if so, then where to describe the migration itself? In fact, you can try to make one SQL request.

sergeyklay:
We'll try to sort out with this in 4.x

niden:
This has been addressed with https://github.com/phalcon/cphalcon/pull/13643

Thank you @jumper423 and @CameronHall

PR #12631, opened 2017-02-17
Install file refactoring, custom phpize & php-config paths

DimazzzZ:
Hello. 

I've made small refactoring for `install` file. Previously it supported only `$PATH`'s binaries for `phpize` and `php-config`, but now users are able to set custom paths for them. It is very useful if you have more than one PHP instance and you want to compile Phalcon for each of them.

For example I've three different PHP versions in my Ubuntu and default one is 7.1. To compile Phalcon for 5.6 instance:

`./install --phpize /usr/bin/phpize5.6 --php-config /usr/bin/php-config5.6`

To save backward compatibility I also added an `--arch` param (for 32bits & 64bits dirs).

### DIFF
--- a/build/install
+++ b/build/install
@@ -13,15 +13,47 @@
 #
 #  Authors: Andres Gutierrez <andres@phalconphp.com>
 #            Eduar Carvajal <eduar@phalconphp.com>
+#
+#  Available params:
+#  --arch
+#  --phpize
+#  --php-config
+#
+#  Example:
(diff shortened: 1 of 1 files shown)


Izopi4a:
i was lazy to read how to do implement that.

Very good job!

sergeyklay:
@DimazzzZ Could you please send this to the 3.0.x ? I'll merge this to the 3.1.x later

sergeyklay:
Thank you

Issue #11934, opened 2016-06-28
APCu classes


Jurigag:
In php7 in apcu there is no `apc_*` methods emulation. Phalcon has classes using only `apc_*`, not apcu. On some point if php 7 support in 2.1.x or later versions like 2.2.x or something will be better/stable we should add classes using APCu, so people don't need to install apcu_bc(which adds polyfill for apc_\* methods on php7).

galki:
+1

Jurigag:
You sure it's loaded ? Check phpinfo.

galki:
That solved it, thank you. I was loading it in the wrong order.

stevendesu:
+1 for APCu support

I'll be honest: I haven't fully done my research on this one. I did a bit of Googling for "PHP in-memory caches" and saw that my options were memcached and APCu, and reading up on the two it sounds like memcached is better for distributed environments or setups where more than PHP needs to read the cache, but if it's just PHP one a single server APCu is supposedly more performant. Based only on that minutia of knowledge I Google'd "Phalcon APCu support" and found this thread

I'll try installing APCu + APCu_bc and if it works as described then that will solve all of my issues for now (I have root access and can install whatever I need), but I know some people may not have the ability to install arbitrary extensions. Not to mention more extensions = more bloat, more memory usage, and longer load times.

sergeyklay:
See https://pear.php.net/bugs/bug.php?id=21007

Jurigag:
@stevendesu 

Extensions are loaded only on startup, apcu_bc is like a few kb of memory usage, and it has almost none bloat beacause it's just calling apcu functions.

sergeyklay:
Fixed in the `3.2.x` branch.

PR #12664, opened 2017-02-27
Improved Phalcon\Cache\Backend\Apc

sergeyklay:
Hello!

* Type: new feature
* Link to issue: #12098, #11934

Small description of change:

- Improved `Phalcon\Cache\Backend\Apc` by introducing support of APCu
- Improved `Phalcon\Annotations\Adapter\Apc` by introducing support of APCu

Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Added `prepend` parameter to `Phalcon\Loader::register` to specify autoloader's loading order to top most
 - Fixed `Phalcon\Session\Bag::remove` to initialize the bag before removing a value [#12647](https://github.com/phalcon/cphalcon/pull/12647)
 - Fixed `Phalcon\Mvc\Model::getChangedFields` to correct detect changes from NULL to Zero [#12628](https://github.com/phalcon/cphalcon/issues/12628)
+- Improved `Phalcon\Cache\Backend\Apc` by introducing support of APCu [#12098](https://github.com/phalcon/cphalcon/issues/12098), [#11934](https://github.com/phalcon/cphalcon/issues/11934)
 
 # [3.0.4](https://github.com/phalcon/cphalcon/releases/tag/v3.0.4) (2017-02-20)
 - Fixed Isnull check is not correct when the model field defaults to an empty string. [#12507](https://github.com/phalcon/cphalcon/issues/12507)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Fixed `Phalcon\Session\Bag::remove` to initialize the bag before removing a value [#12647](https://github.com/phalcon/cphalcon/pull/12647)
 - Fixed `Phalcon\Mvc\Model::getChangedFields` to correct detect changes from NULL to Zero [#12628](https://github.com/phalcon/cphalcon/issues/12628)
 - Improved `Phalcon\Cache\Backend\Apc` by introducing support of APCu [#12098](https://github.com/phalcon/cphalcon/issues/12098), [#11934](https://github.com/phalcon/cphalcon/issues/11934)
+- Improved `Phalcon\Annotations\Adapter\Apc` by introducing support of APCu [#12098](https://github.com/phalcon/cphalcon/issues/12098)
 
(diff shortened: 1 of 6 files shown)


Jurigag:
I don't like this. Why just not introduce new classes?

sergeyklay:
Why don't like?

Jurigag:
Cuz as i wrote - i would like much more just seperated classes. I don't like one class to be adapter for both apc and apcu.

sergeyklay:
Yes, I'm understand. But could you please elaborate on that. In order to discuss something, we need a reason. "I don't like" not reason.

Jurigag:
Well we will remove necessary to have checks, conditions etc, also code will be much cleaner. As i saw in other frameworks if they had support for both apc and apcu - it was done in seperated classes.

andresgutierrez:
We can have this for BC reasons and fully deprecate apc in Phalcon 4, what do you think?

Jurigag:
Not deprecate - just remove apc more likely, but then we would need imho change apc classes names to apcu cuz it will no longer work with apc obviously. That's why seperated classes seems more suitable - someone could just switch to apcu and he would need to do less stuff in phalcon 4.

Jurigag:
Also notice that atm if someone is already using php7 + apcu + phallcon he needs to install apcu_bc. So this classes work for him anyway and he will not notice any difference, except those additional checks.

And if someone would like to use apcu + php7 + phalcon then he would need to change to new class in phalcon 4, cuz there is no point having it named as Apc when it will be working only with Apcu.

dschissler:
I'd prefer if they were to be separate adapters but if this new code can properly check for all of the different possibilities (2^2 I think) then this should be good.  One thing that could be a complication is if someone is having issues and it is unclear which extension they are using.  Its not a really big issue but if something goes wrong it could make for some unhelpful bug reports.  So perhaps a solution to this would be to have the adapter save which extension it is using so that it can be logged on request.

dschissler:
I think that it might be possible to have both `apc` and `apcu` installed on a supported PHP 5.5 or 5.6 system.  If this is the case then someone should at least manually test all of these combinations (or write a test if that is even possible).

Jurigag:
Php apcu already have apc emulation builtin so apc_fetch function will just call apcu function for php 5 users this change will be totally useless

Jurigag:
How this combined have advantage of being compatible in the next years like if someone who already wants to use apcu is already using old class and in phalcon 4 he would need needs to change name anyway? Just separated classes are way to go, new developers could use then new classes and have less things to care about when switch to phalcon 4, old would need to change anyway.

dschissler:
> Php5 apcu already have apc emulation builtin so apc_fetch function will just call apcu function for php 5 users this change will be totally useless

It seems that it uses `apcu` before `apc` if its available.

> How this combined have advantage of being compatible in the next years like if someone who already wants to use apcu is already using old class and in phalcon 4 he would need needs to change name anyway? Just separated classes are way to go, new developers could use then new classes and have less things to care about when switch to phalcon 4, old would need to change anyway.

Couldn't we just keep the same name?  Do the class names really need to be changed from `...apc` to `...apcu`?  I don't think that there needs to be an exact mapping of extension name to class name.

Jurigag:
I think yes - it needs to change, cuz it no longer will work in phalcon 4 with APC, just only with APCU, so why keep old APC name?

dschissler:
@sergeyklay It seems that the argument is coming down to whether or not the Phalcon 4 class name can be left the same or if it will need to be changed from `...Apc` to  `...Apcu`.  I think that its ok not having the "u" in it and that it doesn't cause any confusion for Phalcon 4.  However, if this is an undesirable future situation then probably there shouldn't be a single Apc adapter.

Jurigag:
I just don't see really why we can't have both adapters. In all known major frameworks they have apcu classes.

niden:
I think we should have 2 adapters.

We can deprecate `Apc` at a later stage without having to touch the adapter to remove the apc related code from a common adapter.

My 2 cents

dschissler:
I think that two adapters has overall less issues than a combined adapter.  But I'm not going to be the one writing the code and I'm also fine with the one adapter.

sergeyklay:
https://github.com/phalcon/cphalcon/pull/12788, https://github.com/phalcon/cphalcon/pull/12786

PR #12667, opened 2017-02-28
Need to check that response has not been already sent in phalcon/mvc/micro.zep

Fabsolute:
Hello!

* Type: bug fix

**In raising this pull request, I confirm the following :**

Small description of change:
It is not possible to return a string value from a mapped function for an HTTP Method, and then catch that returned value by using Micro::getReturnedValue() method inside handler on Micro::after($handler), its overriding my response.
Thanks


sergeyklay:

PR #12684, opened 2017-03-07
PHP warning fix for the autoLoad() function

satakecode:
Hello!

* Type: bug fix
* Link to issue:

This small alteration should fix this situation: https://forum.phalconphp.com/discussion/10339/strange-warning-probably-from-zephirfaststrreplace-or-something-

If you have two classes:
* `My\Class\Foo\Bar`
and
* `My\Class\Foo`

When initializing `Foo()` the autoLoad will come to this line 
```zep
let fileName = substr(className, strlen(nsPrefix . ns));
```
at the moment where `className = "My\Class\Foo"` and `nsPrefix = "My\Class\Foo\Bar"`

At this point (could be a Zephir bug) `fileName` is something that will make 
```zep
let fileName = str_replace(ns, ds, fileName);
```
 throw the following warning (Invalid arguments supplied for `str_replace()`). Either `substr` is returning something other than an empy string or `str_replace` is not allowing an empy string either way testing if `fileName` has any value before 
```zep
let fileName = str_replace(ns, ds, fileName);
```
should mitigate this warning.

Thanks

### DIFF
--- a/phalcon/loader.zep
+++ b/phalcon/loader.zep
@@ -337,11 +337,12 @@ class Loader implements EventsAwareInterface
 			 * Append the namespace separator to the prefix
 			 */
 			let fileName = substr(className, strlen(nsPrefix . ns));
-			let fileName = str_replace(ns, ds, fileName);
 
 			if !fileName {
 				continue;
 			}
+			
+			let fileName = str_replace(ns, ds, fileName);
 
 			for directory in directories {
 				/**


Jurigag:
We don't accept PR to master. Please send it to 3.1.x. Also provide some test which was causing this warning to be produced so we know this is fixed as well update changelog.

sergeyklay:
Fixed in the `3.1.x` branch.

PR #12686, opened 2017-03-08
Dialect Class outside of Zephir won't load because of missing CamelCase

lubberscorrado:
Fixes https://github.com/phalcon/incubator/issues/663

The case-insensitive Class-Instantiation does only work when the Class is also a Zephir one. When it's a PHP Class (like Oracle Dialect now in Incubator) it won't be found if not correctly CamelCased


sergeyklay:

Jurigag:
Also update changelog.

PR #12698, opened 2017-03-09
Fix uniqueness collection when persistent

Jurigag:
Hello!

* Type: bug fix
* Link to issue:

- [x] I wrote some tests for this PR. - Probably exisiting one will fail and i will need to fix them to correct one

Small description of change: When collection was persistent there were wrongly condition for it created. This is fixing it.

Thanks


sergeyklay:
Thanks

PR #12724, opened 2017-03-20
Different 404 pages for multi-module

MoneyCoder:
This problem is not new, but still relevant!

https://forum.phalconphp.com/discussion/397/intercepting-not-found-errors-in-multi-module-applications

Please correct this, because people can not assign different 404 pages for each individual module. As I understand it, the problem is that the modules load is terminated if an EXCEPTION_HANDLER_NOT_FOUND exception or EXCEPTION_ACTION_NOT_FOUND is thrown. Accordingly, the loading of the "dispatcher" dependency within the module is not performed and redirection to the necessary controller and action also.

### DIFF
--- a/BACKERS.md
+++ b/BACKERS.md
@@ -78,3 +78,6 @@ No pledges yet.
 - Semen Pinigin
 - Daif Alotaibi ([@dalf](https://github.com/daif))
 - Baruch Spinoza
+
+# Core Team
+Andres, Serghei, Eduar, Nikolay, Nikolaos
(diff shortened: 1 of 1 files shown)


sergeyklay:
We only accept **bug reports**, **new feature requests** and **pull requests** in GitHub. For questions regarding the usage of the framework, support requests, and questions like this please visit the [official forums](https://forum.phalconphp.com).

I'm closing this PR due to fact that this PR does not fix anything.

PR #12726, opened 2017-03-20
fixed bug for phalcon/mvc/collection and phalcon/image/adapter/imagick

youyoubao:
Hello!

* Type: bug fix 

Small description of change:
 fixed bug for phalcon/mvc/collection and phalcon/image/adapter/imagick

Thanks

### DIFF
--- a/phalcon/image/adapter/imagick.zep
+++ b/phalcon/image/adapter/imagick.zep
@@ -360,7 +360,7 @@ class Imagick extends Adapter
 
 		// Imagick >= 2.0.0
 		if likely method_exists(watermark, "getVersion") {
-			let version = \Imagick::getVersion();
+			let version = watermark->getVersion();
 
 			if version["versionNumber"] >= 0x700 {
 				let method = "setImageAlpha";
--- a/phalcon/mvc/collection.zep
+++ b/phalcon/mvc/collection.zep
@@ -414,7 +414,6 @@ abstract class Collection implements EntityInterface, CollectionInterface, Injec
 			/**
 			 * Requesting a single result
 			 */
-			documentsCursor->rewind();
 
 			let document = documentsCursor->current();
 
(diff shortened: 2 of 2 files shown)


Jurigag:
We don't accept PRs to master repo, please send it to 3.1.x, also update changelog and provide some test that issue is fixed.

sergeyklay:
@dormscript It's technically not a bug. Actually `Imagick::getVersion` is static: https://github.com/mkoppanen/imagick/search?l=PHP&q=getVersion

Nevertheless, we can do it the way you suggest. In any case, as noted @Jurigag , we don't add changes directly to stable branches or `master`. Could you please send this PR to the `3.1.x` branch?

Also I would like to have 2 separate PR for this:  Imagick and Collection

Jurigag:
As you see testa are failing, are you sure you wanted to do this change for collection? Keep in mind that existing collection in phalcon is only for using with php 5, for php7 there are classes in incubator https://github.com/phalcon/incubator maybe there you mean this change?

dormscript:
@sergeyklay Thanks for you attention！
About Imagick extends，in my systems , the method "Imagick::getVersion()" show the error "PHP Strict Standards:  Non-static method Imagick::getversion() should not be called statically". As you say，“Imagick::getVersion is static”，you are right , buy don't work in my system.  I'll find ou the error in my system.

dormscript:
I'll do 2 separate PR for this: Imagick and Collection,thanks

dormscript:
@Jurigag @sergeyklay   
 Current phalcon/mvc/collect  is only adapt to the php extend 'mongo', not adapt 'mongodb' ?

Jurigag:
@dormscript Yes, for mongodb you must use classes from incubator.

dormscript:
Thanks ,I'll try

PR #12733, opened 2017-03-22
beforeException on Application getModule

Surt:
For multi-module applications it is possible to attach a beforeException event to the Application.
The default behaviour is the same (throw exception) if not event listener is set.

Hello!

* Type: new feature
* Link to issue:

Small description of change:

Thanks

### DIFF
--- a/phalcon/application.zep
+++ b/phalcon/application.zep
@@ -120,11 +120,28 @@ abstract class Application extends Injectable implements EventsAwareInterface
 		var module;
 
 		if !fetch module, this->_modules[name] {
-			throw new Exception("Module '" . name . "' isn't registered in the application container");
+			exception = new Exception("Module '" . name . "' isn't registered in the application container");
+  			this->_handleException(exception);
 		}
 
 		return module;
 	}
+	
+	/**
+  	 * Handles a user exception
+  	 */
+  	protected function _handleException(<\Exception> exception)
+  	{
+  		var eventsManager;
+  		let eventsManager = <ManagerInterface> this->_eventsManager;
+  		if typeof eventsManager == "object" {
+  			if eventsManager->fire("application:beforeException", this, exception) === false {
+ 				return false;
+  			} else {
+  				throw exception;
+  			}
+  		}
+  	}
 
 	/**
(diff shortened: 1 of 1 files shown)


Surt:
Need some help here. Is the `throw exception` correct in case of no event attached (`== false`)?

phalcon/application.zep
```diff
@@ -118,13 +118,31 @@ abstract class Application extends Injectable implements EventsAwareInterface
 	public function getModule(string! name) -> array | object
 	{
 		var module;
+		var exception;
 
 		if !fetch module, this->_modules[name] {
-			throw new Exception("Module '" . name . "' isn't registered in the application container");
+			let exception = new Exception("Module '" . name . "' isn't registered in the application container");
+  			this->_handleException(exception)
```
sergeyklay: ===> **;**

sergeyklay (changes requested):

sergeyklay:
The `3.0.x` branch is not longer supported. Could you please submit this PR to the 3.1.x

virgofx (commented):
Just for clarity -- this isn't the correct way to handle multi-module issues. It needs to be internal because of the registration of services that need to swap. 

I'm strongly against merging this because it creates overhead which allows userland capability to improperly hook and create multi-module forwarding.

This needs to be handled by the core application which can re-register (and de-register) all modules.

Surt:
> @virgofx Just for clarity -- this isn't the correct way to handle multi-module issues. It needs to be internal because of the registration of services that need to swap.
I'm strongly against merging this because it creates overhead which allows userland capability to improperly hook and create multi-module forwarding.
This needs to be handled by the core application which can re-register (and de-register) all modules.

I don't understand your objections. I mean, really don't understand it. 
A simple event telling that a module is not available, instead/or throwing an exception on the core Application seems pretty legit to me. A "notFound" route is unable to handle this situation, so we only have ways to overcome not found controller or actions, why not modules?

/realmodule/falseroute  => notFound route or dispatcher beforeException
/falsemodule/anyroute  => automatic exception, with no catch, unless you put it outside the application domain code.

sergeyklay:
The 3.0.x branch deleted

PR #12741, opened 2017-03-23
Added way to disable setters in model assign

Jurigag:
Hello!

* Type: new feature
* Link to issue:

Small description of change: currently `Phalcon\Mvc\Model::assign` is always using setters when it's not really necessary to always use it, just in most of apps setting properties could be enough but right now there is no way to do it. This is adding it by setting global/php ini option.

Thanks


virgofx:
This is a great fix. It's related to #11171 

I'd recommend that we actually make the default `true` which potentially breaks B/C; however, as noted in the original issue --- the assign should NEVER use setters as it always deals with raw data.

Jurigag:
Rebased, @sergeyklay what you think?

sergeyklay:
This feature must be documented. Firstly, could you please add PHPDoc example with `ini_set()` ?

Jurigag:
@sergeyklay added ini_set in phpdoc

Jurigag:
Rebased @sergeyklay

sergeyklay:
Thank you

PR #12746, opened 2017-03-24
beforeException event on Application getModule

Surt:
Hello!

* Type: new feature
* Link to issue: https://github.com/phalcon/cphalcon/pull/12733

Small description of change:

With this patch it's possible to attach a `beforeException` event to the Application.
The default behaviour is the same (throw exception) if not event listener is set.
Thanks

### DIFF
--- a/phalcon/application.zep
+++ b/phalcon/application.zep
@@ -118,12 +118,30 @@ abstract class Application extends Injectable implements EventsAwareInterface
 	public function getModule(string! name) -> array | object
 	{
 		var module;
+		var exception;
 
 		if !fetch module, this->_modules[name] {
-			throw new Exception("Module '" . name . "' isn't registered in the application container");
+			let exception = new Exception("Module '" . name . "' isn't registered in the application container");
+  			this->_handleException(exception);
 		}
 
 		return module;
+	}
+	
+	/**
+	 * Handles a user exception
+	 */
+	protected function _handleException(<\Exception> exception)
+	{
+		var eventsManager;
+		let eventsManager = <ManagerInterface> this->_eventsManager;
+		if typeof eventsManager == "object" {
+			if eventsManager->fire("dispatch:beforeException", this, exception) === false {
+				return false;
+			}
+			
+			throw exception;
+		}
 	}
 
 	/**


Jurigag:
3.1.x is only for bug fixes, please send it to 3.2.x, also create test and update changelog.md

phalcon/application.zep
```diff
@@ -118,13 +118,31 @@ abstract class Application extends Injectable implements EventsAwareInterface
...
+	
+	/**
+	 * Handles a user exception
+	 */
+	protected function _handleException(<\Exception> exception)
+	{
+		var eventsManager;
+		let eventsManager = <ManagerInterface> this->_eventsManager;
+		if typeof eventsManager == "object" {
+			if eventsManager->fire("dispatch:beforeException", this, exception) === false {
```
Jurigag: It should be application

phalcon/application.zep
```diff
@@ -118,13 +118,31 @@ abstract class Application extends Injectable implements EventsAwareInterface
 	public function getModule(string! name) -> array | object
 	{
 		var module;
+		var exception;
```
Jurigag: `var module, exception;`

phalcon/application.zep
```diff
@@ -118,13 +118,31 @@ abstract class Application extends Injectable implements EventsAwareInterface
...
 
 		if !fetch module, this->_modules[name] {
-			throw new Exception("Module '" . name . "' isn't registered in the application container");
+			let exception = new Exception("Module '" . name . "' isn't registered in the application container");
+  			this->_handleException(exception);
```
Jurigag: If _handlException will return false it will try to return empty module variable.

Surt:
Thanks @Jurigag I will try to make the corrections as soon as I can.

phalcon/application.zep
Surt: Are you sure about this? I can't get the fire behaviour but I think that if no event is defined, the result is `NULL`. So `false` means other fire event status returned.

The `dispatcher`where I copied this, works that way, only throws exception if the result of calling `fire`is NOT false.

Jurigag: Yes im sure, there is just missing return keyword. if exception will be handled like https://github.com/phalcon/cphalcon/pull/12746/files#diff-9721271d9a975486eef19f86a36c1fe9R139 here returned false the execution of code will just go on and it will return not set module value.

Surt:
Could someone take it from here? I lack the necessary development enviroment. The patch is a simple addition to allow control of "no module found" exception.

sergeyklay:
Because it is a new feature request (NFR), this should be submitted to the 3.2.x branch.

Anyway, I will keep this open for a some time to allow other contributors a chance to speak up, and to take a fresh look at it later. But at the moment it seems good to me.

Yes, we don't add NFRs without tests. And could you please submit this to the 3.2.x branch.

sergeyklay:
@Surt ping

PR #12756, opened 2017-03-27
Take advantage of prefix in router for faster handling

Jurigag:
Hello!

* Type: new feature
* Link to issue: https://github.com/phalcon/cphalcon/issues/12758

Small description of change: Right now router always iterate through all routes even when for example using router groups prefix it couldn't have to. This PR adds such possibility to first check prefix if handledUri starts with it, if not then it continues to next prefix group.

Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - Fixed `Phalcon\Mvc\Micro:handle` to correctly handle `before` handlers [#10931](https://github.com/phalcon/cphalcon/pull/10931)
 - Fixed `Phalcon\Mvc\Micro:handle` to correctly handle `afterBinding` handlers
 - Added way to disable setters in `Phalcon\Mvc\Model::assign` by using `Phalcon\Mvc\Model::setup` or ini option
+- Router now uses `Phalcon\Mvc\Router\Group::getPrefix` to optimize routes into groups for faster handling, added same possibility for `Phalcon\Mvc\Router::add` using `prefix` option in paths(using `Phalcon\Mvc\Router::usePrefix` you can actually select if it should affect pattern)
 
 # [3.1.2](https://github.com/phalcon/cphalcon/releases/tag/v3.1.2) (2017-04-05)
 - Fixed PHP 7.1 issues [#12055](https://github.com/phalcon/cphalcon/issues/12055)
--- a/phalcon/mvc/router.zep
+++ b/phalcon/mvc/router.zep
@@ -95,6 +95,10 @@ class Router implements InjectionAwareInterface, RouterInterface, EventsAwareInt
 
 	protected _notFoundPaths;
 
+	protected _routesSorted = false;
+
+	protected _usePrefix = false;
+
 	const URI_SOURCE_GET_URL = 0;
 
 	const URI_SOURCE_SERVER_REQUEST_URI = 1;
@@ -108,19 +112,18 @@ class Router implements InjectionAwareInterface, RouterInterface, EventsAwareInt
 	 */
 	public function __construct(boolean! defaultRoutes = true)
 	{
-		array routes = [];
+		array routes = ["/" : [],"" : []];
+
 
 		if defaultRoutes {
-
 			// Two routes are added by default to match /:controller/:action and
 			// /:controller/:action/:params
 
-			let routes[] = new Route("#^/([\\w0-9\\_\\-]+)[/]{0,1}$#u", [
-				"controller": 1
-			]);
-
-			let routes[] = new Route("#^/([\\w0-9\\_\\-]+)/([\\w0-9\\.\\_]+)(/.*)*$#u", [
-				"controller": 1,
+			let routes["/"][] = new Route("#^/([\\w0-9\\_\\-]+)[/]{0,1}$#u", [
+                "controller": 1
+            ]);
+			let routes["/"][] = new Route("#^/([\\w0-9\\_\\-]+)/([\\w0-9\\.\\_]+)(/.*)*$#u", [
+            	"controller": 1,
 				"action": 2,
 				"params": 3
 			]);
@@ -329,7 +332,8 @@ class Router implements InjectionAwareInterface, RouterInterface, EventsAwareInt
 			vnamespace, module,  controller, action, paramsStr, strParams,
 			route, methods, dependencyInjector,
 			hostname, regexHostName, matched, pattern, handledUri, beforeMatch,
-			paths, converters, part, position, matchPosition, converter, eventsManager;
+			paths, converters, part, position, matchPosition, converter, eventsManager,
+			prefix, prefixGroup;
 
 		if !uri {
 			/**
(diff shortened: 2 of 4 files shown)


sergeyklay:
First ow all we need to discuss this. Could you please open NFR.

Cc: @phalcon/framework-team

Jurigag:
Yea i will do NFR and post some results from benchmark how it affects speed or something.

Jurigag:
Any info? What you guys think about it?

sergeyklay:
Thank you

sergeyklay:
@Jurigag Could you please send PR to the docs repo with example?

Jurigag:
Sure thing.

michanismus:
@Jurigag You did introduce a new function `Router\Group::getRawRoutes()` which is not added to group interface. In Router::mount() you use this function which is not part of the interface. Your only check is the function argument <GroupInterface>.

michanismus:
In `Router::mount()` you forgot to reset the sorted flag.

Jurigag:
Yes it's not added because we can't add new methods to interfaces to not break BC - this is 3.2.x branch, not 4.0.x, in phalcon 4 it will be added.

About reset the sorted flag - what you mean here?

virgofx:
@Jurigag Any chance you can benchmark before and after --- similar to microbench in zephir ?

Jurigag:
I already did it in nfr

28.04.2017 1:19 AM "Mark Johnson" <notifications@github.com> napisał(a):

Jurigag:
Oh right i know what you mean by sort flag - sure need to reset it after mounting group, thanks.

michanismus:
@Jurigag Yeah, mounting groups I meant ;)

The interface... but right now you are expecting a GroupInterface as argument. So if I create an own group class which of cource implements GroupInterface, the default router will crash because I did not implement getRawRoutes() function.

I guess it also breaks BC right now ;)

michanismus:
To be honest, I don't like this PR.
Why add a prefix option to paths for a single route? So there are 2 places where I can add a prefix, a single route and a `Group`. But why would I add a prefix to a route if i could just simply add it to pattern.

Jurigag:
Will change this to check method_exists then.

Well i added simply to take advantage of faster handling for single routes too. Many people are using just single routes.php file with many routes or something like this when handling many routes will be slow. Adding option to grouping them by prefixes any way will speed up stuff.

michanismus:
Does it really speed it up? 
For me it breaks a lot of stuff because I extended the Router class.
Means a looooot of work for me. 

If the router groups the routes by prefix why don't use the `Router\Group` internally in `Router`.
Btw it is not possible to set the `Route`class to be used. The creation of route objects is hardcoded on different places.

My suggestion is, revert this PR, refactor the whole Router / Group stuff and implement it in 4.0.0

Jurigag:
Yes it really speed it up - check issue https://github.com/phalcon/cphalcon/issues/12758#issue-217299834

michanismus:
Ok for the case someone uses group with prefixes and he wants to handle it 10000x times ;)
But for a real scenario? I mean, what is the speed benefit to handle 1 route, to find the one route that matches? Must be something around 0.05 - 0.010 millisecs...

Jurigag:
Depends how many routes you have in application.

michanismus:
Yes, for me I do have an average of 200 routes, I bet it won't give the expected advantage.

The introduction of `prefix` as a path argument to single routes which will be added to router? Prefix for a single route? I bet everyone would define the prefix in pattern. If not they would use `Router\Group`...

I really like the idea to improve the whole router stuff but I think it would be better to do it with 4.0.0 and refactor the whole stuff together.

Jurigag:
Yea maybe, i can add there already what i mention too, maybe it's better idea,

michanismus:
Yes - I have something in mind to store all routes in `Router\Group` in router internally - if you add a group it will simply added to router stack. Something like this...

Or just drop the `Router\Group` and add this functionality to the Router itself. Something like:

```php 

$prefix = $router->getPrefix();
$router->setPrefix('/admin');
$router->add('/login');
$router->add('/logout');
$router->add('/dashboard');
$router->setPrefix($prefix);
```

Just to simplify the route registration and generation. At the moment there are 2 places where `RouteInterface`objects are initiated, the router itself and the `Router\Group`. 

It would be easier to define the final route class to be used - which is not possible at the moment. Only if you overload the whole Router and Group.

```php
$router->setRouteClass(MyFancyRoute::class);
```

I do have a lot of ideas ;)

Jurigag:
No, `Router\Group` is good - it allows you to have seperated classes with your routes, it's nice feature to have. Im using it in my apps and can't live without it.

This won't simplify things, it will make it longer imho.

About route class - well could be good if someone needs it.

Currently if for example request method is post we are checking all the routes(get, put, delete anyway) and we just check them against method - i want to change this to, so when we do addGet addPut etc it will be added to get/put group and we first check request type and use correct group to handle our route - it will be add speed too.

michanismus:
Yes, grouping routes internally by method makes sense but how would you store routes which are accessible for POST, PUT and GET?

Ok - `Router\Group` should stay, but the whole Route initiation stuff should be taken out of the `Router\Group`. Right now there are two places where it happens. I would suggest to try to put it into `Router`. Just use `Router\Group` as a better simple array collection which will be compiled at mount point, not when you add a route to a group.

Jurigag:
Well i updated my comment, we could actually remove it if we would add way to merge router classes.

something like:

```php
$router1 = new Router();
$router1->add('/login');
$router2 = new Router();
$router2->add('/dashboard');
$router1->merge($router2); // now router 1 have both login and dashboard
```

You mean via method? It will be just added to those 3 groups.

michanismus:
That is another possible option :) 

Yes I mean the via. That would mean you would add the same route to 3 groups?

michanismus:
What if you would filter the routes array according to method used before you iterate through the routes? Does not make sense right? You could just loop the the routes and check the method in place....

Bad idea :)

Jurigag:
The best will be just:

```
protected _getRoutes;

protected _postRoutes;

public function handle()
{
    // ...
    switch request->getMethod {
        case "GET":
            let routes = this->_getRoutes;
            break;
    }

    for prefix, prefixGroup in routes {
        if prefix !== "" && !starts_with(handledUri, prefix) {
            continue;
        }
        for route in prefixGroup {
        
        }
    }
}
```

Jurigag:
The biggest issue will be this via method in route - that's it here. Not really sure how to solve it, doing something like add it to all request types and then remove from certain ones - feels kind of messy.

Jurigag:
I see two ways to solve it:

- add it to all method groups when doing `router->add` without methods, when someone will call via - remove it from method groups which are not selected
- add it to some seperated group like `_nonMethodRoutes` after selecting method group those routes with method groupes, if user will call via remove this route from only this routes and add to other

The question is what will be faster, array_merge or adding to all and then removing from added without a need.

michanismus:
I would store them in `$this->_routes[request->getMethod]`.

What do you think about introducing a Route Factory and a Route Collection?

Jurigag:
Yea but there will be then $this->routes[request->getMethod] and here prefix => group, pretty much.

What you mean by collection? Seperated class for storing routes? That would make more sense.

michanismus:
It would be much easier if we only allow route objects to be added to router/collection. These routes have all necessary informations, methods, name etc.

Yes a storage object for the compiled routes, internally used by any router object.

Jurigag:
But it's like this already, router have only route objects in routes property.

michanismus:
Thats right but at the moment we can add routes like `router->get('/pattern')`. 
I mean only allow route objects to be added `router->add(<RouteInterface> route)`.

Jurigag:
I don't like this really, router->add('/pattern') feels better, what we could do allow adding just router objects alone if someone would like, like `addRoute` method.

It would require people to do a lote of changes to their applications for phalcon 4.

michanismus:
Maybe create a factory within `Route` for example:

```
Route::get('pattern', options)
```

Jurigag:
But this doesn't change really anything, add method already pretty much do this and construction of route in current `add`, what's a difference?

michanismus:
What I am thinking about is the following. If we add a route the current way
```
router->get('/pattern')
```
it would be added to _routes[GET] or _getRoutes. The method return the route itself to be able to name it or even change the method with `via` route method. 
How would you handle the modification of the currently added route?

Jurigag:
Objects are passed as references so name change is not a problem, via is, I posted above how to solve this, not sure how your factory is supposed to fix this

Jurigag:
After some thinking I know what you mean, I guess you want to do something like:

Route::via? This would actually make sense, I think it will be better.

Well I think we should leave old way of adding routes but add method should also just accept route object and add those factories.

Jurigag:
Just instead of return route make it void or return this and problem with via is gone.

michanismus:
Yes, I think thats the best way to archive a solid base for adding routes to router/collection.
The current way is not really compatible with new optimizations...

Anyone else who like to give his/her opinion on this topic? @sergeyklay @virgofx

Issue #12743, opened 2017-03-23
Model getChangedFields() returning empty array


garyrutland:
When updating a model I'm doing some checks on the changed fields which was working up until `3.1.0` where it now seems to return an empty array every time now.

In the example is the PHP printing out the model details and the changed fields after each call, the model saves into the DB without issue but the changed fields is not reflecting that.
My models have `keepSnapshots` set to true and `deleted` is not part of `skipAttributes`.

Example output:
```php
print_r($model->toArray());
print_r($model->getChangedFields());

Array
(
    [id] => 1
    [created] => 2017-03-23 20:34:20
    [updated] => 2017-03-23 20:41:53
    [deleted] => 
)
Array
(
)

Array
(
    [id] => 1
    [created] => 2017-03-23 20:34:20
    [updated] => 2017-03-23 20:41:49
    [deleted] => 2017-03-23 20:41:53
)
Array
(
)

```
Code to reproduce:

```php
$model->save([
    'someField' => 'someValue',
]);

print_r($model->toArray());
print_r($model->getChangedFields());
```

Calling `hasChanged` also returns a boolean result of false.

### Details

* Phalcon version: 3.1.0
* PHP Version: (7.0.17)
* Operating System: Debian (Docker)
* Installation type: Compiling from source
* Server: Nginx

Jurigag:
The old behawior was actually a long sitting bug here https://github.com/phalcon/cphalcon/issues/11007 https://github.com/phalcon/cphalcon/issues/11818 https://github.com/phalcon/cphalcon/issues/11424 there were a lot of discussion and from most opinions as i saw that snapshot after save should be updated which right now is happening. From 3.0.4 after saving obviously there are no changed fields because they are saved - this is why you have empty array and i think this is much more expected result. To make your code working again:

```php
$model->assign([
    'someField' => 'someValue',
]);

print_r($model->toArray());
print_r($model->getChangedFields());
$model->save();

garyrutland:
Ok, so I had an event being triggered on `afterSave` as it confirmed any changes were already saved.  
I guess I have the option to move it to `beforeSave` or try and find a different approach?

Jurigag:
Yes, move it to beforeSave - here you will getChangedFields and it will return changed fields before saving correctly.

garyrutland:
Moving it to `beforeSave` seems to resolve it, not ideal for what I'm trying to achieve but I'll figure something out.

Jurigag:
[...]

Issue #11818, opened 2016-05-22
[NFR] Implement Model snapshot loading call


temuri416:
Hi all,

It would be nice to have an API method to allow manual loading of Model's snapshot:

``` php
$robot = new Robot;
$robot->save();
$robot->loadSnapshot();
```

Thanks!

temuri416:
Hi, @sergeyklay. What are your thoughts on this? Should ->save() automatically set updated snapshot?

Thanks!

sergeyklay:
I'll try to sort out as soon as posible

Jurigag:
I made PR with fix for this, let me know what you think.

sergeyklay:
Fixed in the `3.1.x` branch.

Issue #11424, opened 2016-02-17
Dynamic updates / snapshot data work correctly only once


ioncube:
See: https://forum.phalconphp.com/discussion/10477/snapshot-data-not-updated-following-save-update-bug-limitation-u

When making a save() or update() to a model with snapshots in place, any subsequent changes and save/update to the model instance also include fields changed in prior updates but that have not changed since. It appears that snapshot data is not updated with the values applied during a save/update operation. This is inefficient and problematic if a prior update is not idempotent, such as using now().

It should be possible to make repeated change/update operations on a model and each time have only the changed fields updated as that is the purpose of the feature.

nirnanaaa:
anyone working on this? In our case it is really anoying to perform a new find every time you just want to update a previously saved model.

Jurigag:
So pretty much - save(if succesfull) should just update snapshot if needed ?

duesentrieb26:
`$model = new Address();
        $model->setCity('Munich')
            ->setAddressTypeId(1);

        if (!$model->save()) {
            print_r($model->getMessages());
        }

        $model->setStreet('test');

        if (!$model->save()) {
            print_r($model->getMessages());
        }
`
=>  The record doesn't have a valid data snapshot

virgofx:
@Jurigag Did your recent changes for the cache snapshot fix this? Or is it related? Maybe we can fix both for 3.0.3 real quick

Jurigag:
No, it's not related really, it's diffrent issue. I think that save method in model after successfull save should just update snapshot to current one which is not doing now ?

Jurigag:
I made PR with fix for this, let me know what you think.

sergeyklay:
Fixed in the `3.1.x` branch.

Issue #11007, opened 2015-10-07
Phalcon 2.0.* - creating object does not set model snapshots.


temuri416:
Hi,

My model:

``` php
class Robots extends Model
{
    public function initialize()
    {
        $this->keepSnapshots(true);
    }
}
```

Immediately after creating a new instance of model, its snapshot is blank.

That prevents me from checking whether a field has been updated if I decide to update it later during the same request.

Snapshots only get set when loading a model. But they should also be set after a new model is created.

Thanks!

ashpumpkin:
+1, but I think it's not a bug, but a very annoying feature. Would be very cool if someone will fix it

temuri416:
Snapshot should also be refreshed when Model is saved/updated.

web777:
Old issue. From 1.3.x

ruudboon:
Refreshing the object doesn't help.

```
<?php
$model = new MyModel();
$model->name = "test";
$model->save();
$model->refresh();
```
I solved this by completely reloading the object :(
````
<?php
$model = new MyModel();
$model->name = "test";
$model->save();
$model = MyModel::findFirst($model->payment_id);
```

sergeyklay:
@ruudboon yeah :confused: https://github.com/sergeyklay/incubator/blob/2.0.x/Library/Phalcon/Mvc/Model/Behavior/NestedSet.php#L338

ruudboon:
I'm sorry @sergeyklay I don't understand your feedback.

sergeyklay:
I mean that this is a known issue and I too just had to use this trick

ruudboon:
Ah sorry! Now I see. Completely clear. Hope to get some time soon to contribute to Phalcon. Love this project.

temuri416:
@andresgutierrez 

Any update on this? Do you plan to implement it in 2.1?

Thanks!

andrew-demb:
Up. 
This will be very helpful for development, but if now implement the feature it can break some applications.

@andresgutierrez, can you give us a feedback about this?

temuri416:
@sergeyklay Is this on your radar? :)

Jurigag:
No because 2.0.x is no longer supported.

temuri416:
Sorry, I wasn't expecting this to be fixed in 2.0.x :-) Any hope for this go to on 3.x branch?

Mechzeit:
My work around for now is using a base mode class, with the 'afterCreate' method
[...]

PR #12760, opened 2017-03-28
3.2.x model hasupdated

Jurigag:
Hello!

* Type: new feature
* Link to issue: https://github.com/phalcon/cphalcon/issues/12743

Small description of change: In version 3.1.0 we fixed long sitting bug that snapshot is not updated after save/create/refresh operation being made

> Fixed Phalcon\Mvc\Model to create/refresh snapshot after create/update/refresh operation #11007, #11818, #11424

Since 3.1.0 now it's happening, though in some cases people would specifically need to check if field was really changed(i call this updated in this case) after save/create, right now if someone has old code working on this bug it will no longer work. This PR is adding such ability by adding two new methods so by small change it can work as expected.

Thanks


nsossonko:
What's the difference between this and hasChanged?

sergeyklay:

sergeyklay:
Thank you

Issue #12221, opened 2016-09-14
NFR: Get one element of the array with config->get()


xerron:
The good from the get function is that you can set a default value.

now:

``` php
$adapter = $this->config->database->adapter ? $this->config->database->adapter : 'Mysql' ;
```

This is better:

``` php
$adapter = $this->config->get('database.adapter', 'Mysql');
```

mzf:
Dot notation is very good idea

sergeyklay:
I think will be better:
```php
$adapter = $this->config->path('application.database.adapter', 'Mysql');
$adapter = $this->config->path('application/database/adapter', 'Mysql', '/');
```

mzf:
Ye. Fine option

xerron:
it's a little confusing use 'path', for me is better 'get'

Jurigag:
But what if for some weird reason someone already have named section with dot ?

xerron:
@Jurigag  you mean this?

```php
<?php
return new \Phalcon\Config([
    'database' => [
        'adapter' => 'Postgresql',
        'host' => 'localhost',
        'username' => 'xerron',
        'password' => 'root',
        'dbname' => 'db_uwu'
    ],
   // ...code ...
    ' section.with.dot' =>  'mmm...'
]
```
good question. :+1:

xerron:
@sergeyklay :+1:

michanismus:
I need this so bad! 100x thumbs up

sergeyklay:
Will be implemented in 3.1.0

Jurigag:
So implement it in other places too, as far as i know people wanted this in forms too and somewhere else.

xerron:
In PHP7 default values is most readable:

```php
$adapter =  $this->config->get('database.adapter') ?? 'Mysql';
```

sergeyklay:
but it's longer and we support not only PHP7

Jurigag:
Just add option to just call `$this->config->path('application.database.adapter')` without second argument, and you can use default value like php7 with  null coalescing operator.

xerron:
Well, this is good:

```php
$adapter = $this->config->path('application/database/adapter', '/');
```

daison12006013:
This is cool

udarkness:
@sergeyklay  Actually what @xerron needs, is only:
$adapter = $this->config->database->adapter ?: 'Mysql' ;
Not PHP7 ?? required for this, just the classic Elvis operator ?:

[...]

PR #12762, opened 2017-03-29
3.2.x Method `path` added to `Phalcon\Config`

michanismus:
Hello!

* Type: new feature
* Link to issue: https://github.com/phalcon/cphalcon/issues/12221

Small description of change:

I added a `path` method to `Phalcon\Config` to fetch values by using a dot separated path.

```php

$config = new \Phalcon\Config(
   [
        'phalcon' => [
            'baseuri' => '/phalcon/'
        ],
        'models' => [
            'metadata' => 'memory'
        ],
        'database' => [
            'adapter'  => 'mysql',
            'host'     => 'localhost',
            'username' => 'user',
            'password' => 'passwd',
            'name'     => 'demo'
        ],
        'test' => [
            'parent' => [
                'property' => 1,
                'property2' => 'yeah'
            ],
        ],
   ]
);

$config->path('test.parent.property2'); // yeah
$config->path('test.parent'); // Phalcon\Config
$config->path('test/parent/property3', 'no', '/'); // no

\Phalcon\Config::setPathDelimiter('/');
$config->path('test/parent/property2'); // yeah

```


michanismus:
@sergeyklay I don't know why the PHP7 tests fail.

sergeyklay:
@michanismus I'll take a look a bit later

phalcon/config.zep
```diff
@@ -79,6 +81,61 @@ class Config implements \ArrayAccess, \Countable
...
+	 *
+	 *<code>
+	 * echo $config->get("unknown.path", "default", ".");
+	 *</code>
+	 */
+	public function path(var path, var defaultValue = null, var delimiter = null) -> var
+	{
+		var key, keys, config, temp;
+
+		let path = strval(path);
```
Jurigag: In arguments `path(string path)` ?

phalcon/config.zep
```diff
@@ -79,6 +81,61 @@ class Config implements \ArrayAccess, \Countable
...
+			if !isset config->{key} {
+				break;
+			}
+
+			if empty keys {
+				return config->{key};
+			}
+
+			let temp = config->{key};
+			let config = temp;
```
Jurigag: Why just not let `config = config->{key}` ?

phalcon/config.zep
```diff
@@ -50,6 +50,8 @@ use Phalcon\Config\Exception;
 class Config implements \ArrayAccess, \Countable
 {
 
+	protected static _delimiter = ".";
```
Jurigag: Why static?

phalcon/config.zep
michanismus: I thought I found the PHP7 test error. Previously I wrote 
let config = config->{key}

phalcon/config.zep
michanismus: To archieve a global changement over all Config instances.

phalcon/config.zep
michanismus: Thats true, I am going to change the argument definition.

phalcon/config.zep
Jurigag: Then why `setPathDelimiter` is non-static?

phalcon/config.zep
```diff
@@ -79,6 +81,58 @@ class Config implements \ArrayAccess, \Countable
...
+
+		if isset this->{path} {
+			return this->{path};
+		}
+
+		if empty delimiter {
+			let delimiter = self::_delimiter;
+		}
+
+		let keys = explode(delimiter, path),
```
Jurigag: Actually when i tried to reproduce seg fault i have problem on this line if there is no delimiter argument provided.

phalcon/config.zep
Jurigag: Try to change `_delimiter` to `delimiter` it could fix seg fault.

michanismus: Indeed, I will change it to static.

Jurigag:
Also squash commits at the end.

phalcon/config.zep
michanismus: Hmm, but only on PHP7?

phalcon/config.zep
michanismus: Ok, I'll try.

michanismus:
@Jurigag 

Strange! PHP5 test is fine, no seg fault anymore on PHP7 but now an exception is thrown:

[PHPUnit_Framework_Exception] Invalid arguments supplied for explode()

Exception is thrown on line 39:
expect($config->path('test.parent.property2'))->equals('yeah');

So there is no delimiter argument and the delimiter should be the static default.
Where is the error for PHP7? Or what changed for Zephir and PHP7? Empty != null?

		if empty delimiter {
			let delimiter = self::delimiter;
		}

		let keys = explode(delimiter, path)

Jurigag:
No, it's not about empty stuff, i think it's about something else like some self call problem.

michanismus:
@Jurigag I added a static getter for the path delimiter and use it in path method. Tests passed!

Seems it is a bug which shows up while fetching a protected static var from public method or it is `explode` related, maybe combined!?

phalcon/config.zep
```diff
@@ -50,6 +50,10 @@ use Phalcon\Config\Exception;
 class Config implements \ArrayAccess, \Countable
 {
 
+	protected static _pathDelimiter;
```
sergeyklay: Please don't introduce new static properties, you can use object instance to get properties

sergeyklay (changes requested):

phalcon/config.zep
```diff
@@ -223,6 +270,29 @@ class Config implements \ArrayAccess, \Countable
...
+	 */
+	public static function setPathDelimiter(string! delimiter = null) -> void
+	{
+		let self::_pathDelimiter = delimiter;
+	}
+
+	/**
+	 * Gets the default path delimiter
+	 */
+	public static function getPathDelimiter() -> string
```
sergeyklay: 👎 for static

phalcon/config.zep
```diff
@@ -223,6 +270,29 @@ class Config implements \ArrayAccess, \Countable
 	}
 
 	/**
+	 * Sets the default path delimiter
+	 */
+	public static function setPathDelimiter(string! delimiter = null) -> void
```
sergeyklay:
```zep
/**
 * @var string
 */
protected _pathDelimiter { set, get };
```

sergeyklay (changes requested):

phalcon/config.zep
Jurigag: But he wrote what is purposed of this, to configure delimiter once for all Config instances.

michanismus: @sergeyklay Static because I thought it would be nice to have the possibility to change the default delimiter for all `Phalcon\Config` instances with one call.

sergeyklay: @Jurigag This can be done once in service definition

michanismus: Another point. The toArray method uses get_object_vars which would return the _delimiter object property in each array.

michanismus: It would be impossible/erroneous to use a config key named 'pathDelimiter' or '_delimiter' or what ever I would name the var to hold the path delimiter. `Phalcon\Config` implements ArrayAccess...

I could use a global constant PHALCON_CONFIG_PATH_DELIMITER to test for a changed default delimiter!?

sergeyklay (approved):

sergeyklay:
I will keep this open for a day or two to allow other contributors a chance to speak up, and to take a fresh look at it later. But at the moment it seems good to me. Thank you

Jurigag:
Also as i remember we need similar stuff for forms.

phalcon/config.zep
```diff
@@ -86,7 +86,7 @@ class Config implements \ArrayAccess, \Countable
 	 * Returns a value from current config using a dot separated path.
 	 *
 	 *<code>
-	 * echo $config->get("unknown.path", "default", ".");
+	 * echo $config->path("unknown.path", "default", ".");
```
sergeyklay: 👍

Jurigag:
Squash commits please.

michanismus:
Done...

sergeyklay:
Thank you

PR #12776, opened 2017-04-07
Factory load for di

Jurigag:
Hello!

* Type: new feature
* Link to issue:

Small description of change: I already introduced factory loaders, this is other way to have output as array to pass it to `Phalcon\Di::set()` as second parameter. Is returns array for complex service registration, so for example instead of:

```php
$di->set('db', Factory::load($config->database));
```

Which will create adapter object here

You can use:

```php
$di->set('db', Factory::loadForDi($config->database));
```

Which will create adapter object on `$di->get('db')`

Thanks

### DIFF
--- a/phalcon/cache/backend/factory.zep
+++ b/phalcon/cache/backend/factory.zep
@@ -50,9 +50,57 @@ class Factory extends BaseFactory
 		return self::loadClass("Phalcon\\Cache\\Backend", config);
 	}
 
+	/**
+	 * @param \Phalcon\Config|array config
+	 */
+	public static function loadForDi(var config) -> array
+	{
+		return self::loadAsArray("Phalcon\\Cache\\Backend", config);
+	}
+
 	protected static function loadClass(string $namespace, var config)
 	{
-		var adapter, className, frontend;
+		var frontend, className, params;
+
+		let params = self::checkArguments($namespace, config);
+
+		let frontend = FrontendFactory::load(params["frontend"]);
+		let className = params["className"];
+
+		return new {className}(frontend, params["config"]);
+	}
+
+	protected static function loadAsArray(string $namespace, var config)
+	{
+		var params;
+
+		let params = self::checkArguments($namespace, config);
+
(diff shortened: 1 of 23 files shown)


Izopi4a:
hmm nice 👍 

may I suggest adding something like 

```
function checkLoadArguments() { //or something like that i dont know. A bit tired atm
 if .. throw ...
}
```

so checking of parameters doesnt happen in two places `loadClass` and `LoadAsArray`

Jurigag:
@Izopi4a well good ideay, i will do it

phalcon/cache/backend/factory.zep
```diff
@@ -79,4 +84,60 @@ class Factory extends BaseFactory
...
+		if !fetch frontend, config["frontend"] {
+			throw new Exception("You must provide 'frontend' option in factory config parameter.");
+		}
+
+		if fetch adapter, config["adapter"] {
+			unset config["adapter"];
+			unset config["frontend"];
+			let className = $namespace."\\".camelize(adapter);
+
+			if typeof frontend != "array" && !(frontend instanceof Config) {
```
sergeyklay: frontend instanceof Config ?

sergeyklay (changes requested):

phalcon/cache/backend/factory.zep
Jurigag: Oh right `config->toArray();` will just return array without config instances? My mistake then.

Jurigag:
Some weird random seg fault on php 5.6

Jurigag:
This is ready for mergei  guess.

sergeyklay:
@Jurigag I'm not sure about CHANGELOG.md. It seems to me it is outdated. Could you please rebase onto 3.2.x. Just to be sure that all ok

Jurigag:
Sure thing.

sergeyklay:
Thank you

sergeyklay:
Can you update the docs?

PR #12813, opened 2017-04-25
Dispatcher enhancement

dongasai:
Hello!

Type: new feature

In raising this pull request, I confirm the following:

- [ ] I have read and understood the Contributing Guidelines
- [ ] I have checked that another pull request for this purpose does not exist

Small description of change: Dispatcher enhancement, Allow interrupt action, Produce more of the possible.

Thanks

### DIFF
--- a/phalcon/dispatcher.zep
+++ b/phalcon/dispatcher.zep
@@ -48,6 +48,8 @@ abstract class Dispatcher implements DispatcherInterface, InjectionAwareInterfac
 
 	protected _forwarded = false;
 
+	protected _canAction = true;
+
 	protected _moduleName = null;
 
 	protected _namespaceName = null;
@@ -196,6 +198,22 @@ abstract class Dispatcher implements DispatcherInterface, InjectionAwareInterfac
 	public function setDefaultAction(string actionName)
 	{
 		let this->_defaultAction = actionName;
+	}
+
+	/**
+	 * Returns the canAction
+	 */
+	public function getCanAction() -> boolean
+	{
+		return this->_canAction;
+	}
+
+	/**
+	 * Sets the  canAction 
+	 */
+	public function setCanAction(boolean canAction)
+	{
+		let this->_canAction = canAction;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


phalcon/dispatcher.zep
```diff
@@ -199,6 +201,22 @@ abstract class Dispatcher implements DispatcherInterface, InjectionAwareInterfac
 	}
 
 	/**
+	 * Returns the canAction
+	 */
+	public function getCanAction() -> boolean
```
Jurigag: isCanAction

virgofx:
I don't think this is necessary...

sergeyklay:
I agree with @virgofx

dongasai:
@virgofx  @sergeyklay This is a can enhance the Dispatcher in the MVC mode of Dispatch ,I use the sample is as follows:
`
<?php

class Test9Controller extends Phalcon\Mvc\Controller
{

    public  function initialize()
    {
        # For some reason, are not allowed to perform indexAction
        $this->dispatch->setCanAction(false);
    }

    public function indexAction()
    {
        # code 
    }

    /**
     * noCallAction
     */
    public function noCallAction()
    {
       # Perform other operations
    }

}
`

dongasai:
It solves the: Verifies the permissions before performing the **_Action_** will affect the operation of the Action, but can't abandon **_afterDispatch_**. **_AfterExecuteRoute_** operation

dongasai:
In the **_MVC_** structure, where did you perform the **_acl_** ? @virgofx  **Route**(_Before the new Controller_)?**dispatch:afterInitialize**( _Controller initialize_)?or  **The Action begins**(_This is not a right choice!_)
@sergeyklay

sergeyklay:
@dongasai It seems to me you can address this by using ACL

virgofx:
@dongasai Refer to phalcon documentation on dispatch forwarding. It will address exactly what you need. You can hook an ACL anytime or custom logic and forward to another action/controller as needed.

dongasai:
@sergeyklay  First of all, this is not for enhanced authentication, this is the enhancement of dispatch,Then, if you ACL to return a false, you should know how to deal with the logic of it.
As @virgofx  said,Controller forwarding,But I don't think it's a good solution

dongasai:
And forwarding will lead to continue the while

PR #12816, opened 2017-04-26
Fix the test when sql_mode=only_full_group_by

sjinks:
```
1) PaginatorTest::testQueryBuilderPaginatorGroupBy
PDOException: SQLSTATE[42000]: Syntax error or access violation: 1055 Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'phalcon_test.personnes.cedula' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by
```


sjinks:
Oops, MySQL 5.6.33 does not have `ANY_VALUE()` :-(

Issue #11133, opened 2015-11-13
ResultSet\Simple delete method returns true when a constraint violation occurs if the model has a virtual FK defined


GavinMcL:
Using Robots demo as the example;

If I have a Robots model, which is related to a Parts model via RobotsParts, and Robots has this relationship map;

``` php
$this->hasMany('id', 'RobotsParts', 'robots_id', ['alias' => 'RobotsParts']);
```

if I do 

``` php
$robots = Robots::find();
$robots->delete();
```

I correctly get a PDOException

``` sql
PDOException: SQLSTATE[23000]: Integrity constraint violation: 
1451 Cannot delete or update a parent row: a foreign key constraint fails (`robots`.`robots_parts`, CONSTRAINT `robots_parts_ibfk_1` FOREIGN KEY (`robots_id`) REFERENCES `robots` (`id`))
```

**however**
if I alter the Robots relationship map to define a virtual foreign key with a friendlier message, eg

``` php
$this->hasMany('id', 'RobotsParts', 'robots_id', [
    'alias' => 'RobotsParts', 
    'foreignKey' => ['message' => 'robot is associated to a part']
]);
```

then re-run

``` php
$robots = Robots::find();
$robots->delete();
```

$robots->delete() returns true, even though $robots->getMessages() has the FK message I defined.  The actual database records are still correctly in place.

I would expect $robots->delete() to return false if any of the contained records fail to delete, in the same way as calling

``` php
$robot = Robots::findFirst();
$robot->delete();
```

will return false with the same FK violation.

GavinMcL:
I've created 2 files in gist which demonstrate the behaviour I'm getting.

https://gist.github.com/GavinMcL/4d58e9ee2182ede3a7da

Jurigag:
What happens if you add , 'foreignKey' => true also in RobotsParts relations ? Also try maybe to add it like this:

'foreignKey' => ['message'=>'test message','action'=>Relation::NO_ACTION(it should be already be default)]

GavinMcL:
Hi sorry for the delay in responding, got diverted onto different project.

I've added foreignKey into RobotsParts, makes no difference

Adding foreignKey with a message makes no difference, but at least shows it is the fk in the Robots class that is being triggered

Adding 'action'=>Relation::NO_ACTION seems to have the same effect as NOT declaring the relations with foreignKey - the original behaviour happens where the underlying PDO Exception is thrown, which is what I was trying avoid in the first place.

PR #12828, opened 2017-04-28
Give different result from Resultset::delete() depending on success

1RV34:
Hello!

* Type: bug fix / new feature (unsure, it's unwanted behavior that is inconsistent with the behavior of `Model::delete()`)
* Link to issue: #11133

Small description of change: When any `Model::delete()` fails the result bubbles to cause `ResultSet::delete()` to return `false` as well.

Thanks


Jurigag:
Update changelog also and add some simple test.

sergeyklay (approved):

Jurigag (changes requested):
Update changelog, add test

1RV34:
I've added a message to the changelog but not sure if it's according to the standards that those messages should be.

Secondly, I'm not completely sure how to properly create a test for this. Basically if any of the `Model::delete()` calls that are made fails and causes a transaction rollback to happen the result should be `false` just like the `Model::delete()` result. Could I get some help with this?

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 - Fixed `Phalcon\Mvc\Micro:handle` to correctly handle `before` handlers [#10931](https://github.com/phalcon/cphalcon/pull/10931)
 - Fixed `Phalcon\Mvc\Micro:handle` to correctly handle `afterBinding` handlers
 - Added way to disable setters in `Phalcon\Mvc\Model::assign` by using `Phalcon\Mvc\Model::setup` or ini option
+- Made `Resultset::delete()` return `false` when any of the `Model::delete` calls fails [PR #12828](https://github.com/phalcon/cphalcon/pull/12828) fixing [issue #11133](https://github.com/phalcon/cphalcon/issues/11133)
```
Jurigag: No need for link ot pr, just only add issue like other changes aboves. Also without "issue"

Jurigag:
About test - just return some random false in some of model::delete, then check if result returned by resultset->delete is false.

sergeyklay:
Just updated CHANGELOG

sergeyklay (approved):

sergeyklay:
Thank you

PR #12841, opened 2017-05-10
Adding hasFunction/Auto Include PHP Functions Method To Volt Compiler

Fenikkusu:
Hello!

* Type:  new feature
* Link to issue: NA

Small description of change:

In many of my projects, I find myself constantly adding PHP Internal Functions to the Volt Compiler Engine (IE: `substr`).

I found this very repetitive and frustrating to have to add to all of them. With this in mind, I've updated the compiler to have a new option of `includePhpFunctions`. When `_compileSource` is ran, it looks for this option and automatically registers all php internal functions within the compiler.

To maintain existing code bases, I've added a `hasFunction` method which I use in the function loop such that it only adds it if it wasn't previously added.

Thanks

### DIFF
--- a/phalcon/mvc/view/engine/volt/compiler.zep
+++ b/phalcon/mvc/view/engine/volt/compiler.zep
@@ -2684,6 +2684,10 @@ class Compiler implements InjectionAwareInterface
 
 	    for functionName in configOptions {
 	        if !this->hasFunction(functionName) {
+	            if !call_user_func("function_exists", functionName) {
+	                throw new Exception(functionName . " does not exist");
+	            }
+
 	            this->addFunction(functionName, functionName);
 	        }
 	    }
--- a/phalcon/mvc/view/engine/volt/compiler.zep
+++ b/phalcon/mvc/view/engine/volt/compiler.zep
@@ -226,6 +226,14 @@ class Compiler implements InjectionAwareInterface
 		return this;
 	}
 
+    /**
+     * Checks to see if a function was previously defined.
+     */
+	public function hasFunction(string! name) -> boolean
+	{
+	    return isset this->_functions[name];
+	}
+
 	/**
 	 * Register the user registered functions
 	 */
@@ -2191,9 +2199,10 @@ class Compiler implements InjectionAwareInterface
 	{
 		var currentPath, intermediate, extended,
 			finalCompilation, blocks, extendedBlocks, name, block,
-			blockCompilation, localBlock, compilation, options, autoescape;
+			blockCompilation, localBlock, compilation, options, autoescape, includePhpFunctions;
 
 		let currentPath = this->_currentPath;
+        let includePhpFunctions = false;
 
 		/**
 		 * Check for compilation options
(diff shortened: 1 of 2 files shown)


phalcon/mvc/view/engine/volt/compiler.zep
```diff
@@ -2210,8 +2219,23 @@ class Compiler implements InjectionAwareInterface
 				}
 				let this->_autoescape = autoescape;
 			}
+
+            /**
+             * Configures the compiler to register PHP Functions automatically.
+             */
+            if isset options["includePhpFunctions"] {
+                let includePhpFunctions = options["includePhpFunctions"];
```
Jurigag: How about make it possible to select which exactly php functions we want to include? Just accept boolean or array maybe?

phalcon/mvc/view/engine/volt/compiler.zep
```diff
@@ -2210,8 +2219,23 @@ class Compiler implements InjectionAwareInterface
...
+                if typeof includePhpFunctions != "boolean" {
+                    throw new Exception("'compileAlways' must be a bool value");
```
Fenikkusu: Oops...Error Mistake

phalcon/mvc/view/engine/volt/compiler.zep
Fenikkusu: Well technically I wouldn't really need to use function_exists. Just loop through the list of functions. If not set from the option, set default to be actual functions. I see no reason as why not...

Jurigag: Because someone could provie wrong, non existing function name.

Fenikkusu:
I've added the new updates. You can pass either and array or true. @sergeyklay, a question for you. When running 'function_exists' on 'substr', I kept getting false. After looking at the compiled code, I found that function_exists was being changed to zephir_function_exists. I got around it by using call_user_func instead of function_exists. Does this seem like the valid way to do this, or is it possible that zephir_function_exists wasn't working correctly?

Also, not sure what's going on with failures in Travis.

sergeyklay:
@Fenikkusu We'll double check `zephir_function_exists` asap. 

Cc: @sjinks

Jurigag:
Failures on travis are codeception bug, you will need to cherry pick some commit from @sergeyklay repo

sergeyklay:
https://github.com/phalcon/cphalcon/pull/12861

sergeyklay:
@Fenikkusu  The `zephir_function_exists ` function  has been fix in the latest Zephir

phalcon/mvc/view/engine/volt/compiler.zep
```diff
@@ -2210,8 +2219,23 @@ class Compiler implements InjectionAwareInterface
...
 				let this->_autoescape = autoescape;
 			}
+
+            /**
+             * Configures the compiler to register PHP Functions automatically.
+             */
+            if isset options["includePhpFunctions"] {
+                let includePhpFunctions = options["includePhpFunctions"];
+                if typeof includePhpFunctions != "boolean" && typeof includePhpFunctions != "array" {
+                    throw new Exception("'includePhpFunctions' must be a bool or array value");
```
sergeyklay:
```zep
"The 'includePhpFunctions' option must be be either a bool or array"
```

phalcon/mvc/view/engine/volt/compiler.zep
```diff
@@ -2191,9 +2199,10 @@ class Compiler implements InjectionAwareInterface
 	{
 		var currentPath, intermediate, extended,
 			finalCompilation, blocks, extendedBlocks, name, block,
-			blockCompilation, localBlock, compilation, options, autoescape;
+			blockCompilation, localBlock, compilation, options, autoescape, includePhpFunctions;
 
 		let currentPath = this->_currentPath;
+        let includePhpFunctions = false;
```
sergeyklay: Could you please you the same indents in the whole file?

sergeyklay (changes requested):

sergeyklay:
@Fenikkusu ping

PR #12849, opened 2017-05-15
Volt Compiler: Remove exception when stat option is true, and when fi…

SachaMorard:
Hello!

- [x] Type: code quality

For production environnements, to improve a lot the volt performance, it's really interresting to work with `stat` option = true on Volt compiler. But the problem is that there is not an easy way to warmup Volt cache. Also, if you clear the cache folder, then you'll have many exceptions. I propose to remove this exception and  compile the file.

Thanks


sergeyklay:
Cc: @phalcon/collaborators @phalcon/framework-team

sergeyklay:
Thank you

SachaMorard:
Thank You !

PR #12919, opened 2017-06-21
3.2.1

sergeyklay:


stamster:
When we can expect 3.2.1 to be officially released?

stamster:
@sergeyklay when we can expect 3.2.1 to be officially released?

sergeyklay:
@stamster We'll try next week

Issue #12950, opened 2017-07-10
Paginator QueryBuilder not escaping rowcount column


dnapierata:
### Expected and Actual Behavior

When using the paginator with a `GROUP BY` and `HAVING` clause, the sql that is generated does not escape the aliased column and thus I am receiving a syntax error from my database (MSSQL 12.0). [https://github.com/phalcon/cphalcon/blob/master/phalcon/paginator/adapter/querybuilder.zep#L228](https://github.com/phalcon/cphalcon/blob/master/phalcon/paginator/adapter/querybuilder.zep#L228)

I would expect Line 228 to be 
```zep
let row = db->fetchOne("SELECT COUNT(*) as \"rowcount\" FROM (" .  sql["sql"] . ") as T1", Db::FETCH_ASSOC, sql["bind"]),
```
### Details

* Phalcon version: (`3.2.0`)
* PHP Version: (` 7.0.18-0ubuntu0.16.04.1`)
* Operating System: Linux
* Installation type:  installing via package manager
* Server: Apache
* Other related info (Database, table schema): MSSQL 12.0

sergeyklay:
Phalcon does not have build-in MSSQL dialect class. So it is most likely a new feature request for Incubator.

dnapierata:
I realize MSSQL isn't currently supported by phalcon but this fix is just to make the generated SQL more standards compliant and in fact it is already being done here [https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/model.zep#L1135](https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/model.zep#L1135)

I already have the dialect and adapter classes for sql server but this does not solve the issue I have posted above. I am just asking you add double quotes around `rowcount` so that the query builder paginator is more compatible with future db dialects.

Jurigag:
Can you do PR with fix to 3.2.x branch?

sergeyklay:
Fixed in the `3.2.x branch`. Feel free to open a new issue if the problem appears again. Thank you for contributing.

PR #12951, opened 2017-07-10
escape rowcount alias

dnapierata:
Hello!

* Link to issue: [#12950](https://github.com/phalcon/cphalcon/issues/12950)

Small description of change: Escaped row count aliases with double quotes

Thanks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # [3.2.2](https://github.com/phalcon/cphalcon/releases/tag/v3.2.2) (2017-XX-XX)
+- Added delimiters to escape reserverd words in `Phalcon\Paginator\Adapter\QueryBuilder` method `getPaginate` [#12950](https://github.com/phalcon/cphalcon/issues/12950)
 
 # [3.2.1](https://github.com/phalcon/cphalcon/releases/tag/v3.2.1) (2017-07-10)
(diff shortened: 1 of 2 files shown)


sergeyklay:
@dnapierata Could you please update the CHANGELOG.md

sergeyklay:
As you can see the tests failed. And yes, I'm not sure about escaping the PHQL

dnapierata:
I agree I don't think that is the correct way to escape when using PHQL. 
I used this as a reference:
[https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/model.zep#L1135](https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/model.zep#L1135)
 but it is not actally using PHQL, just raw SQL. 
What about using the built in escaping of PHQL by using brackets? 
Ex: `[rowcount]`

Jurigag:
Well the failing test is about PHQL, i guess it should be `[rowcount]` most likely yea.

dnapierata:
Any idea about the AppVeyor tests? It is getting a 404 when trying to download a php release

Jurigag:
Not sure, seems unrelated, also update changelog.md and squash commits.

Jurigag (approved):

sergeyklay:
@dnapierata Could you please create a small test which will cover code you have changed?

dnapierata:
The tests which are already written for the querybuilder paginator cover this block of code. The only true way to test this feature would be to set up a database that has `ROWCOUNT` as a reserved word such as MSSQL, OpenEdge, Rave, etc. I'm not well familiar enough in this testing environment with setting up this level of config.

sergeyklay:
Thank you

Issue #12957, opened 2017-07-12
[3.2.x] Paginator QueryBuilder with HAVING doesn't support multiple db services


dnapierata:
### Expected and Actual Behavior

When using the paginator querybuilder with a `HAVING` and with a model which uses a different db connection service. The paginator incorrectly assumes that the db service is `'db'` when performing a count of the total records. [https://github.com/phalcon/cphalcon/blob/3.2.x/phalcon/paginator/adapter/querybuilder.zep#L227](https://github.com/phalcon/cphalcon/blob/3.2.x/phalcon/paginator/adapter/querybuilder.zep#L227)

I expect that it would use the connection service of the model that the query was built for.
Example: 
```Zephir
let modelClass = builder->_models;
if typeof modelClass == "array" {
    let modelClass = array_values(modelClass)[0];
}
let model = new {modelClass}();
let dbService = model->getReadConnectionService();
let db = totalBuilder->getDI()->get(dbService);
```

### Details

* Phalcon version: (`3.2.1`)
* PHP Version: (`7.0.18`)
* Operating System: Ubuntu
* Installation type: Installing via package manager
* Server: Apache

Jurigag:
Good point, can you do PR with this code?

sergeyklay:
Could you please create a small test which will cover code you have changed?

sergeyklay:
@dnapierata 
I'm hope this fixed in the #12960 PR. Could you please create a small test which will cover code you have changed? Thank you for contributing.

PR #12960, opened 2017-07-13
Use db service from the model when paginating

dnapierata:
* Type: bug fix
* Link to issue: #12957

Small description of change: Use the db service from the model when paginating instead of assuming `'db'` as the service name.


Jurigag:
Also some test could be good.

Jurigag (approved):

sergeyklay:
Thank you

Issue #12842, opened 2017-05-10
Wrong metadata data types for PostgreSQL DB


tztztztz:
Method \Phalcon\Mvc\Model\MetaData::getDataTypes is returning wrong datatype for PostgreSQL database and column types double precision and real. 

It is returning 2 which wrongly corresponding to \Phalcon\Db\Column ::TYPE_VARCHAR instead of respectively: \Phalcon\Db\Column::TYPE_DOUBLE for double precision and \Phalcon\Db\Column::TYPE_FLOAT for real.

Example dump of print_r of array returned by getDataTypes($model), where $model is subclass of \Phalcon\Mvc\Model:

```
$dataTypes = $entity->getModelsMetaData()->getDataTypes($entity);
echo '<pre>'; print_r($dataTypes); exit;
Array
(
    [id] => 0
    [name] => 2
    [is_self_delivery] => 8
    [is_active_for_client] => 8
    [is_active_for_admin] => 8
    [default_delivery_price_netto] => 2
    [is_default] => 8
    [can_delivery_to_client] => 8
    [can_delivery_returns] => 8
    [php_manager_class] => 2
    [default_delivery_price_vat_pct] => 2
    [default_delivery_price_vat_val] => 2
    [default_delivery_price_brutto] => 2
)
```

Table structure dump from Postgresql version 9.5:

```
CREATE TABLE public.k3_delivery_method
(
  id integer NOT NULL DEFAULT nextval('k3_delivery_method_id_seq'::regclass),
  name character varying(128),
  is_self_delivery boolean NOT NULL DEFAULT false,
  is_active_for_client boolean NOT NULL DEFAULT false,
  is_active_for_admin boolean NOT NULL DEFAULT false,
  default_delivery_price_netto double precision NOT NULL,
  is_default boolean NOT NULL DEFAULT false,
  can_delivery_to_client boolean NOT NULL DEFAULT true,
  can_delivery_returns boolean NOT NULL DEFAULT false,
  php_manager_class character varying(255),
  default_delivery_price_vat_pct real NOT NULL DEFAULT 0,
  default_delivery_price_vat_val double precision NOT NULL DEFAULT 0,
  default_delivery_price_brutto double precision NOT NULL DEFAULT 0,
  CONSTRAINT k3_delivery_method_pkey PRIMARY KEY (id)
)
WITH (
  OIDS=FALSE
);
ALTER TABLE public.k3_delivery_method
  OWNER TO postgres;
```

System details:

[...]

PR #12962, opened 2017-07-13
Update postgresql.zep - added support for "double precision" and "real" data types for metadata

tztztztz:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/12842

Small description of change:

* removed check for `UNSIGNED` as PostgreSQL does not support it
* removed double check for `TIMESTAMP` type
* added support of `DOUBLE PRECISION` and `REAL` data types

Thanks


Jurigag:
Update changelog also.

sergeyklay:
Thank you

PR #12980, opened 2017-07-24
Internal cookies property is now always an array (fixes #12978)

### LINKED ISSUE #12978: Phalcon\Http\Response\Cookies->send() cause "The argument is not initialized or iterable"
I try use `Phalcon\Http\Response\Cookies`, separate from `Phalcon\Http\Response`. In my app "Cookies" object can contain cookie, or not. And i do not have way to check, is there exists at least one cookie in it, or not, before try to send it, because class do not provide method for this.
So if there are was not added at least one cookie, on calling `send()` method appear exception **"The argument is not initialized or iterable"**. 
I think that because `_cookies` variable in class defined without value and stay `null` if there are no cookie set (if i understand Zephir right). And we try to iterate through it in `send() `  method when `_cookies` is `null`

There are few ways to fix that:

1. Make `protected _cookies = [];` instead `protected _cookies;`
2. Check is `_cookies` not empty/iterable before try iterate through it
3. Provide for `Phalcon\Http\Response\Cookies` class method that allow to check, is there are added cookie (something about `isEmpty()`)

> **Expected behavior**

When there are no cookie set, do nothing

> **How to reproduce**

```php
(new Phalcon\Http\Response\Cookies())->send();
```
### Details

* Phalcon version: 3.1.2
* PHP Version: 7.1.4

SidRoberts:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/12978

Small description of change: Internal cookies property is now always an array

Thanks

### DIFF
--- a/phalcon/http/response/cookies.zep
+++ b/phalcon/http/response/cookies.zep
@@ -42,6 +42,11 @@ class Cookies implements CookiesInterface, InjectionAwareInterface
 
 	protected _cookies;
 
+	public function __construct()
+	{
+		let this->_cookies = [];
+	}
+
 	/**
 	 * Sets the dependency injector
 	 */


phalcon/http/response/cookies.zep
```diff
@@ -40,7 +40,7 @@ class Cookies implements CookiesInterface, InjectionAwareInterface
 
 	protected _useEncryption = true;
 
-	protected _cookies;
+	protected _cookies = [];
```
sergeyklay: @SidRoberts Actually this will negatively impact performance in ZendEngine 2.x (PHP 5.x). Could you please refactor this to initialize property at construct time?

Refs: https://github.com/phalcon/zephir/issues/1255

sergeyklay (changes requested):

Jurigag:
Also update changelog and maybe add some test?

sergeyklay (approved):

PR #13026, opened 2017-08-19
Model::setSnapshotData() now properly sets the old snapshot.

SidRoberts:
* Type: bug fix

`Model::setSnapshotData()` now properly sets the old snapshot.


phalcon/mvc/model.zep
```diff
@@ -3847,7 +3847,7 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 			let snapshot = data;
 		}
 
-		let this->_oldSnapshot = snapshot;
+		let this->_oldSnapshot = this->_snapshot;
```
sergeyklay: How about 3.2.x branch?

SidRoberts: Done.

Jurigag: I think you need to check if it's not empty otherwise set it to actually snapshot value, also im not sure if other things will keep working. oldSnapshot in model is just snapshot which was before save/update. setSnapshotData is used when findFirst, so pretty much when doing Model::findFirst, both snapshots should be the same, this change cause it to set it to empty value, and https://github.com/SidRoberts/cphalcon/blob/99235ca1b4d67ba3e40f955c8d798137d4b31081/phalcon/mvc/model.zep#L4077 to return true which isn't true really.

sergeyklay:
@SidRoberts Could you please take a look

Issue #12288, opened 2016-10-04
[4.0.0]Change catch Exception to Throwable, drop PHP 5 support


Jurigag:
- Change all `catch \Exception` to `catch \Throwable` in `Phalcon\Dispatcher`, `Phalcon\Micro`, `Phalcon\Pdo\Postgresql`
- This means drop php 5 support

PR #13029, opened 2017-08-21
Change catch Exception to Throwable

sergeyklay:
Refs: #12288

### DIFF
--- a/phalcon/annotations/exception.zep
+++ b/phalcon/annotations/exception.zep
@@ -24,6 +24,6 @@ namespace Phalcon\Annotations;
  *
  * Class for exceptions thrown by Phalcon\Annotations
  */
-class Exception extends \Exception
+class Exception extends \Phalcon\Exception
 {
 }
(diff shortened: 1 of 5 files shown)


Jurigag:
Still i would like to see option to handlError diffrently than handleException? What you think? Sometimes people would want this.

sergeyklay:
What do you think about something like that:
```php
$dispatcher->setExceptionHandler($callable);
```

Jurigag:
Well what i mean is to add to dispatcher event called beforeError and if we have error - fire it first. Right not we have fired beforeException on error too, not sure if this is what someone would like, i think we need `dispatch:beforeException`, `dispatch:beforeThrowable` and `dispatch:beforeError`

Jurigag:
https://github.com/phalcon/cphalcon/pull/12292 like this

sergeyklay:
@Jurigag Could you open nfr?

PR #13048, opened 2017-08-29
3.2.x Validate date NULL, make required.

renatogabrielbr:
Its possible when to send null or empty string dont ask the presence, date validator i don't think must have PresenceOf. 

https://github.com/phalcon/cphalcon/blob/master/phalcon/validation/validator/date.zep

Something like

```
if empty value {
      return
}
```

Or 

```
if is_null (value) {
      return
}
```

```

$post['time'] = NULL;

$var = new Text('time');
$var
->setFilters(['string','null'])
 ->addValidator(new Date(['format' => 'Y-m-d H:i:s', 'message' => $this->translate->_('Wrong')]));
```

Return asking a valid date. this should be done by the PresenceOf validator.

### DIFF
--- a/CHANGELOG-3.2.md
+++ b/CHANGELOG-3.2.md
@@ -1,3 +1,5 @@
+# [3.2.3](https://github.com/phalcon/cphalcon/releases/tag/v3.2.3) (2017-XX-XX)
+
 # [3.2.2](https://github.com/phalcon/cphalcon/releases/tag/v3.2.2) (2017-08-14)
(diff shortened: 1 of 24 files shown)


sergeyklay:
We don't add changes directly to stable branches or `master`. Could you please send this PR to the `3.2.x` branch?

tidytrax:
Sorry, how i do that ?

sergeyklay:
First of all lets clarify. Do you want report an issue or submit a pull request?

Jurigag:
Just add `allowEmpty => true` to options....

tidytrax:
Thanks works fine.

PR #13070, opened 2017-09-14
Fixed `Phalcon\Mvc\ModelInterface` to include `getModelsMetaData()`.

virgofx:
Fixed `Phalcon\Mvc\ModelInterface` to include `getModelsMetaData()`. This is important for static analysis as the ModelInterface is used as the interface parameter for various plugins and user components.

### DIFF
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -20,32 +20,31 @@
 namespace Phalcon\Mvc;
 
 use Phalcon\Di;
+use Phalcon\Db\AdapterInterface;
 use Phalcon\Db\Column;
+use Phalcon\Db\DialectInterface;
 use Phalcon\Db\RawValue;
 use Phalcon\DiInterface;
-use Phalcon\Mvc\Model\Message;
-use Phalcon\Mvc\Model\ResultInterface;
 use Phalcon\Di\InjectionAwareInterface;
-use Phalcon\Mvc\Model\ManagerInterface;
-use Phalcon\Mvc\Model\MetaDataInterface;
+use Phalcon\Events\ManagerInterface as EventsManagerInterface;
+use Phalcon\Mvc\Model\BehaviorInterface;
 use Phalcon\Mvc\Model\Criteria;
-use Phalcon\Db\AdapterInterface;
-use Phalcon\Db\DialectInterface;
 use Phalcon\Mvc\Model\CriteriaInterface;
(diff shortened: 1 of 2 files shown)


virgofx:
ping @sergeyklay

sergeyklay:
I'll cherry pick this to the 4.0.x a bit later if the need arises. Thank you for contributing. Could you please return back the changes in the `phalcon/mvc/model.zep` related to import classes.

virgofx:
The 2 imports that were removed were

```
use Phalcon\Validation\Message\Group;
use Phalcon\Validation\Message\Group as ValidationMessageGroup;
```

Not used + duplicated. Others were reordered alphabetically. You want these back in even though not used?

sergeyklay:
Actually I prefer to keep it as is (unordered) and just add new ones to decrease diff.

virgofx:
It's not about the diff.... It's about correctness. 2 of the imports are not used and shouldn't be imported.

Jurigag:
Btw, shouldn't this change aim 4.0.x?

sergeyklay:
@virgofx I see 14 additions and 17 deletions. I wold like to see 2 deletions.

sergeyklay:
@virgofx Removed not used imports in the  #13089. Could you please rebase

virgofx:
Rebased .. but left alphabetized as that's how coding standard is. The reason for the extra deletion is because there was another duplicate `use Phalcon\Mvc\Model\Message` which was missed in your last fix. The reason you missed this is because imports should always be optimized/alphabetically. Please merge.

sergeyklay:
We can't change interface in minor release. Thus I added this to the 4.0.0 milestone.

sergeyklay:
Also just to you know there is no benefit from ordering imports in code which will be compiled and optimized before release

virgofx:
I agree compiled code doesn't care; however, the coding standard and those contributing all follow best practices. No reason not to keep things inline best practice. If this is for the 4.0 milestone ... will you fix in a separate PR then with the rest of the interfaces?

virgofx:
Best practice would also have moving a lot of the constants from Classes into Interfaces as well.

virgofx:
I could but is 4.0 selectively cherry picking commits? Seems like it would miss core functionality from 3.x?  Just making sure other 3.x stuff is also in 4.0?

sergeyklay:
The 4.0.x branch is actual now.

PR #13073, opened 2017-09-18
fix the bugs for log's formatter

qzfzz:
Hello!

Type: bug fix

Fix the bugs for phalcon\logger\formatter.php when calling the method format with the array param context it not produce the result expected!

Thanks


andrew-demb:
Formatter working correctly according to PSR-3 specification.

See link http://www.php-fig.org/psr/psr-3/#12-message

PR #12564, opened 2017-01-23
Clean new PR for 12409 add transaction in model::find()

chilimatic:
Hello!

* Type: new feature
* Link to issue: [#12409](https://github.com/phalcon/cphalcon/pull/12409)

**In raising this pull request, I confirm the following (please check boxes):**

- [X] I have read and understood the [Contributing Guidelines](https://github.com/phalcon/cphalcon/blob/master/CONTRIBUTING.md)?
- [X] I have checked that another pull request for this purpose does not exist.
- [X] I wrote some tests for this PR.

**Adds transaction to model::find() so you can query inserted data before comiting**

***Main usecases***
 * complex insert processes (you may need to check the state inside of the transaction or add a model to a transaction)
 * unitesting -> start a transaction at the beginning and rollback at the end so you don't polute your test database

***Minor change*** 
 * DRY for selectWrite and selectRead transaction inside of query.zep
 * Adds general unitests for query.zep

PR #12409, opened 2016-11-17
added transaction parameter for Model::find() 

chilimatic:
Thank you for wanting to contribute to Phalcon.

Before you submit a Pull Request, please bear in mind that PRs must be made on a
version branch (eg. `3.0.x`). `master` marks the latest stable release so PRs
targeting `master` will be **rejected**.

Please also check the following points are being met:

- [x] Make sure that the PR passes in Travis CI to make the process more efficient.
- [x] Only use tabs for indentation.
- [x] if needed, rebase to the proper branch before submitting your pull request.
      If it doesn't merge cleanly with master you may be asked to rebase your changes.
- [x] Don't put submodule updates in your pull request unless they are to landed commits.
- [x] Add tests relevant to the fixed bug or new feature. See our testing guide for
      more information.
- [x] Phalcon 2 is written in Zephir, please do not submit commits that modify C generated
      files directly or those whose functionality/fixes are implemented in the C
      programming language
- [x] Remove any change to ext/kernel / *.zep.c / *.zep.h files before submitting the PR

so unitests can be executed inside one transaction

PR #13097, opened 2017-09-28
Followup for #12564 added transaction to param in model::find() and model::findFirst() 

chilimatic:
Hello!

* Link to issue: #12409, #12564

Small description of change:

This change is to achieve the following

In Unit-testing one usual procedure is to create a transaction for a TestCase and after the execution rollback this change. 

Which means there is a "long transaction" going on and we want to check if the transaction actually is writing the correct values in the database. But the static context of model::find / model::findFirst does not get the transaction passed.

To achieve that passing a new index "transaction" containing an transaction in the $param array seams to be a viable solution. 

Another use-case can be having multiple transactions and wanting to look inside. 

Just to be precise -> ofc we could just get the connection and query manually inside of it, but I want to be as close to the framework as possible when testing certain services hence I want to use the model. 

I took your reviews and advice from the last attempt and used the params to not break the interfaces / change method signatures.  

the logic is implemented as follows: 

if you pass in a transaction via param in the find/findFirst method. It will set the transaction to query object and this transaction will be used for all queries and executions within this query context.

This should not affect any standard behaviour but still allow me to finally achieve cleaner unittests for my usecase.

I still owe you some unitests, but please look through it first so it's not after several hours of refactoring and cleanups that this gets rejected.


chilimatic:
@sergeyklay any feedback on the functional code implementation ? if it looks okay to you we can move the semantics and syntaxation discussion and I will start writing tests for it.

phalcon/mvc/model/query.zep
```diff
@@ -2544,21 +2584,14 @@ class Query implements QueryInterface, InjectionAwareInterface
...
-			} else {
-				let connection = model->getReadConnection();
-			}
+			let connection = this->getReadConnection(model, intermediate, bindParams, bindTypes);
 
-			// More than one type of connection is not allowed
-			let connectionTypes[connection->getType()] = true;
-			if count(connectionTypes) == 2 {
-				throw new Exception("Cannot use models of different database systems in the same query");
+			if (connection !== null) {
```
sergeyklay:
```zep
if typeof connection == "object"
```

sergeyklay (changes requested):

chilimatic:
I did a force push to add the ticket reference correctly to all the commits

sergeyklay (approved):
LGTM

chilimatic:
@sergeyklay thx,  I will write the unitests now :)

sergeyklay:
@chilimatic Could you please use 3.3.x branch for that

chilimatic:
@sergeyklay ofc, rebase my changes onto 3.3.x ?

sergeyklay:
The 3.3.x branch is upstream for now

chilimatic:
I will try to finish this asap. I just have some project deadlines comming up

chilimatic:
added new issue #13226  and PR #13227

Issue #13099, opened 2017-09-29
Can't create nullable timestamps


1RV34:
### Expected and Actual Behavior

I'm trying to use the `Column` class to create tables in my database. In these tables I try to create a timestamp column that is nullable using the following code:
```php
new Column('deadline', 'type' => Column::TYPE_TIMESTAMP, 'notNull' => false, 'default' => null)
```
I would expect to have created a nullable timestamp column with its default value being `null`, however instead a timestamp column would be created with `current_timestamp` as default together with the `on update current_timestamp` property, all following timestamp columns will have default `'0000-00-00 00:00:00'`.

### Details

* Phalcon version: 3.2.1
* PHP Version: 7.1.9
* Operating System: ubuntu 16.04.1
* Installation type: Compiling from source
* Zephir version: 0.9.9
* Server: Apache
* Other related info:
    Generated query will have the following in it:
    ```sql
    CREATE TABLE `table` (
        `timestamp1` timestamp,
        `timestamp2` timestamp
    );
    ```
    which will result in a table with columns like:
    ```sql
    CREATE TABLE `table` (
        `timestamp1` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURENT_TIMESTAMP,
        `timestamp2` timestamp NOT NULL DEFAULT '0000-00-00 00:00:00'
    );
    ```
    Instead it should generate:
    ```sql
    CREATE TABLE `table` (
        `timestamp1` timestamp NULL DEFAULT NULL,
        `timestamp2` timestamp NULL DEFAULT NULL
    );
    ```

sergeyklay:
As you can see tests failed

> [`Failures: 40`](https://travis-ci.org/phalcon/cphalcon/jobs/281230808#L2723)

1RV34:
I saw, working on updating the tests.

1RV34:
Do you think this fix could be included with **3.2.x**, since it's a bugfix not a feature? (Current Phalcon system doesn't respond to `'notNull' => false` correctly and causes issues. It's causing our applications to break.)

1RV34:
Sorry, wrong button.

PR #13100, opened 2017-09-29
Explicitly define nullable columns

1RV34:
Hello!

* Type: bug fix
* Link to issue: #13099

- [x] I have read and understood the [Contributing Guidelines](https://github.com/phalcon/cphalcon/blob/master/CONTRIBUTING.md).

Small description of change:

Explicitly define nullable columns. This fixes an issue I had with timestamp columns. By default when creating a timestamp column without explicitly telling it's allowed to be null and defaulted to null it will automatically give it an "on update current_timestamp" property.

To clarify: In my code I used the following:
```php
new Column('deadline', [
    'type' => Column::TYPE_TIMESTAMP,
    'notNull' => false,
    'default' => null
])
```
and instead it would create a timestamp with `CURRENT_TIMESTAMP` as default and the "on update current_timestamp" property.

Thanks

Ricardo Vermeltfoort


1RV34:
1. Not exactly sure how to properly write a test for this.
2. SQLite seems to default to `null` but i'm not 100% sure on this as I don't use it much at all.

1RV34:
I just saw the test that checks this and will update it now.

1RV34:
Still updating PostgreSQL tests.

1RV34:
Not exactly sure why the tests still fail, it's only the PHP 5.6 build, do you know?

sergeyklay:
@1RV34 I've changed the base branch from `3.2.x`  to `3.3.x` and going to merge. There is some minor things. Could you please update change log also squash commits into one to keep history clean. Thank you.

niden (approved):

niden:
Thank you!

PR #12209, opened 2016-09-09
Fixed segfault in PHP5.6 dispatching, prevented double dispatch

virgofx:
Fixes the long-standing Mvc Dispatcher bugs in PHP5.6.  [#12154](https://github.com/phalcon/cphalcon/issues/12154), [#11819](https://github.com/phalcon/cphalcon/issues/11819)
- Required Zephir fix for try/catch block fix required
- Corresponding documentation updates included in separate PR
- Test suite updated to use Codeception along with addition of full 100% test coverage of dispatcher functionality.

image

Issue #12154, opened 2016-08-20
[3.0.0]Seg fault on php5 when forwarding from beforeException


Jurigag:
### Expected and Actual Behavior

Should return test, but it doesn't, there is just seg fault when accessing just index page(`/`)

**On php7 it works though.**

Backtrace - http://pastebin.com/jD5GQJaJ

Github repo - https://github.com/Jurigag/phalcon-segfault-before-forward
### Details
- Phalcon version:

```
Web framework delivered as a C-extension for PHP
phalcon => enabled
Author => Phalcon Team and contributors
Version => 3.0.1
Build Date => Aug 20 2016 15:06:39
Powered by Zephir => Version 0.9.4a-dev-121e9b4bf1

Directive => Local Value => Master Value
phalcon.db.escape_identifiers => On => On
phalcon.db.force_casting => Off => Off
phalcon.orm.events => On => On
phalcon.orm.virtual_foreign_keys => On => On
phalcon.orm.column_renaming => On => On
phalcon.orm.not_null_validations => On => On
phalcon.orm.exception_on_failed_save => Off => Off
phalcon.orm.enable_literals => On => On
phalcon.orm.late_state_binding => Off => Off
phalcon.orm.enable_implicit_joins => On => On
phalcon.orm.cast_on_hydrate => Off => Off
phalcon.orm.ignore_unknown_columns => Off => Off
```
- PHP Version:

```
PHP 5.6.19 (cli) (built: Apr 24 2016 12:31:51) 
Copyright (c) 1997-2016 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.6-dev, Copyright (c) 1999-2016, by Zend Technologies
```
- Operating System: Ubuntu 14.04
- Installation type: zephir
- Zephir version (if any): Version 0.9.4a-dev-121e9b4bf1
- Server: Apache

virgofx:
Likely related to: #11819  -- Can we mark both of these as relevant for 3.0.2 milestone as this has been an issue for awhile.

Jurigag:
Yea but what is weird it works on php 7 though.

virgofx:
Confirmed issue as well.

*Backtrace*
http://vpaste.net/wGczw

virgofx:
There are actually 2 bugs. 

1. A return `false` when forwarding dispatcher in exception will cause sigfault
[...]

Issue #11819, opened 2016-05-23
[2.1.x] Dispatcher doesn't forward when handling exception


virgofx:
Code works in 2.0.x. It may have to do with the creation of the new `_dispatch()` method in 2.1.x and multiple levels of try/catch which do not work in zephir.

Pseudo code:

**bootstrap**
```php
$eventsManager = new EventsManager();
$eventsManager->attach('dispatch:beforeException', function($event, $dispatcher, $exception) {
    $dispatcher->forward([
        'controller' => 'test',
        'action' => 'second'
    ]);
    // Prevent the exception from bubbling
    return false;
}

$dispatcher = new \Phalcon\Mvc\Dispatcher();
$dispatcher->setEventsManager($eventsManager);

//...
$di->setShared('dispatcher', $dispatcher);
```

**TestController.php**
```php
class TestController extends \Phalcon\Mvc\Controller
{
    public function indexAction()
    {
        throw new \Exception('whups bad controller');
    }

    public function secondAction() 
    {
        echo 'I should be displayed';
        exit;
    }
}
```

temuri416:
Hi @virgofx, is there any progress on this issue?

Thanks!

virgofx:
@temuri416 We have it fixed; however, still waiting on a core bug in Zephir land which causes a segfault in the updated tests. In the meantime, I've had to override the dispatcher manually which fixes this in our production boxes - If you need this snippet email me directly and I can share with you.

temuri416:
@andresgutierrez Can we please have your attention? We need your help to get the issue fixed. Thank you!

temuri416:
Verified, working as expected! This is an enormous improvement, thanks a lot.

mladen7:
A bit late to the party but having the same problem at the moment... Using php 5.6.10 and phalcon 3.2.4 version, i cannot forward either from the beforeException or the beforeExecuteRoute methods.. When i try to forward from the beforeExecuteRoute, i trigger a dispatch cycle loop eventually being picked up by the beforeException, which then also does not manage to forward but returns the index page.. I used the basic INVO tutorial available on the phalcon docs website.. Relevant code follows:
bootstrap
`$di->set(
    'dispatcher',
    function () use ($di) {
        // Create an events manager
        $eventsManager = new EventsManager();
        // Listen for events produced in the dispatcher using the Security plugin
        $eventsManager->attach(
            'dispatch:beforeExecuteRoute',
            new \Plugins\SecurityPlugin()
[...]

PR #13112, opened 2017-10-07
Fixed Dispatcher::dispatch() to ensure proper flow for all event handling, forwarding, exceptions, bubbling and performance improvements. Fixes #12931

### LINKED ISSUE #12931: [BUG] Action forwarding is not working if Exception is handled in the Controller's initialize() phase
Hi,

I use this approach to handle exceptions thrown in Controller Actions. Nothing fancy here:

```php
$eventsManager->attach('dispatch:beforeException', function($event, $dispatcher, $exception) {
    // Forward to Error controller:
    $dispatcher->forward([
        'controller' => 'error',
        'action' => 'oops',
        'params' => null,
    ]);
}
```

Everything works fine if an Exception is thrown inside Controller's Action, i.e. past this line:

https://github.com/phalcon/cphalcon/blob/master/phalcon/dispatcher.zep#L654

The problem is that if an Exception is thrown inside Controller's `initialize()`, then **forwarding code above does not work** (it does get executed).

Looking at the code I see something that looks like an oversight to me. This is the code that handles Exception in the named action ([link](https://github.com/phalcon/cphalcon/blob/master/phalcon/dispatcher.zep#L653)):

```php
try {
    // We update the latest value produced by the latest handler
    let this->_returnedValue = this->callActionMethod(handler, actionMethod, params);
} catch \Exception, e {
    if this->{"_handleException"}(e) === false {
        if this->_finished === false {
            continue;
        }
    } else {
        throw e;
    }
}
```

As you can see, an Exception is caught and `_handleException` is fired inside the dispatch loop, thus giving Dispatcher's `forward()` a chance  to affect execution in subsequent iteration.

And this is how Controller's `initialize` is called ([link](https://github.com/phalcon/cphalcon/blob/master/phalcon/dispatcher.zep#L600)):

```php
if wasFresh === true {
    if method_exists(handler, "initialize") {
        handler->initialize();
    }
    ...
}

```

So, if an Exception is thrown in `initialize`, the error bubbles up and out of the dispatch loop.

So, shouldn't there be a try\catch around the initialize() call as well?

Thanks!

### Details

* Phalcon version: 3.2.x
* PHP Version: (`php -v`) 7.1
* Operating System: Unix
* Installation type: Compiling from source
* Zephir version (if any): Latest

virgofx:
This is a an update of previously closed PR #12209.  It resolves all dispatch issues, adds additional notes in the code for future deprecations in 4.0. Various performance improvements in the resulting dispatch process as result of variable caching and dropping the global try/catch which previously hooked into the RVAL for every statement previously.

Fixes current issue: #12931

Related to (and ensures fixes for) previous issues:
#12154
#11819

cc @andresgutierrez
cc @temuri416 
cc @sergeyklay 
cc @niden 

```
Unit Tests (69) -------------------------------------------------------------------------------
✔ DispatcherAfterDispatchLoopTest: After dispatch loop forward (0.03s)
✔ DispatcherAfterDispatchLoopTest: After dispatch loop return false (0.00s)
✔ DispatcherAfterDispatchLoopTest: After dispatch loop with before exception returning false (0.00s)
✔ DispatcherAfterDispatchLoopTest: After dispatch loop with before exception bubble (0.00s)
✔ DispatcherAfterDispatchLoopTest: After dispatch loop with before exception forward once (0.00s)
✔ DispatcherAfterDispatchTest: After dispatch forward once (0.00s)
✔ DispatcherAfterDispatchTest: After dispatch return false (0.00s)
✔ DispatcherAfterDispatchTest: After dispatch with before exception returning false (0.00s)
✔ DispatcherAfterDispatchTest: After dispatch with before exception bubble (0.00s)
✔ DispatcherAfterDispatchTest: After dispatch with before exception forward once (0.00s)
✔ DispatcherAfterExecuteRouteMethodTest: After execute route forward once (0.00s)
✔ DispatcherAfterExecuteRouteMethodTest: After execute route return false (0.00s)
✔ DispatcherAfterExecuteRouteMethodTest: After execute route with before exception returning false (0.00s)
✔ DispatcherAfterExecuteRouteMethodTest: After execute route with before exception bubble (0.00s)
✔ DispatcherAfterExecuteRouteMethodTest: After execute route with before exception forward once (0.00s)
✔ DispatcherAfterExecuteRouteTest: After execute route forward once (0.00s)
✔ DispatcherAfterExecuteRouteTest: After execute route return false (0.00s)
✔ DispatcherAfterExecuteRouteTest: After execute route with before exception returning false (0.00s)
✔ DispatcherAfterExecuteRouteTest: After execute route with before exception bubble (0.00s)
✔ DispatcherAfterExecuteRouteTest: After execute route with before exception forward once (0.00s)
✔ DispatcherAfterInitializeTest: After initialize forward once (0.00s)
✔ DispatcherAfterInitializeTest: After initialize return false (0.00s)
✔ DispatcherAfterInitializeTest: After initialize with before exception returning false (0.00s)
✔ DispatcherAfterInitializeTest: After initialize with before exception bubble (0.00s)
✔ DispatcherAfterInitializeTest: After initialize with before exception forward once (0.00s)
✔ DispatcherBeforeDispatchLoopTest: Before dispatch loop forward (0.00s)
✔ DispatcherBeforeDispatchLoopTest: Before dispatch loop return false (0.00s)
✔ DispatcherBeforeDispatchLoopTest: Before dispatch loop with before exception returning false (0.00s)
✔ DispatcherBeforeDispatchLoopTest: Before dispatch loop with before exception bubble (0.00s)
✔ DispatcherBeforeDispatchLoopTest: Before dispatch loop with before exception forward (0.00s)
✔ DispatcherBeforeDispatchTest: Before dispatch forward once (0.00s)
✔ DispatcherBeforeDispatchTest: Before dispatch return false (0.00s)
✔ DispatcherBeforeDispatchTest: Before dispatch with before exception returning false (0.00s)
✔ DispatcherBeforeDispatchTest: Before dispatch with before exception bubble (0.00s)
✔ DispatcherBeforeDispatchTest: Before dispatch with before exception forward once (0.00s)
✔ DispatcherBeforeExecuteRouteMethodTest: Before execute route forward once (0.00s)
✔ DispatcherBeforeExecuteRouteMethodTest: Before execute route return false (0.00s)
✔ DispatcherBeforeExecuteRouteMethodTest: Before execute route with before exception returning false (0.00s)
✔ DispatcherBeforeExecuteRouteMethodTest: Before execute route with before exception bubble (0.00s)
✔ DispatcherBeforeExecuteRouteMethodTest: Before execute route with before exception forward (0.00s)
✔ DispatcherBeforeExecuteRouteTest: Before execute route forward once (0.00s)
✔ DispatcherBeforeExecuteRouteTest: Before execute route return false (0.00s)
✔ DispatcherBeforeExecuteRouteTest: Before execute route with before exception returning false (0.00s)
✔ DispatcherBeforeExecuteRouteTest: Before execute route with before exception bubble (0.00s)
✔ DispatcherBeforeExecuteRouteTest: Before execute route with before exception forward once (0.00s)
✔ DispatcherInitalizeMethodTest: Initialize forward (0.00s)
✔ DispatcherInitalizeMethodTest: Initialize return false (0.00s)
✔ DispatcherInitalizeMethodTest: Initialize with before exception returning false (0.00s)
✔ DispatcherInitalizeMethodTest: Initialize with before exception bubble (0.00s)
✔ DispatcherInitalizeMethodTest: Initialize with before exception forward (0.00s)
✔ DispatcherTest: Default dispatch loop events (0.00s)
✔ DispatcherTest: Default dispatch loop events with no handlers (0.00s)
✔ DispatcherTest: Controller action local forward (0.00s)
✔ DispatcherTest: Controller action external forward (0.00s)
✔ DispatcherTest: Controller action return value string (0.00s)
✔ DispatcherTest: Controller action return value int (0.00s)
✔ DispatcherTest: Params and return value (0.00s)
✔ DispatcherTest: Cyclical routing (0.00s)
✔ DispatcherTest: Handler not found (0.05s)
✔ DispatcherTest: Handler invalid (0.00s)
✔ DispatcherTest: Handler action not found (0.00s)
✔ DispatcherTest: Last handler (0.00s)
✔ DispatcherTest: Last handler forward (0.00s)
✔ DispatcherTest: No namespaces (0.00s)
✔ DispatcherTest: Mixing namespace forward (0.00s)
✔ DispatcherTest: Default resolve (0.02s)
✔ DispatcherTest: Manual call action (0.00s)
✔ DispatcherTest: Last handler exception forward (0.00s)
✔ DispatcherTest: Exception in before exception (0.00s)
-------------------------------------------------------------------------------------------

Time: 2.23 seconds, Memory: 12.00MB

OK (69 tests, 173 assertions)
```


sergeyklay:
Good job. Could you deal with failed test
https://github.com/phalcon/cphalcon/blob/master/unit-tests/DispatcherMvcEventsTest.php#L206-L209

virgofx:
For clarity --- the test expected value is marked in green and the new behavior is in red (Wording backwards)...

The reason that failed was because of a nuance in the old test. Specifically, it did not contain an **initialize()** in the controller in the old case and thus it should never fire the `afterInitialize`.  I can see this both ways ... the change was made to interpret the following logic:

* Event should be thrown right before/after the corresponding action is about to happen. E.g. in the case of the view, if it's disabled the view events are not thrown.  

My original thought is that the existing behavior is incorrect; however, I would be compelled to add a new test case to handle this. E.g. If the use case was to somehow globally instantiate controllers from this method via some common functionality.

In the case of the dispatcher, it depends how you read the **afterInitialize** event. Do you:

A) interpret *initialize* as being specific to initializing via the `initialize()` method ... or
B) refer to it as being general such that the event should be fired in it's relative location even without a direct call into the `initialize()` - e.g. The controller is now initialized.


Side note --- there is a separate note in the code. The whole [ initialize/afterInitialize ] block is in the wrong location. The beforeExecuteRoute should never be executed before the initialize; however, this was a bug in the original code/documentation. I left it in place currently to not break B/C; however, it would be worth moving this section such that the `initialize` handles certain constructor/core controller logic that is outside of the main action portion.

sergeyklay:
Yes, the test expected value is marked in green and the new behavior is in red . I just copied phpunit output from the terminal. Well, could you just remove wrong test?

virgofx:
@sergeyklay I updated original comment with more info and questions. It is removed as noted now (hence passing); however, do you have thoughts on [A] vs. [B] above? I'm open to reverting back to the [B] model as I could see there be a global use case for performing actions from an eventhook globally in controllers regardless of the presence of `initialize()`

sergeyklay:
Actually I prefer [B] but in general your PR look really great. I will keep this open for a day or two to allow other contributors a chance to speak up, and to take a fresh look at it later. But at the moment it seems good to me.

cc: @phalcon/core-team

sergeyklay (approved):

niden:
This looks great. I would also prefer [B]. 

Great work buddy, thank you so much for the contribution!

virgofx:
Modified to include option [B] above. This means:
* Patch is 100% backwards compatible (with proper bugfixes only)
* The legacy `unit-tests` that failed now passes (even though it doesn't exist anymore). Replaced by the new suite and specifically one of the core dispatcher tests: https://github.com/virgofx/cphalcon/blob/da5f3e0736263ed8eba7505c11c644286f90aee3/tests/unit/Mvc/Dispatcher/DispatcherTest.php#L93-L104
* A few documentation updates forthcoming to include some things that are not able to fix in this version.  (For example returning `false` does not work when there are more event listeners. Currently, the event must be stopped. Need to outline how to do this in the documentation for 3rd party applications/libraries that may publish plugins that hook into events and need to stop dispatch actions. The only sure fire way would be to return false AND stop the event). This could be potentially reworked for 4.0.

niden:
For 3.x we should keep things the way they are. If I need events to stop (subsequent listeners do not fire) we will need to do `$event->stop()` and `return false`. This has to be documented for 3.x

For 4.x this behavior needs to change. Simply `return false` should suffice to stop the execution of subsequent listeners to the same event.

The above will be OK behavior so long as we ensure that we highlight it in the documentation.

niden:
This looks good to me @sergeyklay whenever you get a chance hit the magic button

sergeyklay:
Thank you!

virgofx:
@sergeyklay What's the 4.0 plan? Is 3.2.x getting merged up to 4.0 currently? Should I submit a new PR against the **current** 4.0 branch?

sergeyklay:
We're plan to release 3.2.3 in next days and start 3.3.0 branch. You can cherry-pick these changes to 4.0.x branch or leave this job for me. I'll do this later (but can't say when exactly).

virgofx:
Doesn't matter to me ... I can just help with the 4.0 changes related to this feature. Is the 4.0 branch working/building and is it PHP7.x only ?  What would you prefer I do ?

sergeyklay:
Yes, we'll drop support of the Zend Engine 2.x from Phalcon 4.x.

temuri416:
@virgofx @sergeyklay 

It looks like this patch needs a small review.

It broke an existing behaviour: after forwarding to an action within the same controller - its `initialize()` method does not get called after forward.

However, when forwarding to a different controller, controller's `initialize()` does get invoked again.

```php
<?php

use Phalcon\Mvc\Controller;

class IndexController extends Controller
{
    public function initialize()
    {
        $this->view->location = $this->di->get('dispatcher')->getHandlerClass();
        $this->view->action = $this->di->get('dispatcher')->getActionName();
    }

    public function indexAction()
    {
        /**
         * Before #13112 this forward will result in "Your location: IndexController/welcome".
         *
         * After #13112 this forward will result in "Your location: IndexController/index".
         *
         * @var IndexController
         */
        $this->di['dispatcher']->forward([
            'controller' => 'index',
            'action' => 'welcome',
        ]);
    }

    public function welcomeAction()
    {
    }
}
```

This line seems to be the key: https://github.com/phalcon/cphalcon/blob/v3.2.3/phalcon/dispatcher.zep#L520.

Current controller is no longer considered a "fresh" instance.

Code to reproduce is attached.
[issue-13112.zip](https://github.com/phalcon/cphalcon/files/1418972/issue-13112.zip)

Thanks!

sergeyklay:
@virgofx Could you please take a look

virgofx:
@temuri416 That's the correct behavior. The initialize() is only called ONCE ever per controller. It's more of a static onConstruct() .. Right now it maps to how the models were done with initialize as well in the sense that 

It's a TERRIBLE implementation; however. I made tons of notes in the code. It doesn't make sense in it's current location being called AFTER beforeExecuteRoute().

Also .. this is how things were in the past (although there were other bugs that may have blocked the intended behavior and thus it may appear different)

The next version really needs to fix the wording. "Initialize" is confusing:

1) Next version **WILL** move that logic **BEFORE** the `beforeExecuteRoute()` handlers.

I believe the `initialize()` is correct in that it should only be executed once per Dispatcher/Di setup. It's a one-time initialization --- stuff like doing ACL for controllers is where this should happen.   If you need something to happen EVERY time in some sort of cyclical cycle -- that should happen in the `beforeExecuteRoute`

temuri416:
ok, no big deal. I can move my logic to `beforeExecuteRoute`.

sergeyklay:
@virgofx These changes already in 3.3.x branch. I'll deal with 4.0.0 a bit later

PR #13120, opened 2017-10-18
Fix get relation when new array

rudiservo:
Hello!

* Type: bug fix
* Link to issue: none created

Small description of change:

When working with a model and you set a relation with an Array of objects that might not be saved but you want to work on, if you try to call that relation the model will not try and return a ResultSet

i.e.
$artist->albums = [$a1, $a2];

Thanks


rudiservo:
Having issues with some unit tests, i'll be debuging for a while.

Issue #13121, opened 2017-10-18
Dispatcher's params override any setVar()


dugwood:
### Expected and Actual Behavior

#13046 introduce a fix and a breaking change: params are now send to `setVars()`, erasing previously set variables. Works great, but:
https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/application.zep#L351
=> injects dispatcher's params, and erase previously set variables in controllers.

### Before 3.2.3
Url: /test-page/1234
```php
$router->add('/test-page/{bar:[0-9]+}', ['controller' => 'MyController', 'action' => 'test']);
```
in MyController::testAction():
```php
$this->view->setVar('bar', 'foo');
```
in test.volt:
```
{{ bar }}
```
Returns:
```
foo
```

### After 3.2.3
Returns:
```
1234
```

### Details

* Phalcon version: 3.2.3
* PHP Version: 7.0
* Operating System: Linux
* Installation type: package from packagecloud.io
* Server: Other

sergeyklay:
Fixed in the [latest stable version](https://github.com/phalcon/cphalcon/releases/tag/v3.2.4). Feel free to open a new issue if the problem appears again.  Thank you for contributing, and for helping us make Phalcon better.

PR #13122, opened 2017-10-18
Don't send dispatcher parameters

dugwood:
* Type: bug fix
* Link to issue: #13121 

Small description of change:
These will erase previously set variables, and dispatcher's params were never accessible this way so far, so removing the argument in this function is fine.


Jurigag:
We don't accept PRs to master. Send it to 3.2.x please. Also update changelog.

dugwood:
@Jurigag sorry, just seen this... but every release is impacted!

Jurigag:
We will merge 3.2.x to master. Master is just stable branch without direct changes to it.

Issue #13107, opened 2017-10-03
[NFR] Switch block for Volt


lajosbencz:
New feature request, as per: https://forum.phalconphp.com/discussion/17035/feature-request-switch-for-volt

# Intended behaviour

Allows for switch/case conditional rendering in Volt

# Example

```volt
{% switch matrixBlock.type %}

 {% case "text" %}

     {{ matrixBlock.textField | markdown }}

 {% case "image" %}

    {{ matrixBlock.image[0].getImg() }}

 {% default %}

    <p>A font walks into a bar.</p>
    <p>The bartender says, “Hey, we don’t serve your type in here!”</p>

{% endswitch %}
```

JABirchall:
Good suggestion Would be great for example a role badge would need a way to break out the switch as well so that it doesn't cascade. I suggest a `{% casebreak 'match' %}`

```php
{% switch user.getRoles() %}
    {% case "Owner" %}
    <span class="label label-pill label-primary">Owner</span>
    {% case "Admin" %}
     <span class="label label-pill label-primary">Admin</span>
    {% casebreak "Partner" %} // this should break the switch
    <span class="label label-pill label-primary">Partner</span>
    {{ matrixBlock.image[0].getImg() }}
    {% default %}
    Member
{% endswitch %}
```

Slyvampy:
Thanks for creating this NFR for me :)

sergeyklay:
Actually I prefer more something like: 

```jinja
{% switch username %}
    {# Each `case` statement can take multiple values to compare the variable #}
    {% case "Jim" "Bob" "Joe" %}
        Hello {{ username }}
    {% case "Nik" %}
        Hey hey {{ username }}!
    {% else %}
        {#
            An optional `else` statement sets off the default output
            if no matches could be found
        #}
        Who are you?
{% endswitch %}
```

I'll try to deal with this, but note it will be a switch without `break`-statements implementation.

Jurigag:
Imho there shouldn't be endcase. Break should be just detected by next `{% case` or `{% default`.

Also not sure about this multiple values for case. Is it possible in plain php? I don't think so, not sure if i saw syntax like this in any language/template. But it might be good thing.

oscarmolinadev:
@Jurigag You "can" more or less with true/false.

```
<?php
        $value = 'Nik';
        switch (true) {
            case ($value === 'Jim' || $value === 'Bob' || $value === 'Joe'): {
                echo 'Hello Multiple';
[...]

PR #13130, opened 2017-10-21
Initial support of switch-case-default-endswitch statement for Volt

sergeyklay:
Hello!

* Type: **new feature**
* Link to issue: #13107

Small description of change:

Implemented `switch-case` recognition for Volt. Introduced syntax:
```jinja
{% switch foo %}
    {% case 0 %}
    {% case 1 %}
    {% case 2 %}
        `foo` is less than 3 but not negative
        {% break %}
    {% case 3 %}
        `foo` is 3
        {% break %}
    {% default %}
        `foo` is {{ foo }}
{% endswitch %}
```

It is important to understand how the switch statement is executed in order to avoid mistakes.
The switch statement executes line by line (actually, statement by statement). This is why `break` is necessary in some cases.

**NOTE:** Any output (including whitespace) between a switch statement and the first case will result in a syntax error. This is the responsibility of the user. However, we can clear empty lines and whitespaces here to reduce the number of errors. For more see: http://php.net/control-structures.alternative-syntax

Thus, for example the Volt syntax:
```jinja
{% switch username %}
    {% case "Jim" %}
        Hello username
    {% case "Nik" %}
        {{ username }}!
        {% break %}
    {% default %}
        Who are you?
{% endswitch %}
```

will be compiled to:
```html+php
<?php switch ($username): ?>
<?php case 'Jim': ?>
Hello username
<?php case 'Nik': ?>
<?= $username ?>!
<?php break; ?>
<?php default: ?>
Who are you?
<?php endswitch ?>
```

--------

##### `case` without `switch`

```jinja
 {% case EXPRESSION %}
```

Will trow Fatal error: `Uncaught Phalcon\Mvc\View\Exception: Unexpected CASE`.

--------

##### `switch` without `endswitch`

```jinja
 {% switch EXPRESSION %}
```

Will trow Fatal error: `Uncaught Phalcon\Mvc\View\Exception: Syntax error, unexpected EOF in ..., there is a 'switch' block without 'endswitch'`.

--------

##### `default` without `switch`

```jinja
 {% default %}
```

Will _not throw_ error because `default` is reserved word for filters like `{{ EXPRESSION | default(VALUE) }}` but _in this case_ the expression don't do anything - just empty char `''` .

--------

##### nested `switch`

```jinja
{% switch EXPRESSION %}
  {% switch EXPRESSION %}
  {% endswitch %}
{% endswitch %}
```

Will trow Fatal error: `Uncaught Phalcon\Mvc\View\Exception: A nested switch detected. There is no nested switch-case statements support in ... on line ...`.

--------

##### a `switch` without expression

```jinja
{% switch %}
  {% case EXPRESSION %}
      {% break %}
{% endswitch %}
```

Will trow Fatal error: `Uncaught Phalcon\Mvc\View\Exception: Syntax error, unexpected token %} in ... on line ...`.

--------

Thanks.


sergeyklay:
cc: @phalcon/core-team

xboston:
👍

PR #13148, opened 2017-11-07
Use PSR-1 for PHPUnit TestCase

carusogabriel:
Use [PSR-1](http://www.php-fig.org/psr/psr-1/#namespace-and-class-names) while extending PHPUnit TestCase class. This will help us when to migrate to PHPUnit 6, that [no longer support snake case namespaces](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

### DIFF
--- a/unit-tests/CacheTest.php
+++ b/unit-tests/CacheTest.php
@@ -20,7 +20,9 @@
 
 require_once 'helpers/xcache.php';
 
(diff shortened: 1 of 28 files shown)


sergeyklay:
Thank you

Issue #12921, opened 2017-06-21
Remove session from factorydefault or start it when creating instance


Jurigag:
Right now having session service in `Phalcon\Di\FactoryDefault` is pointless, because you need to start session instance anyway to make it working properly. So you need to have code like this anyway:

```php
$di->setShared('session', function() {
    $session= new Phalcon\Session\Adapter\Files();
    $session->start(); // this is required pretty much to make session things working properly, including persistent
    return $session
});
```

If you don't have this and you are using just session from `FactoryDefault` then it won't work properly because it's not started until you will start it by accessing this service and executing start method. Like value after doing `$session->set()` will be not available right now when using session from `FactoryDefault` in other request.

Or we could change it to just start session on instance construction.

This is somehow related issue - https://github.com/phalcon/cphalcon/issues/12918

sergeyklay:
Actually, as you can see, we just use "connect" approach in many adapters if they does not connected yet. So what do you think about check session status? IMO we can remove session from factory default in 4.x branch only.

Jurigag:
To make it working and as a bugfix for now for phalcon 3 - fine. But in phalcon 4 to avoid overhead better just remove it from factory default.

sergeyklay:
Cc: @phalcon/framework-team

Jurigag:
Also i would consider thinking about for example memcache adapters and other - do we really need to do 

```
		let memcache = this->_memcache;

		if typeof memcache != "object" {
			this->_connect();
			let memcache = this->_memcache;
		}
```

In every single method? Why just not call _connect one time in `__construct`?

sergeyklay:
Refs: https://github.com/phalcon/cphalcon/issues/10530

dschissler:
If we do leave the session adapter in the factory web DI then it should be setup in a way that allows it do be loaded from a static YML or INI configuration file for the most simple files setup.

I'd like to make it so that simple builtin plugins could be added from the configuration file under the `plugins` key.  For example now I'm adding a database reconnect plugin and it would be great if I could hook it in using the same yml file.

So I bring this up because neither the existing convention nor this new proposal would exactly work.  What if we added an interface to classes that are meant to be used as services so that the DI could check if they are started and if not then autostart it by calling the interface defined `start()`?

dschissler:
@Jurigag Is there a crazy performance penalty on checking if an object implements an interface?  I ask because sometimes PHP surprises me.

dschissler:
[...]

PR #13157, opened 2017-11-12
Service class API change and addition of helper class

dschissler:
I removed the name parameter from the Service class since the only place where it is used in the Phalcon code is for throwing an exception when there was a problem with the service.  I added a new exception class which the DI catches and since the DI knows what the name of the service should be it is able to provide the same error reporting capabilities.

This API change allows for `Phalcon\Di\Service` classes to more gracefully wrap service closures.

I added a `Phalcon\Di\Service\SharedService` class which simply offers API sugar to make the second parameter unnecessary when using a shared service.

I added a new helper class `Phalcon\Di\Service\ServiceRegistry` for registering a directory of service files in either yaml or php.  These files can return a php array with a static definition, a closure, a `Service` object or a string for the class name.

[edit] I'm going to take a week off from this and then I'll take down this pull request and start over with a clean repo with just 2-3 commits.  I'm going to redesign the API a bit, modify the tests, etc.

### DIFF
--- a/phalcon/di/exception/serviceResolutionException.zep
+++ b/phalcon/di/exception/serviceResolutionException.zep
@@ -1,0 +1,30 @@
+
+/*
+ +------------------------------------------------------------------------+
+ | Phalcon Framework                                                      |
+ +------------------------------------------------------------------------+
+ | Copyright (c) 2011-2017 Phalcon Team (https://phalconphp.com)          |
+ +------------------------------------------------------------------------+
+ | This source file is subject to the New BSD License that is bundled     |
+ | with this package in the file LICENSE.txt.                             |
+ |                                                                        |
+ | If you did not receive a copy of the license and are unable to         |
+ | obtain it through the world-wide-web, please send an email             |
+ | to license@phalconphp.com so we can send you a copy immediately.       |
+ +------------------------------------------------------------------------+
+ | Authors: Andres Gutierrez <andres@phalconphp.com>                      |
+ |          Eduar Carvajal <eduar@phalconphp.com>                         |
+ +------------------------------------------------------------------------+
+ */
+
+namespace Phalcon\Di\Exception;
+
+use Phalcon\Di\Exception\ServiceResolutionException;
+
+/**
+ * Phalcon\Di\Exception\ServiceResolutionException
+ *
+ */
+class ServiceResolutionException extends \Phalcon\Di\Exception
+{
+}
--- a/phalcon/di/service.zep
+++ b/phalcon/di/service.zep
@@ -21,6 +21,7 @@ namespace Phalcon\Di;
 
 use Phalcon\DiInterface;
 use Phalcon\Di\Exception;
+use Phalcon\Di\Exception\ServiceResolutionException;
 use Phalcon\Di\ServiceInterface;
 use Phalcon\Di\Service\Builder;
 
@@ -40,9 +41,6 @@ use Phalcon\Di\Service\Builder;
  */
 class Service implements ServiceInterface
 {
-
-	protected _name;
-
 	protected _definition;
 
 	protected _shared = false;
@@ -55,20 +53,12 @@ class Service implements ServiceInterface
 	 * Phalcon\Di\Service
 	 *
 	 * @param mixed definition
+	 * @param boolean shared
 	 */
-	public final function __construct(string! name, definition, boolean shared = false)
+	public function __construct(definition, boolean shared = false)
 	{
-		let this->_name = name,
-			this->_definition = definition,
+		let this->_definition = definition,
 			this->_shared = shared;
-	}
-
-	/**
-	 * Returns the service's name
-	 */
-	public function getName() -> string
-	{
-		return this->_name;
 	}
 
 	/**
@@ -121,6 +111,7 @@ class Service implements ServiceInterface
 	 * Resolves the service
 	 *
 	 * @param array parameters
+	 * @param \Phalcon\DiInterface dependencyInjector
 	 * @return mixed
 	 */
 	public function resolve(parameters = null, <DiInterface> dependencyInjector = null)
@@ -202,7 +193,7 @@ class Service implements ServiceInterface
 		 * If the service can't be built, we must throw an exception
 		 */
 		if found === false  {
-			throw new Exception("Service '" . this->_name . "' cannot be resolved");
+			throw new ServiceResolutionException();
 		}
 
 		/**
@@ -254,6 +245,7 @@ class Service implements ServiceInterface
 	/**
 	 * Returns a parameter in a specific position
 	 *
+	 * @param int position
 	 * @return array
 	 */
 	public function getParameter(int position)
@@ -291,10 +283,6 @@ class Service implements ServiceInterface
 	public static function __set_state(array! attributes) -> <Service>
 	{
 		var name, definition, shared;
-
-		if !fetch name, attributes["_name"] {
-			throw new Exception("The attribute '_name' is required");
-		}
 
 		if !fetch definition, attributes["_definition"] {
 			throw new Exception("The attribute '_definition' is required");
--- a/phalcon/di/service.zep
+++ b/phalcon/di/service.zep
@@ -282,6 +282,6 @@ class Service implements ServiceInterface
 			throw new Exception("The attribute '_shared' is required");
 		}
 
-		return new self(name, definition, shared);
+		return new self(definition, shared);
 	}
 }
--- a/phalcon/di/service.zep
+++ b/phalcon/di/service.zep
@@ -52,10 +52,8 @@ class Service implements ServiceInterface
 	/**
 	 * Phalcon\Di\Service
 	 *
-	 * @param mixed definition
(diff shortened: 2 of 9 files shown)


phalcon/di/exception/serviceResolutionException.zep
```diff
@@ -0,0 +1,30 @@
...
+
+use Phalcon\Di\Exception\ServiceResolutionException;
+
+/**
+ * Phalcon\Di\Exception\ServiceResolutionException
+ *
+ */
+class ServiceResolutionException extends \Phalcon\Di\Exception
+{
+}
```
sergeyklay: Could you rename this file `phalcon/di/exception/serviceResolutionException.zep` => `phalcon/di/exception/serviceresolutionexception.zep` to follow common style

phalcon/di/service.zep
```diff
@@ -55,23 +53,15 @@ class Service implements ServiceInterface
 	 * Phalcon\Di\Service
 	 *
 	 * @param mixed definition
+	 * @param boolean shared
```
sergeyklay: Annotation not needed at all. Just use
```zep
public function __construct(var definition, boolean shared = false)
```

phalcon/di/service.zep
```diff
@@ -121,6 +111,7 @@ class Service implements ServiceInterface
 	 * Resolves the service
 	 *
 	 * @param array parameters
+	 * @param \Phalcon\DiInterface dependencyInjector
 	 * @return mixed
 	 */
 	public function resolve(parameters = null, <DiInterface> dependencyInjector = null)
```
sergeyklay:
```zep
public function resolve(array parameters = null, <DiInterface> dependencyInjector = null) -> var
```
and remove `@param`s

phalcon/di/service.zep
```diff
@@ -254,6 +245,7 @@ class Service implements ServiceInterface
 	/**
 	 * Returns a parameter in a specific position
 	 *
+	 * @param int position
 	 * @return array
 	 */
 	public function getParameter(int position)
```
sergeyklay:
```zep
public function getParameter(int position) -> array | null
```

and remove `@param` and `@return`

phalcon/di/service/serviceregistry.zep
```diff
@@ -0,0 +1,115 @@
...
+ * Phalcon\Di\Service\ServiceRegistry
+ *
+ **/
+class ServiceRegistry
+{
+
+	/**
+	 *
+	 */
+	protected di;
```
sergeyklay: Why do we need empty comment here?

phalcon/di/service/serviceregistry.zep
```diff
@@ -0,0 +1,115 @@
...
+			default:
+				throw new Exception("Invalid services.");
+				break;
+		}
+	}
+
+	/**
+	 *
+	 */
+	protected function _registerDir(string! dirPath)
```
sergeyklay: Method names should not be prefixed with a single underscore to indicate protected nature. Actually when you extend this class in PHP your code will not pass the code style tests.

A propose get rid of underscores in method names for 4.x

sergeyklay (changes requested):

sergeyklay:
You have successfully broken all the tests :)

phalcon/di/service/serviceregistry.zep
```diff
@@ -0,0 +1,115 @@
...
+
+		switch (gettype(dirPath)) {
+			case "string":
+				this->_registerDir(dirPath);
+				break;
+			case "array":
+			case "object":
+				for dirPathItem in iterator(dirPath) {
+					this->_registerDir(dirPathItem);
+				}
```
sergeyklay:
```zep
if dirPath instanceof Traversable
```
?

phalcon/di/service/serviceregistry.zep
```diff
@@ -0,0 +1,115 @@
...
+				break;
+			default:
+				throw new Exception("Invalid services.");
```
sergeyklay: I would like to see here a error message with an explanation of what happened and the reasons for this

phalcon/di/service/serviceregistry.zep
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	/**
+	 *
+	 */
+	public function registerDir(dirPath)
+	{
+		var dirPathItem;
+
+		switch (gettype(dirPath)) {
```
sergeyklay:
```zep
switch typeof dirPath {
    // ...
}
```

phalcon/di/service/serviceregistry.zep
```diff
@@ -0,0 +1,115 @@
...
+	/**
+	 *
+	 */
+	protected function _registerDir(string! dirPath)
+	{
+		var dir, fileinfo, extension, serviceNameSection, forceShared, serviceName;
+
+		let dir = new \DirectoryIterator(dirPath);
+		for fileinfo in iterator(dir) {
+			if (fileinfo->isDir()) {
```
sergeyklay:
```zep
if unlikely fileinfo->isDir() === true {
    // ...
}
```

Also I'm pretty sure we don't need to scan this dir in each request

sergeyklay: How about **`.`** and **`..`** ?

phalcon/di/service/serviceregistry.zep
dschissler: I like that.  I did this without Internet access or Zephir documentation.

phalcon/di/service/serviceregistry.zep
dschissler: sure

dschissler: Not much has changed except for the `Service` constructor.

phalcon/di/service/serviceregistry.zep
dschissler: I did it that way since I wanted the method name to be the exact same.  What do you think about `registerDirInternal`?

phalcon/di/service/serviceregistry.zep
dschissler: Well I don't know about caching here.  I don't want to handle any directory (current or parent) and for now I think that its best not to recurse.  It only reads files.

phalcon/di/service/serviceregistry.zep
sergeyklay: @dschissler Frankly speaking I'm a little confused with `Dir`. Probably it's because I'm not a native speaker (like many of our users). Dir it is: DIRection. DIRectory, Do It Right, ...? Why do we need abbreviations for methods that aren't part of the public API?

phalcon/di/service/serviceregistry.zep
sergeyklay: In this case I propose consider such  a design with which user can easily inject a caching mechanism. For example Decorator Pattern

dschissler: Ok well this should be before the DI has been setup, since this is is intended to be a possible first step for getting services going.

I won't be working on this for 7 days and then I will take down this pull request and do more work on it and then resubmit just two clean commits.

phalcon/di/service/serviceregistry.zep
dschissler: Yes `Dir` is short for `Do it right`.  Its known as IDD (Intent Driven Development).  Saying nice things and that sort of thing.

Kevin and I were chatting about it and I'll be doing more API work in about 7 days at which time I'll take down this pull request.

phalcon/di/service/serviceregistry.zep
dschissler: btw what is the time difference between scanning a directory with 1-25 files in it versus reading a single file?  Also how would changes to individual files be detected?

If someone could propose to me a mechanism for caching this data then it would give me something to work with.

Jurigag: Php already caches stuff like this.

dschissler:
@sergeyklay @Jurigag Kevin and I fixed much test breakage but I don't know what this form breakage is about.  https://travis-ci.org/phalcon/cphalcon/jobs/309308215#L2504

dschissler:
@sergeyklay @Jurigag Kevin and I fixed much test breakage but I don't know what this form breakage is about.  https://travis-ci.org/phalcon/cphalcon/jobs/309308215#L2504

dschissler:
@sergeyklay Please review this and make suggestions so that I can resubmit all of these changes as two clean commits.  Its currently many commits and I'd like to make it come into the official tree better.  Look at this and if you like it then I'll redo it.  Don't pull this.

sergeyklay:
@dschissler Could you remove session from factory default? See: https://github.com/phalcon/cphalcon/issues/12921

dschissler:
@sergeyklay Kevin and I have done R&D into using an interface to allow the DI to autostart the session.  It seems that it may not be possible due to other factors but I have not yet concluded my work and so I would like to defer that.  Also I think that this change would be more appropriately done by itself instead of bundled.

PR #13160, opened 2017-11-15
Fix adding empty string attributes

NikolasSumrak:
In model.zep at 2104 line:
```if isset emptyStringValues[field]```
if we have array $array = ['field' => null]; - function isset returns false!

Hello!

* Type: bug fix

Small description of change:

Thanks


sergeyklay:
We don't add changes directly to stable branches or `master`. Could you please send this PR to the `3.3.x` branch? Also could you please provide a small example to reproduce the issue?

NikolasSumrak:
@sergeyklay  
So i close it and open another one here https://github.com/phalcon/cphalcon/pull/13163

PR #13171, opened 2017-11-20
Bug Fix #13170

### LINKED ISSUE #13170: Dynamic update does not work as expected for fields which raw values were assigned to
### DynamicUpdate And RawValue 

> 
When I set useDynamicUpdate=true, and the origin value = 1, then I set
model->value = \Phalcon\Db\RawValue('value + 1').
When I save, The value is not changed.

When I change value type is bigint, the result is right.
When the value type is int, the result is wrong.

```php
<?php
class TDksRouteRequestLog extend \Phalcon\Mvc\Model {
    public function initialize()
    {
        $this->useDynamicUpdate(true);
    }
}

$hour = date('YmdH');
$route = '/api/log';

$model = TDksRouteRequestLog::findFirst([
    'conditions' => 'hour = ?0 AND route = ?1',
    'bind' => [$hour, $route],
]);

if (empty($model)) {
    $model = new TDksRouteRequestLog();
    $model->hour = $hour;
    $model->route = $route;
    $model->request = 1;
} else {
    $model->request = new RawValue('request + 1');
}

$model->save();

SQL: UPDATE `dks`.`t_dks_route_request_log` SET `gmt_modify` = ? WHERE `id` = ?
BINDS: ["2017-11-20 11:55:19",12]
```

### Details

* Phalcon version: (3.2.4)
* PHP Version: (7.0.25)
* Operating System:
* Installation type:installing via package manager
* Zephir version (if any):
* Server: Nginx
* Other related info (Database, table schema):

limingxinleo:
Hello!

* Type: bug fix
* Link to issue: #13170

Small description of change:

Dynamic update does not work as expected for fields which raw values were assigned to

Thanks

### DIFF
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -2562,15 +2562,12 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
  											let changed = (boolean) snapshotValue !== (boolean) value;
  											break;
 
- 										case Column::TYPE_INTEGER:
- 											let changed = (int) snapshotValue !== (int) value;
- 											break;
-
  										case Column::TYPE_DECIMAL:
  										case Column::TYPE_FLOAT:
  											let changed = floatval(snapshotValue) !== floatval(value);
  											break;
 
+ 										case Column::TYPE_INTEGER:
  										case Column::TYPE_DATE:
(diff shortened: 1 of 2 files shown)


sergeyklay:
Thank you

PR #13190, opened 2017-12-01
Service class changed and ServiceRegistry

dschissler:
I removed the name parameter from the Service class since the only place where it is used in the Phalcon code is for throwing an exception when there was a problem with the service. I added a new exception class which the DI catches and since the DI knows what the name of the service should be it is able to provide the same error reporting capabilities.

This API change allows for Phalcon\Di\Service classes to more gracefully wrap service closures.

I added a Phalcon\Di\Service\SharedService class which simply offers API sugar to make the second parameter unnecessary when using a shared service.

I added a new helper class Phalcon\Di\Service\ServiceRegistry for registering a directory of service files in either yaml or php. These files can return a php array with a static definition, a closure, a Service object or a string for the class name.


dschissler:
@sergeyklay Fix the 4 branch form tests or treat this PR as passing.

sergeyklay:
@dschissler I'll take a look a bit later. Thank you

PR #13198, opened 2017-12-06
Fix probable boolean logic error

mmcco:
* Type: bug fix
* Link to issue:

- [N/A] I wrote some tests for this PR.

Small description of change: This if condition is always false, and it seems clear that the author meant to use || rather than &&.

Thanks for your time,
Michael McConville
University of Utah

### DIFF
--- a/ext/phalcon/assets/filters/cssminifier.c
+++ b/ext/phalcon/assets/filters/cssminifier.c
@@ -113,7 +113,7 @@ static int phalcon_cssmin_machine(cssmin_parser *parser, unsigned char c TSRMLS_
 
 	switch (parser->state) {
 		case STATE_FREE:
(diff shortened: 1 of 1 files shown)


mmcco:
The corresponding test (that for the CSS minifier) failed in Travis CI. However, the original condition seems inherently and obviously wrong. Maybe the surrounding code evolved to correct for it?

On the other hand, the test that failed seemed to expect the minifier to leave `' '` (a single space) unchanged, when the minifier now returns `''` (an empty string). Maybe the test expects a result that's actually erroneous?

sergeyklay:
@mmcco We don't add changes directly to stable branches or `master`. Could you please send this PR to the `3.3.x` branch? Also update please test and the change log. 

Thank you

mmcco:
Does this work? https://github.com/phalcon/cphalcon/pull/13200

PR #13200, opened 2017-12-08
Fix probable boolean logic error

mmcco:
* Type: bug fix

Small description of change: This if condition is always false, and it seems clear that the author meant to use || rather than &&.

Thanks for your time,
Michael McConville
University of Utah

### DIFF
--- a/ext/phalcon/assets/filters/cssminifier.c
+++ b/ext/phalcon/assets/filters/cssminifier.c
@@ -113,7 +113,7 @@ static int phalcon_cssmin_machine(cssmin_parser *parser, unsigned char c TSRMLS_
 
 	switch (parser->state) {
 		case STATE_FREE:
-			if (c == ' ' && c == '\t' && c == '\n' && c == '\r') {
+			if (c == ' ' || c == '\t' || c == '\n' || c == '\r') {
 				c = 0;
 			} else if (c == '@'){
 				parser->state = STATE_ATRULE;
--- a/tests/unit/Assets/Filters/CssMinTest.php
+++ b/tests/unit/Assets/Filters/CssMinTest.php
@@ -72,7 +72,7 @@ class CssMinTest extends UnitTest
             "The cssmin filter with space as parameter does not return correct results",
             function () {
                 $cssmin = new Cssmin();
-                expect($cssmin->filter(' '))->equals(' ');
+                expect($cssmin->filter(' '))->equals('');
             }
         );
 


Ultimater:
Good eye! And thanks for repairing the test!

Interestingly enough, this change wouldn't affect the CSS minimizing logic whatsoever without a `break;` as the STATE_FREE and STATE_SELECTOR states currently have no differences and would still encounter:  
```
p = cssmin_back_peek(parser);
if (p == ',' || p == '>' || p == ':') {
     c = 0;
} else {
     c = ' ';
}
```

In other words we could actually write:  
```
case STATE_FREE:
case STATE_SELECTOR:
```
and it would be the equivalent of either before or after your change. 

I think what the author intended was actually to trim ALL whitespace when in the `STATE_FREE`, without using looking around to make sure we're not in a selector:
```
if (c == ' ' || c == '\t' || c == '\n' || c == '\r') {
    c = 0;
    break;
}
```
Note the usage of my break.

Preserving the space is necessary on the selector for example:  
`div p{margin:2px 4px 5px 7px;}`
However if we're in the `STATE_FREE`, we have not yet encountered the first character `d` of `div`, so it's safe to remove all whitespace until such a character is encountered.

While we're in here, it would be nice to change:  
`parser.state = 1;` to `parser.state = STATE_FREE;`
This way it's easier to search for STATE_FREE and find where it's initially set.

Also `last_state` doesn't appear to be used whatsoever and would be nice to trash.

Probably some room to fix the parser code further as mentioned at the top of the file:  
```
** cannot handle nested { blocks but will ignore aditional { in parens ()
** no in quote detection for ( or }
```

Ultimater:
Oh wait, a slight difference:  

```
if ((c == ' ' || c == '\t')) {
    p = cssmin_peek(parser);
    if (p == '{' || p == '\t' || p == ' ' || p == '>' || p == ',') {
    c = 0;
} else {
    p = cssmin_back_peek(parser);
    if (p == ',' || p == '>' || p == ':') {
        c = 0;
    } else {
        c = ' ';
    }
}
```

c would be set to 0 in STATE_FREE of your version before it runs into the above code in STATE_SELECTOR.
Yeah, so that would explain the need to adjust the test when it was working before.  

I would sooner still see a `break;` though. If we're changing `c`, it's spaghetti code if we let the states spill through, specially after changing an important variable holding the character we're working on...

sergeyklay (approved):

sergeyklay:
@mmcco Could you please update change log?

sergeyklay:
Thank you

Issue #13226, opened 2017-12-18
Transactions passed in find() or findFirst() to query inside of started transactions


chilimatic:
Hello!

Type: new feature | enhancement
Link to issue: #12409, #12564, #13097 
In raising this pull request, I confirm the following (please check boxes):

 - [x] I have read and understood the Contributing Guidelines?
 - [x] I have checked that another pull request for this purpose does not exist.
 - [x] I wrote some tests for this PR.

this ticket exists because I was to incompetent to first create a Ticket before starting a PR. The target branch has changed and I need to rewrite the history with the correct ticket number :)

Small description of change:

Currently you only can use the internal Transaction behaviour without an option to use a specific transaction in find / findFirst. This should be changed

This change is to achieve the following:

In Unit-testing one usual procedure is to create a transaction for a TestCase and after the execution rollback this change.

Which means there is a "long transaction" going on and we want to check if the transaction actually is writing the correct values in the database. But the static context of model::find / model::findFirst does not get the transaction passed.

To achieve that passing a new index "transaction" containing an transaction in the $param array seams to be a viable solution.

Another use-case can be having multiple transactions and wanting to look inside.

Just to be precise -> ofc we could just get the connection and query manually inside of it, but I want to be as close to the framework as possible when testing certain services hence I want to use the model.

I took your reviews and advice from the last attempt and used the params to not break the interfaces / change method signatures.

the logic is implemented as follows:

if you pass in a transaction via param in the find/findFirst method. It will set the transaction to query object and this transaction will be used for all queries and executions within this query context.

This should not affect any standard behaviour but still allow me to finally achieve cleaner unittests for my usecase.

sergeyklay:
Fixed in the `3.3.x` branch. Feel free to open a new issue if the problem appears again. Thank you for contributing.

chilimatic:
cool thx ! :)

PR #13227, opened 2017-12-18
#13226 add transaction option to find / findFirst

chilimatic:
Hello!

* Type: new feature | code quality | documentation
* Link to issue: https://github.com/phalcon/cphalcon/issues/13226

Small description of change:

please check:
https://github.com/phalcon/cphalcon/pull/13097

Thanks

### DIFF
--- a/phalcon/mvc/model.zep
+++ b/phalcon/mvc/model.zep
@@ -14,6 +14,7 @@
  +------------------------------------------------------------------------+
  | Authors: Andres Gutierrez <andres@phalconphp.com>                      |
  |          Eduar Carvajal <eduar@phalconphp.com>                         |
+ |          Jakob Oberhummer <cphalcon@chilimatic.com>                    |
  +------------------------------------------------------------------------+
  */
 
@@ -84,7 +85,6 @@ use Phalcon\Events\ManagerInterface as EventsManagerInterface;
  */
 abstract class Model implements EntityInterface, ModelInterface, ResultInterface, InjectionAwareInterface, \Serializable, \JsonSerializable
 {
-
 	protected _dependencyInjector;
 
 	protected _modelsManager;
@@ -97,7 +97,7 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 
 	protected _dirtyState = 1;
 
-	protected _transaction;
+	protected _transaction { get };
 
 	protected _uniqueKey;
 
@@ -112,6 +112,8 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 	protected _snapshot;
 
 	protected _oldSnapshot = [];
+
+	const TRANSACTION_INDEX = "transaction";
 
 	const OP_NONE = 0;
 
@@ -467,7 +469,7 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 	{
 		var key, keyMapped, value, attribute, attributeField, metaData, columnMap, dataMapped, disableAssignSetters;
 
-        let disableAssignSetters = globals_get("orm.disable_assign_setters");
+		let disableAssignSetters = globals_get("orm.disable_assign_setters");
 
 		// apply column map for data, if exist
 		if typeof dataColumnMap == "array" {
@@ -809,14 +811,66 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 	 * foreach ($robots as $robot) {
 	 *	 echo $robot->name, "\n";
 	 * }
+	 *
+	 * // encapsulate find it into an running transaction esp. useful for application unit-tests
+	 * // or complex business logic where we wanna control which transactions are used.
+	 *
+	 * $myTransaction = new Transaction(\Phalcon\Di::getDefault());
+	 * $myTransaction->begin();
+	 * $newRobot = new Robot();
+	 * $newRobot->setTransaction($myTransaction);
+	 * $newRobot->save(['name' => 'test', 'type' => 'mechanical', 'year' => 1944]);
+	 *
+	 * $resultInsideTransaction = Robot::find(['name' => 'test', Model::TRANSACTION_INDEX => $myTransaction]);
+	 * $resultOutsideTransaction = Robot::find(['name' => 'test']);
+	 *
+	 * foreach ($setInsideTransaction as $robot) {
+	 *     echo $robot->name, "\n";
+	 * }
+	 *
+	 * foreach ($setOutsideTransaction as $robot) {
+	 *     echo $robot->name, "\n";
+	 * }
+	 *
+	 * // reverts all not commited changes
+	 * $myTransaction->rollback();
+	 *
+	 * // creating two different transactions
+	 * $myTransaction1 = new Transaction(\Phalcon\Di::getDefault());
+	 * $myTransaction1->begin();
+	 * $myTransaction2 = new Transaction(\Phalcon\Di::getDefault());
+	 * $myTransaction2->begin();
+	 *
+	 *  // add a new robots
+	 * $firstNewRobot = new Robot();
+	 * $firstNewRobot->setTransaction($myTransaction1);
+	 * $firstNewRobot->save(['name' => 'first-transaction-robot', 'type' => 'mechanical', 'year' => 1944]);
+	 *
+	 * $secondNewRobot = new Robot();
+	 * $secondNewRobot->setTransaction($myTransaction2);
+	 * $secondNewRobot->save(['name' => 'second-transaction-robot', 'type' => 'fictional', 'year' => 1984]);
+	 *
+	 * // this transaction will find the robot.
+	 * $resultInFirstTransaction = Robot::find(['name' => 'first-transaction-robot', Model::TRANSACTION_INDEX => $myTransaction1]);
+	 * // this transaction won't find the robot.
+	 * $resultInSecondTransaction = Robot::find(['name' => 'first-transaction-robot', Model::TRANSACTION_INDEX => $myTransaction2]);
+	 * // this transaction won't find the robot.
+	 * $resultOutsideAnyExplicitTransaction = Robot::find(['name' => 'first-transaction-robot']);
+	 *
+	 * // this transaction won't find the robot.
+	 * $resultInFirstTransaction = Robot::find(['name' => 'second-transaction-robot', Model::TRANSACTION_INDEX => $myTransaction2]);
+	 * // this transaction will find the robot.
+	 * $resultInSecondTransaction = Robot::find(['name' => 'second-transaction-robot', Model::TRANSACTION_INDEX => $myTransaction1]);
+	 * // this transaction won't find the robot.
+	 * $resultOutsideAnyExplicitTransaction = Robot::find(['name' => 'second-transaction-robot']);
+	 *
+	 * $transaction1->rollback();
+	 * $transaction2->rollback();
 	 * </code>
 	 */
 	public static function find(var parameters = null) -> <ResultsetInterface>
 	{
-		var params, builder, query, bindParams, bindTypes, cache, resultset, hydration, dependencyInjector, manager;
-
-		let dependencyInjector = Di::getDefault();
-		let manager = <ManagerInterface> dependencyInjector->getShared("modelsManager");
+		var params, query, resultset, hydration;
 
 		if typeof parameters != "array" {
 			let params = [];
@@ -827,36 +881,7 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 			let params = parameters;
 		}
 
-		/**
-		 * Builds a query with the passed parameters
-		 */
-		let builder = manager->createBuilder(params);
-		builder->from(get_called_class());
-
-		let query = builder->getQuery();
-
-		/**
-		 * Check for bind parameters
-		 */
-		if fetch bindParams, params["bind"] {
-
-			if typeof bindParams == "array" {
-				query->setBindParams(bindParams, true);
-			}
-
-			if fetch bindTypes, params["bindTypes"] {
-				if typeof bindTypes == "array" {
-					query->setBindTypes(bindTypes, true);
-				}
-			}
-		}
-
-		/**
-		 * Pass the cache options to the query
-		 */
-		if fetch cache, params["cache"] {
-			query->cache(cache);
-		}
+		let query = static::getPreparedQuery(params);
 
 		/**
 		 * Execute the query passing the bind-params and casting-types
@@ -886,7 +911,7 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 	 *
 	 * // What's the first mechanical robot in robots table?
 	 * $robot = Robots::findFirst(
-	 *     "type = 'mechanical'"
+	 *	 "type = 'mechanical'"
 	 * );
 	 *
 	 * echo "The first mechanical robot name is ", $robot->name;
@@ -900,18 +925,27 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 	 * );
 	 *
 	 * echo "The first virtual robot name is ", $robot->name;
-	 * </code>
 	 *
-	 * @param string|array parameters
-	 * @return static
+	 * // behaviour with transaction
+	 * $myTransaction = new Transaction(\Phalcon\Di::getDefault());
+	 * $myTransaction->begin();
+	 * $newRobot = new Robot();
+	 * $newRobot->setTransaction($myTransaction);
+	 * $newRobot->save(['name' => 'test', 'type' => 'mechanical', 'year' => 1944]);
+	 *
+	 * $findsARobot = Robot::findFirst(['name' => 'test', Model::TRANSACTION_INDEX => $myTransaction]);
+	 * $doesNotFindARobot = Robot::findFirst(['name' => 'test']);
+	 *
+	 * var_dump($findARobot);
+	 * var_dump($doesNotFindARobot);
+	 *
+	 * $transaction->commit();
+	 * $doesFindTheRobotNow = Robot::findFirst(['name' => 'test']);
+	 * </code>
 	 */
 	public static function findFirst(var parameters = null) -> <Model>
 	{
-		var params, builder, query, bindParams, bindTypes, cache,
-			dependencyInjector, manager;
-
-		let dependencyInjector = Di::getDefault();
-		let manager = <ManagerInterface> dependencyInjector->getShared("modelsManager");
+		var params, query;
 
 		if typeof parameters != "array" {
 			let params = [];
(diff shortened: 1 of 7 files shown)


sergeyklay:
Could you remove annotations? We'll generate PHPDoc from the source code by using `zephir stubs` command.

chilimatic:
@sergeyklay 
just to be sure you mean all annotations ? so the execeptions and property annotations as well ? or just the method signature ones where you actually can infer them? 

or do they get ignored anyway ? :) just so I understand for the next time

sergeyklay:
In mos cases we don't need for `@param` and `@return`. There is no support for `@throws` and property's `@var`

chilimatic:
@sergeyklay fair enough, I will remove them all :). Maybe the exceptions would come in handy for example the phalcon rollback behaviour is rather annoying and a hint that I need to catch it to resume would be nice ... but that's ofc off topic. 

I will remove them :) thx ! :)

chilimatic:
@sergeyklay 

-> the travis error for php5.6 is 

```
1) CacheTest::testDataFileCacheIncrement

Failed asserting that 2 matches expected null.


/home/travis/build/phalcon/cphalcon/unit-tests/CacheTest.php:52
/home/travis/build/phalcon/cphalcon/tests/_ci/phpunit.php:10
```

should I retrigger ? or can we ignore this ? :)

sergeyklay:
@chilimatic Could you please squash commits into one to keep history clean.

chilimatic:
@sergeyklay i thought github supports that merge option already, but okay I will do it by hand :)

Ultimater:
Can we rename these to indicate they are tests?
```
useTransactionWithinFind()
useTransactionWithinFindFirst()
useTransactionOutsideFind()
useTransactionOutsideFindFirst()
```

Ultimater:
Everything else looks fine to me.

phalcon/mvc/model.zep
```diff
@@ -900,18 +925,27 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
...
-	 * @param string|array parameters
-	 * @return static
+	 * // behaviour with transaction
+	 * $myTransaction = new Transaction(\Phalcon\Di::getDefault());
+	 * $myTransaction->begin();
+	 * $newRobot = new Robot();
+	 * $newRobot->setTransaction($myTransaction);
+	 * $newRobot->save(['name' => 'test', 'type' => 'mechanical', 'year' => 1944]);
+	 *
+	 * $findsARobot = Robot::findFirst(['name' => 'test', Model::TRANSACTION_INDEX => $myTransaction]);
```
sergeyklay: I'm pretty sure there is a lot model with `transaction` field. Could we use another way to use this?

sergeyklay (changes requested):

phalcon/mvc/model.zep
chilimatic: good concern, yes ofc. should we use modelsTransaction than ? this seems to be a common pattern through the framework.

sergeyklay: how about this:
```php
Robot::findFirst([
    'condition' => '...',
    'transaction' => '...',
    ...
]);
```

As you probably know `Robot::findFirst(['name' => 'test'])` is alias for `Robot::findFirst(['condition' => 'name = test'])`

chilimatic: hmmm yes I read it and I understand what you're suggesting. Too me the problem was the possible field-name collisions. 

And I read the builder implementation and alias is a simplification since using an array as conditions is always a conjunction and not a disjunction, at least if I read it correctly. 

hence "condition" is obviously much more flexible it is not a triplet based conjunctive composition.

So even if I would be to write it, with the "conditions" field. we still could face similar issue due to a lack of consistency. 

Since we cannot claim the field-name "transaction" without introducing a possible breaking change. So using a more complex / a composed field-name reduces the  possible collisions. 

As you can clearly read if someone uses an "order" or "limit" we have clear semantic collisions as well. So either way we have a clear problem unless we really use a prefix
that's why I actually picked the Model::TRANSACTION_INDEX not because it's pretty but it's a readable approach and we can add a prefix "namespace" like "ph-transaction".

But I digress, it boils down to the classic:

The 2 hardest things in computer science: cache invalidation, naming things and off by 1 errors. 

Basically we have a conflation issues. we conflated the 
SQL "where" clause predicates with other clauses. 

So I will ofc do as you suggest but we are inconsistent with our behaviour if we only would pass it in certain cases. 

tl;dr: 

your solution is better, but still can confusing to the user since we display inconsistent behaviour.

Jurigag: Wait what, if there is no conditions key then first element in array is always consider as a conditions. So i don't know what `transaction` key makes a problem, it won't be used anywhere as a column name. Same with order or limit. If someone puts as first element `'order' => 10` and then conditions like `name = Jurigag` then it's only their problem and they missed a docs.

sergeyklay:
```
$ mysql

Your MySQL connection id is 553049
Server version: 5.5.53-log Source distribution

mysql> use phalcon_test;
Reading table information for completion of table and column names
You can turn off this feature to get a quicker startup with -A

Database changed
mysql> CREATE TABLE BUDGET_TRANSACTION (
    ->   transaction BIGINT PRIMARY KEY NOT NULL,
    ->   created     DATE NOT NULL,
    ->   payee       CHAR(50) NOT NULL,
    ->   amt         REAL NOT NULL,
    ->   type        CHAR(6) NOT NULL
    -> );
Query OK, 0 rows affected (0.01 sec)

mysql> describe BUDGET_TRANSACTION;
+-------------+------------+------+-----+---------+-------+
| Field       | Type       | Null | Key | Default | Extra |
+-------------+------------+------+-----+---------+-------+
| transaction | bigint(20) | NO   | PRI | NULL    |       |
| created     | date       | NO   |     | NULL    |       |
| payee       | char(50)   | NO   |     | NULL    |       |
| amt         | double     | NO   |     | NULL    |       |
| type        | char(6)    | NO   |     | NULL    |       |
+-------------+------------+------+-----+---------+-------+
5 rows in set (0.00 sec)

mysql>
mysql> show create table BUDGET_TRANSACTION;
+--------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Table              | Create Table                                                                                                                                                                                                                                              |
+--------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| BUDGET_TRANSACTION | CREATE TABLE `BUDGET_TRANSACTION` (
  `transaction` bigint(20) NOT NULL,
  `created` date NOT NULL,
  `payee` char(50) NOT NULL,
  `amt` double NOT NULL,
  `type` char(6) NOT NULL,
  PRIMARY KEY (`transaction`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1 |
+--------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.00 sec)

mysql>
```

chilimatic: @Jurigag yes, I missread that, thx. 

```zephir
                        /**
			 * Process conditions
			 */
			if fetch conditions, params[0] {
				let this->_conditions = conditions;
			} else {
				if fetch conditions, params["conditions"] {
					let this->_conditions = conditions;
				}
			}
```
the conjunctive behaviour is only relevant here. 

So just to be save.
```php
BudgetTransaction::findBy(['transaction' => $transaction, 'name' => 'foo']);  // wrong 

BudgetTransaction::findBy(
    [
        'test = "blubb"',
        'transaction' => $transaction
    ]
); // correct since the first position will be taken as query

BudgetTransaction::findBy(
    [
        'conditions' => 'test = "blubb"',
        'transaction' => $transaction
    ]
); // correct since it'S mapped correct

BudgetTransaction::findBy(
    [
        [['field', 'value', 'type'], ['field', 'value', 'type]], // well be generate
        'transaction' => $transaction
    ]
); // correct since the first position will be taken as query

BudgetTransaction::findBy(
    [
        'conditions' => [['field', 'value', 'type'], ['field', 'value', 'type]], // well be generate
        'transaction' => $transaction
    ]
); // correct since the first position will be taken as query
```
@Jurigag do I understand this correct?

chilimatic: @sergeyklay should I just add all those variations as examples?

sergeyklay: @chilimatic yes, please

chilimatic: @sergeyklay is this too much ? or is it okay ?

```php
 // How many robots are there?
use Phalcon\Db\Column;
use Phalcon\Mvc\Model;
use Phalcon\Mvc\Model\Transaction;

$robots = Robots::find();

echo "There are ", count($robots), "\n";

// How many mechanical robots are there?
$robots = Robots::find(
 "type = 'mechanical'"
);

echo "There are ", count($robots), "\n";

// Get and print virtual robots ordered by name
$robots = Robots::find(
    [
         "type = 'virtual'",
         "order" => "name",
    ]
);

foreach ($robots as $robot) {
 echo $robot->name, "\n";
}

// Get first 100 virtual robots ordered by name
$robots = Robots::find([
     "type = 'virtual'",
     "order" => "name",
     "limit" => 100,
]);

foreach ($robots as $robot) {
 echo $robot->name, "\n";
}

// how the find method works
// you can either use it directly with string representing the where clause
// And for more dynamic queries we can use an array
Robots::find("type = 'virtual");

// The internal used construct is as follows:
// If there is no associative index "conditions" it will automatically use the element on position 0
//
// If you use not an array but a string it will be wrapped into an array at position 0
//
// It has to be on index position 0! so be aware that if you modify the PHP array (array_unshift), the index can change.
// The order still could remain and it wont work, same goes for a sort() which will REINDEX an array
//
// A PHP array is not to be confused with other language implementations.
// you can set an explicit position 0 anywhere you want ORDER != POSITION
// PHP7 (https://nikic.github.io/2014/12/22/PHPs-new-hashtable-implementation.html)
//
// so you either always have the query on position 0 OR you explicitly use the "conditions" index

// In an associative manner it looks as follows
// can be changed by reindexing methods (sort,...) but doesn't matter since it's a hashmap and is resolved by name
// FQCNModel stands for Fully qualified Classname -> Robots::class should for example return it
$paramAssociative = [
    'conditions' => '', // or an array
    'transaction' => null, // TransactionInterface
    'bind' => [], // array ['fieldName' => 'fieldValue']
    'bindTypes' => [], // array ['fieldName' => Column::BIND_PARAM_STR]
    'order' => '', // "name desc"
    'distinct' => '', // fieldName
    'models' => '', // will be set by default but in theory you could override the default ['FQCNModel']
    'columns' => '', // fieldName, fieldName1, fieldName2
    'joins' => [], // only relevant in joins
    'group' => [], // [FQCNModel.fieldName, FQCNModel.fieldName1, FQCNModel.fieldName2]
    'having' => '', // ["SUM(Robots.price) > :sum:", [ "sum" => 100]]
    'limit' => '', // [0, 10] or an integer
    'offset' => 0, // integer
    'hydration' => '',
    'cache' => null // cache options
];

// the implicit numeric version would be something along the lines of this
$paramNumericImplicit = [
    '', // or an array
    // .....
]; // it can be changed by reindexing methods (sort,...)

// the explicit numeric version would be something along the lines of this

$paramNumericExplicit = [
    0 => '', // or an array
    // .....
]; // it can be changed by reindexing methods (sort,...)

// so basically the only mistake you make is if you don't use the explicit index conditions
// always correct
Robots::find([
    'conditions' => "name = 'test'"
]);

// but use it implicitly and mess up the ordering
// will not work -> the query would be something like "SELECT .... FROM robots WHERE name desc"
Robots::find([
    'order' => 'name desc',
    "name = 'test'"
]);

// lets focus on the common use cases and use different examples to show the same outcome
// first we just pass in the array -> position 0 will be taken as condition
Robots::find([
    "name = 'test'"
]);

// this is if we would pass the first position as an array
Robots::find([
    [
        'name = :name: OR type = :type: ',
        [
            'name' => 'test',
            'type' => 'virtual'
        ],
        [
            'name' => Column::BIND_PARAM_STR,
            'type' => Column::BIND_PARAM_STR
        ]
    ]
]);

// this is if we would pass the 0 position as an array
Robots::find([
    'order' => 'name desc',
    0 => [
        'name = :name: OR type = :type: ',
        [
            'name' => 'test',
            'type' => 'virtual'
        ],
        [
            'name' => Column::BIND_PARAM_STR,
            'type' => Column::BIND_PARAM_STR
        ]
    ]
]);

// this is if we would pass it explicitly as an array
Robots::find([
    'conditions' => [
        'name = :name: OR type = :type: ',
        [
            'name' => 'test',
            'type' => 'virtual'
        ],
        [
            'name' => Column::BIND_PARAM_STR,
            'type' => Column::BIND_PARAM_STR
        ]
    ]
]);

// As soon as we use an array inside the conditions index,
// we can now dynamically construct a query in a triple manor
//
// every passed parameter in an condition array consists of 3 parts
//
// 1. condition
// 2. values to be bound
// 3. types to be bound
//
// we only need the first 1 if we write explicit queries but lets say for a dynamic query builder or filter mechanismn
// it could make sense to send JSON requests that will be transformed on the server side
//
// Everything that is not statically hardcoded and ensured is considered a side-effect and we should take care
// that we don't let our application gets compromised by such a thing so we use Prepared statements to save guard against
//
// Classic injection vectors but a dynamic filtering process can be nice. the same applies in the backend if we just use basic POST / GET methods
// that's why should separate the statements from the values so we can ensure that the values are not harmful

// it could look as follows
$postParam = json_decode('
[
    {
        "field": "name",
        "value": "test",
        "type": 2
    },
    {
        "field": "type",
        "value": "virtual",
        "type": 2
    }
]
');

$queryParam = [];
foreach ($postParam as $paramSet) {
    $queryParam[] = [
        "${paramSet['field']} = :${paramSet['field']}:",
        [
            $paramSet['field'] => $paramSet['value']
        ],
        [
            $paramSet['field'] => $paramSet['type']
        ]
    ];
}

// or in a more functional manor
$queryParam = array_map(function(array $paramSet) {
    return [
        "${paramSet['field']} = :${paramSet['field']}:",
        [
            $paramSet['field'] => $paramSet['value']
        ],
        [
            $paramSet['field'] => $paramSet['type']
        ]
    ];
}, $postParam);

Robots::find($queryParam); // the query should look something like this SELECT .... FROM robots WHERE name = 'test' AND type = 'virtual';
// All composed array triples will use a conjunction (AND) so if you want to use an OR it has to be in the condition or you should use
// the query builder instead of the Model find method

// as an example
$param = [
    'conditions' => [
        '(this = :this: OR that = :that:)',
        [
            'this' => 'this',
            'that' => 'that'
        ]
    ]
];


// encapsulate find it into an running transaction esp. useful for application unit-tests
// or complex business logic where we wanna control which transaction is used.

$myTransaction = new Transaction(\Phalcon\Di::getDefault());
$myTransaction->begin();
$newRobot = new Robots();
$newRobot->setTransaction($myTransaction);

$newRobot->save([
    'name' => 'test',
    'type' => 'mechanical',
    'year' => 1944
]);


$resultInsideTransaction = Robots::find([
     'condition' => "name = 'test'",
     Model::TRANSACTION_INDEX => $myTransaction
]);

$resultOutsideTransaction = Robots::find([
    "name = 'test'"
]);

foreach ($resultInsideTransaction as $robot) {
    echo $robot->name, "\n";
}

foreach ($resultOutsideTransaction as $robot) {
    echo $robot->name, "\n";
}

// reverts all not commited changes
$myTransaction->rollback();

// creating two different transactions
$myTransaction1 = new Transaction(\Phalcon\Di::getDefault());
$myTransaction1->begin();
$myTransaction2 = new Transaction(\Phalcon\Di::getDefault());
$myTransaction2->begin();

// add a new robots
$firstNewRobot = new Robots();
$firstNewRobot->setTransaction($myTransaction1);
$firstNewRobot->save([
    'name' => 'first-transaction-robot',
    'type' => 'mechanical',
    'year' => 1944
]);

$secondNewRobot = new Robots();
$secondNewRobot->setTransaction($myTransaction2);
$secondNewRobot->save([
    'name' => 'second-transaction-robot',
    'type' => 'fictional',
    'year' => 1984
]);

// this transaction will find the robot.
$resultInFirstTransaction = Robots::find([
    "name = 'first-transaction-robot'",
    Model::TRANSACTION_INDEX => $myTransaction1
]);
// this transaction won't find the robot.
$resultInSecondTransaction = Robots::find([
    "name = 'first-transaction-robot'",
    Model::TRANSACTION_INDEX => $myTransaction2
]);

// this transaction won't find the robot.
$resultOutsideAnyExplicitTransaction = Robots::find([
    "name = 'first-transaction-robot'"
]);

// this transaction won't find the robot.
$resultInFirstTransaction = Robots::find([
    "name='second-transaction-robot'",
    Model::TRANSACTION_INDEX => $myTransaction2
]);

// this transaction will find the robot.
$resultInSecondTransaction = Robots::find([
    "name = 'second-transaction-robot'",
    Model::TRANSACTION_INDEX => $myTransaction1
]);

// this transaction won't find the robot.
$resultOutsideAnyExplicitTransaction = Robots::find([
    "name = 'second-transaction-robot'"
]);

$transaction1->rollback();
$transaction2->rollback();
```

sergeyklay: :thinking: Hmm.. Then better place for this the docs repo

chilimatic: @sergeyklay I see so maybe just 

```php
<?php
// How many robots are there?
use Phalcon\Mvc\Model;
use Phalcon\Mvc\Model\Transaction;

$robots = Robots::find();

echo "There are ", count($robots), "\n";

// How many mechanical robots are there?
$robots = Robots::find(
    "type = 'mechanical'"
);

echo "There are ", count($robots), "\n";

// Get and print virtual robots ordered by name
$robots = Robots::find(
    [
        "type = 'virtual'",
        "order" => "name",
    ]
);

foreach ($robots as $robot) {
    echo $robot->name, "\n";
}

// Get first 100 virtual robots ordered by name
$robots = Robots::find([
    "type = 'virtual'",
    "order" => "name",
    "limit" => 100,
]);

foreach ($robots as $robot) {
    echo $robot->name, "\n";
}


// encapsulate find it into an running transaction esp. useful for application unit-tests
// or complex business logic where we wanna control which transaction is used.

$myTransaction = new Transaction(\Phalcon\Di::getDefault());
$myTransaction->begin();
$newRobot = new Robots();
$newRobot->setTransaction($myTransaction);

$newRobot->save([
    'name' => 'test',
    'type' => 'mechanical',
    'year' => 1944
]);


$resultInsideTransaction = Robots::find([
    'condition' => "name = 'test'",
    Model::TRANSACTION_INDEX => $myTransaction
]);
```
and the rest as PR for the docs ?

chilimatic: @sergeyklay yes? no ? maybe? we can close this before christmas ! :D

sergeyklay (approved):

sergeyklay:
Thank you

PR #13285, opened 2018-01-29
Make shared instances removable.

xueron:
When I use Phalcon  in a multi-process scene,  I have to remove any DB connections copied from parent.

so I add two methods: one check a shared instance for a service exists or not, and another to remove a shared instance. That we can recreate it if needed.

Refs: https://github.com/phalcon/cphalcon/issues/13440

### DIFF
--- a/phalcon/di.zep
+++ b/phalcon/di.zep
@@ -549,4 +549,20 @@ class Di implements DiInterface
 			this->set(name, service, isset service["shared"] && service["shared"]);
 		}
 	}
+
+	/**
+	 * Check whether a shared instance exists.
+	 */
+	public function hasSharedInstance(string! name) -> boolean
+	{
+		return isset this->_sharedInstances[name];
+	}
+
+	/**
+	 * Remove a shared instance from DI. Make it recreateable.
+	 */
+	public function removeSharedInstance(string! name)
+	{
+		unset $this->_sharedInstances[name];
+	}
 }


virgofx:
This is not needed as 2 methods already exist [ has, remove ] which handles this and has proper test cases.

xueron:
@virgofx has() checks a service defined or not, but not check if a shared service created or not.
remove() drop both the service definition and created instance.

I only want to drop the instance, so when I call getShared(), DI will recreate one, as I needed.

virgofx:
You're creating unnecessary code. Your premise is flawed. The service you're checking shouldn't be marked as shared.

xueron:
@virgofx I do register db service as not shared.  But ModelsManager::_getConntion() use getShared() to obtain it. So if the connection copied to child process and closed, parent will get Mysql gone away.

sergeyklay:
@xueron I still don't fully understand. Could you please provide a small code to show use case / real needs?

xueron:
Hi @sergeyklay , I just want to recreate a service instance even if it is already created. 

I'm using phalcon in a multi-process project, DI container instanced in the parent process.

I register 'db' service as not shared. By default modelsManager use getShared() to obtain one instance, and then saved in _sharedInstances. When I fork a child, I have to make a new one to reconnect to db.

`
declare(ticks = 1);
$di = new \Phalcon\Di();
$di->set('db', function () {
    return new \Phalcon\Db\Adapter\Pdo\Mysql([
        'host'     => 'localhost',
        'port'     => 3306,
        'username' => 'root',
        'password' => '',
        'dbname'   => 'test',
        'charset'  => 'utf8',
    ]);
});

$data = $di->getShared('db')->query("select 1");

$pid = pcntl_fork();
if ($pid == -1) {
    die('could not fork');
} else if ($pid) {
    //$di->removeSharedInstance('db');

    // we are the parent
    pcntl_signal(SIGCHLD, function () {
        echo "Child exited";
        pcntl_wait($status, WNOHANG);
    });

    // DO other works
    while (true) {
        $di->getShared('db')->query("select 1"); // Mysql has gone away after child exit.
        sleep(1);
    }
} else {
    // we are the child
    //$di->removeSharedInstance('db');

    $di->getShared('db')->query('select 1');
    sleep(5);
    exit;
}
`

I think if i can remove the shared instance, parent and child can auto reconnect while next getShared() called.

virgofx:
You're using the services wrong --- the database service should ALWAYS be shared. So your initial definition is incorrect.

Secondly, when you fork, your child instances will no longer be able to use any service that held open sockets -- -this is true for databases -  MySQL, Postgres, redis ... there's a few tricks you'll have to do to get forking to work with each, regardless, you'll need to create another instance.

/close

xueron:
@virgofx I remember that database service should not defined as shared if I need to use transaction manager. If do so, transaction manager cannot obtain a new isolate connection.

I think if the db  instance (and any other services that hold sockets/descriptors) saved in DI::$_sharedInstance could be dropped, there is no need to do any other things, getShared() call can establish new connection when needed.

As said above, before I do a fork, just drop the created services that held open sockets, then fork, both parent and child can auto (re)connect.

xueron:
Drop instance before fork may be batter:

```
declare(ticks = 1);
$di = new \Phalcon\Di();
$di->set('db', function () {
    return new \Phalcon\Db\Adapter\Pdo\Mysql([
        'host'     => 'localhost',
        'port'     => 3306,
        'username' => 'root',
        'password' => '',
        'dbname'   => 'test',
        'charset'  => 'utf8',
    ]);
});

$data = $di->getShared('db')->fetchOne("select 1");

// we are the parent
pcntl_signal(SIGCHLD, function ($sig, $siginfo) {
    echo "Child exited\n";
    pcntl_wait($status, WNOHANG);
});

// Drop connections before fork. So no socket copied to child.
$di->removeSharedInstance('db');

$pid = pcntl_fork();
if ($pid == -1) {
    die('could not fork');
} else if ($pid) {

    // DO other works
    while (true) {
        $res = $di->getShared('db')->fetchOne("select 1");
        sleep(1);
    }
} else {
    // we are the child
    $di->getShared('db')->fetchOne('select 1');
    sleep(5);
    echo "byebye\n";
    exit;
}

**Comment by sergeyklay** [comment]
Well, why do you need to call `getShared` and then `removeSharedInstance`? Let me explain real needs of the `getShared` method.

If a service *isn't registered as shared* and you want to be sure that a shared instance will be accessed every time the service is obtained from the DI, you can use the `getShared` method. Otherwise simple use `get` to fetch new instance each time. The `getShared` method was introduced only for these purposes.

**Comment by xueron** [comment]
@sergeyklay Thanks for your reply.

I'm using models, and, modelsManager is now using getShared() to obtain connection service from DI. But I need to do a reconnect when I do a fork. I think remove the shared instance is a simple way.

The above code missing models ...

**Comment by sergeyklay** [comment]
@xueron Please take a look at `Phalcon\Di::remove`:

```zep
	/**
	 * Removes a service in the services container
	 * It also removes any shared instance created for the service
	 */
	public function remove(string! name)
	{
		unset this->_services[name];
		unset this->_sharedInstances[name];
	}
```

Is this satisfies your needs?

virgofx:
I think I can summarize:

In rare cases (e.g. forking) ... a shared service that uses sockets (e.g. Redis, MySQL), may need to retrieve a new instance to a shared service that has been previously registered in the DI container. In this case, one simply wants to remove an existing (and likely corrupted instance) and then retrieve a new instance **without having to redefine the service definition**. 

Thus, the ability to remove the shared instance would allow forking children and implementing classes to remove the shared instance and then follow it up with a proper `$di->getShared('db');` which will return a NEW connection in this case ... but the key part is that it's already been defined in the bootstrap .

If the user removes the `db` from the DI container .... `$di->remove('db');` .... the subsequent call would require first **re**defining the service prior to retrieving a new instance. I believe that's what the issue is and what @xueron is trying to resolve.  

Took a little while to see what was going on; however, I can see a use case for this.

sergeyklay:
Well, what do you think about this:

```diff
	/**
	 * Registers a service in the services container
	 */
	public function set(string! name, var definition, boolean shared = false) -> <ServiceInterface>
	{
		var service;
		let service = new Service(name, definition, shared),
			this->_services[name] = service;

++		unset this->_sharedInstances[name];
++
		return service;
	}
```

sergeyklay:
Well, at least it isn't `removeShared`. Most likely it is `unshare` or something like this. Also, just to you know guys we don't plan to release v4.0.0 ASAP but we don't add new features in minor releases.

xueron:
@virgofx @sergeyklay Thanks a lot. That's what I want :)   I have create a subclass that implements this in my project and works well. Hope it comes within v4.0 ^_^

sergeyklay:
@dragoonis Just assigned to you since you are currently working on ContainerInterface implementation in context of Phalcon's DI/ServiceLocator

niden:
Reading through the comments this is not necessary because it will be addressed with the introduction of a new PSR-11 compliant container.

Closing this

niden:
Thank you @xueron for the effort and discussion

Issue #12648, opened 2017-02-24
Variable scope when is set as view param and rendered by getRender()


slechtic:
When I set view params and call getRender, the variables are visible in the view but out of their scope (in action) too.

```php
$di->set('view', function () use ($config) {

    $view = new View();

    $view->setViewsDir($config->application->viewsDir);

    $view->registerEngines(array(
        '.volt' => function ($view, $di) use ($config) {

            $volt = new VoltEngine($view, $di);

            $volt->setOptions(array(
                'compiledPath' => $config->application->cacheDir,
                'compiledSeparator' => '_'
            ));

            return $volt;
        },
        '.phtml' => 'Phalcon\Mvc\View\Engine\Php'
    ));

    $view->setParamToView('paramFromService', 'Param from service');

    return $view;
}, true);

public function indexAction()
{
    $content = $this->view->getRender('index', 'index', [
        'viewParam' => 'Test view param'
    ]);
    var_dump($content);
    var_dump($viewParam);
    var_dump($paramFromService);
    die;
}
```
* Variable $content is corrent.
* Variable $viewParam contains string 'Test view param'. I expected notice Undefined variable.
* Variable $paramFromService is assign to view in service.php and is accessible in this action too. I expected notice Undefined variable.

### Testing environment

* Phalcon version: 3.0.4
* PHP Version: 5.6.24 as Apache module, 7.0.16. as Fcgi
* Operating System: Windows 10
* Server: Apache 2.4

### Source files
https://github.com/slechtic/phalcon_variable_scope

Jurigag:
Does this happen for 5.6.24 too?

slechtic:
Yes, for both version of PHP.

jesseforrest:
@sergeyklay : I was wondering if `Bug - Medium` refers to medium complexity or medium priority.  If it's related to priority, I would like to push that this is high-priority for an MVC framework.  imo, variable scope for views seems like an important topic for an MVC framework to get right.  

Phalcon might be performing faster than it actually should be, since it likely is not be making copies of variables for each view, which may be required to resolve this problem.  Either way, I just want to voice that I think this one is larger than it might seem (in complexity and priority), in order to do it right where it doesn't add a lot of overhead.

Jurigag:
Well tbh as far as i see this https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/view/engine/volt.zep#L114

[...]

Issue #12176, opened 2016-08-29
Diffrent volt partial() in php 5.6/7.0


FlexIDK:
### Expected and Actual Behavior

In PHP 5.6

Volt: 
`partial("email/tag_open", {'type': 'p'})`  - `$type` is local  var
`{{ type }}` - not set
`partial("email/tag_open", {})` - `$type` inside - not set

In PHP 7

Volt: 
`partial("email/tag_open", {'type': 'p'})`  - $type is global  var
`{{ type }}` - last state set
`partial("email/tag_open", {})` - $type inside - last state set

### Details
- Phalcon version: 3.0.1
- PHP Version: 5.6.24-0+deb8u1 / 7.0.10-1~dotdeb+8.1
- Operating System: Debian 8.5
- Installation type: Compiling from source

### Refs
- https://github.com/phalcon/zephir/issues/1621

andresgutierrez:
PHP 5 allowed us to create intermediate symbol tables per render, in php7 that is not possible, we have to research more about it.

janisbizkovskis:
I would like to suggest this fixing ASAP, as this is braking all the partial system in the views. (quite important thing)

challet:
It's also driving the [loop context](https://docs.phalconphp.com/en/3.0.0/reference/volt.html#loop-context) crazy :

```volt
{% for d in data %}
    {{ dump(loop.first) }} // true 
    // there is a {% for t in tags %} inside it
    {% include "header" with ['active': loop.first, 'tags': d.tags] %} 
    {{ dump(active) }} // true 
    {{ dump(loop.first) }} // false 
{% endfor %}
```

jl91:
some news about it? someone knows  when it will be fixed? what branch ? or date? i've identified this problem in phalcon 3.0.0, but now i'm migrating to phalcon 3.0.3!

Jurigag:
Im guessing if someone would know then he would do PR with fix. We have to wait until @andresgutierrez will do something about this.

jellisii:
Can confirm this is still a problem, particularly the loop context that @challet referenced.  I can put together a test case if needed, as this is causing me headaches.

janisbizkovskis:
Can confirm as now. (23.11.2017). Due to this issue cannot upgrade two of my projects to php 7+ :( Same behavior is with .phtml, so @sergeyklay `Volt` label can be removed, as it's affecting all kind of views.

sergeyklay:
I'll try to sort out

[...]

PR #13288, opened 2018-01-30
Fix unit test for issue #12648 - Incorrect scope of view variables

piit79:
Hello!

* Type: bug fix
* Link to issue: #12648

Small description of change:
The unit test for issue #12648 is incorrect and doesn't test the failure condition of the issue properly. The issue is marked as fixed but the problem is still present in 3.3.x.

Thanks
Petr Sedlacek

### DIFF
--- a/tests/unit/Mvc/ViewTest.php
+++ b/tests/unit/Mvc/ViewTest.php
@@ -1022,6 +1022,7 @@ class ViewTest extends UnitTest
      *
      * @issue  12648
      * @author Wojciech Ślawski <jurigag@gmail.com>
+     * @author Petr Sedlacek <petr@sedlacek.biz>
      * @since  2017-03-17
      */
     public function testIssue12648()
@@ -1040,6 +1041,7 @@ class ViewTest extends UnitTest
                 expect($content)->equals("<html>lol<p>test</p></html>\n");
                 try {
                     echo $a_cool_var;
+                    $this->fail('Variable a_cool_var is defined and is set to "' . $a_cool_var . '"');
                 } catch (\PHPUnit_Framework_Exception $e) {
                     expect($e->getMessage())->contains("Undefined variable: a_cool_var");
                 }


piit79:
@sergeyklay Just to be sure, could you please take a look at the unit test and confirm it's correct? It seems strange to me nobody has complained about #12648 which makes me think if we're doing something wrong :) We're using parametric partials which does not work with PHP 7 (the variables leak between partials). Thanks for your help on this!

sergeyklay:
/ cc @Jurigag

sergeyklay:
@piit79 As I can see it happens only in Zend Engine 3 (PHP 7).

Refs: https://github.com/phalcon/zephir/issues/1621, https://github.com/phalcon/cphalcon/issues/12176

piit79:
@sergeyklay that's right, we only started having the issue when we switched to PHP 7 recently (with the same code). Thanks for looking into this.

sergeyklay:
@piit79 I'll try to sort out with this

sergeyklay:
I hope we managed to sort out with this issue finally here https://github.com/phalcon/cphalcon/pull/13606. All necessary changes are already in the appropriate branch. We will try to release the next version in the coming days.  Thank you for your patience, the report, and for helping us make Phalcon better.

piit79:
@sergeyklay That's great! Can you please point me to the branch? I'll do a quick test.

However, I see you haven't merged the fix for the unit test, can I ask why? It's still broken in master:

                try {
                    echo $a_cool_var;
                } catch (\PHPUnit_Framework_Exception $e) {
                    expect($e->getMessage())->contains("Undefined variable: a_cool_var");
                }

The problem is that if the variable `$a_cool_var` **is** actually set (which it should not), there will be no exception and the test will not catch this failure condition. That's why I added the `fail()` after the `echo`.

sergeyklay:
Please see: https://github.com/phalcon/cphalcon/issues/12648#issuecomment-440808113

Issue #13315, opened 2018-03-08
The detailed mode for Phalcon/Debug is not work


limingxinleo:
> **When I use Phalcon/Debug detailed mode,The result is not expected.**

```php
<?php
        $obj = new \stdClass();
        $obj->aaa = 111;

        array_map(function ($x) {
            $string = (new \Phalcon\Debug\Dump(null, true))->variable($x);

            echo(PHP_SAPI == 'cli' ? strip_tags($string) . PHP_EOL : $string);

        }, [$obj]);
```
> The Result

```
The argument is not initialized or iterable() code:0 in phalcon/debug/dump.zep line 215
#0 [internal function]: Phalcon\Debug\Dump->output(Object(stdClass), '')
#1 /Users/limx/Applications/limingxinleo/basic-phalcon/app/Tasks/TestTask.php(16): Phalcon\Debug\Dump->variable(Object(stdClass))
#2 [internal function]: App\Tasks\TestTask->App\Tasks\{closure}(Object(stdClass))
#3 /Users/limx/Applications/limingxinleo/basic-phalcon/app/Tasks/TestTask.php(20): array_map(Object(Closure), Array)
#4 [internal function]: App\Tasks\TestTask->mainAction(Array, Array)
#5 [internal function]: Phalcon\Cli\Dispatcher->callActionMethod(Object(App\Tasks\TestTask), 'mainAction', Array)
#6 [internal function]: Phalcon\Dispatcher->dispatch()
#7 /Users/limx/Applications/limingxinleo/basic-phalcon/vendor/limingxinleo/x-phalcon-console/src/Cli/XConsole.php(36): Phalcon\Cli\Console->handle(Array)
#8 /Users/limx/Applications/limingxinleo/basic-phalcon/run(47): Xin\Phalcon\Cli\XConsole->handle(Array)
#9 {main}%
```

I find the source code for Phalcon/Debug/Dump, using `for` instead of `each`, `for` is not support object, so it throw a exception, but `each` is deprecated for php7.2.

### Details

* Phalcon version: (`3.3.2`)
* PHP Version: (`7.1.14`)
* Operating System: MAC
* Installation type: Compiling from source || installing via package manager
* Zephir version (if any):
* Server: Nginx | Apache | Other
* Other related info (Database, table schema):

sergeyklay:
cc: @jmsfwk

sergeyklay:
[...]

PR #13308, opened 2018-02-26
Replace deprecated `each()` function with for...in loop

jmsfwk:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/13253

**In raising this pull request, I confirm the following (please check boxes):**

- [x] I have read and understood the [Contributing Guidelines](https://github.com/phalcon/cphalcon/blob/master/CONTRIBUTING.md)?
- [x] I have checked that another pull request for this purpose does not exist.
- [ ] I wrote some tests for this PR.

Small description of change:

The `each()` function in `Phalcon\Debug\Dump` is deprecated. I have replaced it with of `for...in` loop.

Thanks

PR #13316, opened 2018-03-08
Delete detailed mode for Class `Phalcon\\Debug\\Dump` , because the protected and private vars of the object should not be obtained outside.

limingxinleo:
Hello!

* Type: bug fix
* Link to issue: #13315

**Delete detailed mode for Class `Phalcon\\Debug\\Dump` when php version >= 7.2, because I can not get the object protected or private vars.**

Small description of change:

Thanks


tests/unit/DebugTest.php
```diff
@@ -49,4 +50,33 @@ function () {
...
+                    (new Dump(null, true))->variable([1, 2, 3]);
+
+                    $obj = new \stdClass();
+                    $obj->test = 'value';
+                    (new Dump(null, true))->variable($obj);
+
+                    expect(true)->true();
+                } catch (\Exception $ex) {
+                    expect(false)->true();
+                }
```
sergeyklay: Here is not necessary to use try/catch because tests will fail in case of exception.

phalcon/debug/dump.zep
```diff
@@ -193,10 +195,21 @@ class Dump
 				}
 			} else {
 				// Debug all properties
-				for key, value in variable {
+                do {
+
+                    let attr = each(variable);
```
sergeyklay: Could you please avoid `each` function. See: https://github.com/phalcon/cphalcon/pull/13308

tests/unit/DebugTest.php
sergeyklay: Here is not necessary to use try/catch because tests will fail in case of exception.

sergeyklay (changes requested):

tests/unit/DebugTest.php
limingxinleo: I see , thanks

phalcon/debug/dump.zep
limingxinleo: Sorry, I not find a method to replace each.So I remove detailed mode for `Phalcon\\Debug\\Dump `. When php version < 7.2, it works.

sergeyklay: Anyway this code will introduce `E_DEPRECATED` error

limingxinleo: I see, maybe someone can fix it later. I'll try it, too.

limingxinleo: Thanks

limingxinleo:
This is not a good PR, I think the best solution should start with Zephir, not cphalcon. But my level is limited and I'm sorry.

sergeyklay:
@limingxinleo Closing this in favor of https://github.com/phalcon/cphalcon/pull/13332. 
Thank you for the report, and for helping us make Phalcon better.

PR #13343, opened 2018-04-11
Fixed invalid parameter number issue while passing non-sequential array 

uzrnem:
Hello!

* Type: bug fix
SQLSTATE[HY093]: Invalid parameter number: parameter was not defined
SQL Error was coming, while binding non-sequential array to phalcon query builder


sergeyklay:
@uzrnem Could you please provide a small test which will cover these changes?

uzrnem:
Please tell me how can i provide you test of this changes in detail

sergeyklay:
@Jurigag Could you please take a look?

sergeyklay:
@uzrnem Could you please provide script to reproduce the issue? Actually `$params` for `Phalcon\Mvc\Model\Query\Builder` constructor will be enough

uzrnem:
`$models = Model::find(
  [
    "id IN ({ids:array})",
    "bind" => [
      "scacs" => [
        0 => 1,
        2 => 3,
        8 => 4
      ]
      /*
      Here 1, 3, 4 are the Ids to be find 
      and 0, 2, 8 are the index that are non-sequential, 
      because we use functions like array_diff
      */
    ]
  ]
);` try this code

PR #13353, opened 2018-04-20
Update doc return types ResultsetInterface

Joilson:
Hello!

* Type: documentation


virgofx:
Thanks --- I do believe these only return `false` and don't have capability to return `true` so we should switch it to `bool|\Phalcon\Mvc\ModelInterface`

sergeyklay:
Fixed in the `3.4.x` branch.  Thank you for contributing.

Issue #13360, opened 2018-04-29
[NFR] Improve Phalcon\Autoloader speed


dugwood:
### Expected and Actual Behavior

Phalcon\Autoloader is slow because the `is_file()` test is slow. With a lot of files, this can add a serious amount of time, especially if the whole PHP run is small.

See #10472 for instance, and specifically comment: https://github.com/phalcon/cphalcon/issues/10472#issuecomment-384090810

Reproduce script:
```php
$time = -1 * microtime(true);
include __DIR__.'/../library/ModelParameters.php';
include __DIR__.'/../models/Model.php';
include __DIR__.'/../models/Promotion.php';
$files = 3;
foreach (glob(__DIR__.'/../models/*.php') as $file)
{
	$class = 'MyNamespace\\Models\\'.substr(basename($file), 0, -4);
	if ($class === 'MyNamespace\\Models\\Model' || $class === 'MyNamespace\\Models\\Promotion')
	{
		continue;
	}
	include $file;
	$files++;
}
echo ($time + microtime(true)) * 1000, ' - files: '.$files;
```
=> for 120 files
=> first load: about 120ms
=> next loads: **4-8ms** (opcache must play a part here)

```php
$time = -1 * microtime(true);
$loader = new Loader();
$loader->registerNamespaces([
	'MyNamespace\Controllers' => __DIR__.'/controllers/',
	'MyNamespace\Backend\Controllers' => __DIR__.'/../backend/controllers/',
	'MyNamespace\Frontend\Controllers' => __DIR__.'/../frontend/controllers/',
	'MyNamespace\Models' => __DIR__.'/../models/',
	'MyNamespace\Library' => __DIR__.'/../library/',
	'MyNamespace\Plugins' => __DIR__.'/../plugins/',
]);
$loader->register();
echo ($time + microtime(true)) * 1000;
```
=> takes less than a millisecond as expected

```php
$time = -1 * microtime(true);
$files = 2;
foreach (glob(__DIR__.'/../models/*.php') as $file)
{
	$class = 'MyNamespace\\Models\\'.substr(basename($file), 0, -4);
	class_exists($class);
	$files++;
}
echo ($time + microtime(true)) * 1000, ' - files: '.$files;
```
=> for 121 files
=> first load: 250ms
=> next loads: **100-150ms** (about 1ms per loaded class)

[...]

PR #13361, opened 2018-04-29
Added Phalcon\Loader::setFileCheckingCallback

sergeyklay:
Hello!

* Type: new feature
* Link to issue: #13360

Small description of change:

Added `Phalcon\Loader::setFileCheckingCallback` to set internal file existence resolver.
```php
// Default behavior.
$loader->setFileCheckingCallback("is_file");

// Faster than `is_file()`, but implies some issues if
// the file is removed from the filesystem.
$loader->setFileCheckingCallback("stream_resolve_include_path");

// Do not check file existence.
$loader->setFileCheckingCallback(null);
```

Thanks

----

/cc @dugwood

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,7 +45,7 @@ lemon
 .libs/
 .temp/
 autom4te.cache/
-vendor/
+/vendor/
 ide/
 
 composer.lock
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,7 @@ env:
   global:
     - CC="ccache gcc"
     - PATH="$PATH:$HOME/bin"
-    - ZEPHIR_PARSER_VERSION="v1.1.1"
+    - ZEPHIR_PARSER_VERSION="v1.1.2"
     - RE2C_VERSION="1.0.3"
     - REPORT_EXIT_STATUS=1
     - NO_INTERACTION=1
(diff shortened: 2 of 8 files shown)


dugwood:
Should be `Callback` instead of `Callaback` :-)

sergeyklay:
@dugwood  I will keep this open for a day or two to allow other contributors a chance to speak up, and to take a fresh look at it later. Please feel free to relay your comments, suggestions or corrections

sergeyklay:
@dugwood Just regenerated build for PHP7 and PHP5. So you can install current PR using `cd build && ./install`. Please use my own fork to test.

dugwood:
Thanks, I've got the building step okay before, so now it's working as expected.

Runs for 121 files (as in my bug report #13360):
 - default or `is_file`: about 100-150ms
- `stream_resolve_include_path`: 10-150ms (mostly the first call is slow, afterwards it's ~8ms)
- `null`: same as `stream_resolve_include_path`, which is weird, isn't it?

dugwood:
Without opcache:

 - default or `is_file`: about 350-500ms
 - `stream_resolve_include_path`: 250-400ms
 - `null`: same as `stream_resolve_include_path`

So `null` isn't that useful in the end, but `stream_resolve_include_path` is anyway, that's a big win.

dugwood:
Real life example with opcache and a script that loads 24 classes from `Phalcon::autoLoad()`:
 - default: 28ms in autoLoad, 30% of total script run: 94ms
 - stream_resolve_include_path: either 34% or 3%... depending on the call and most probably on the fact that opcache keep files cached for 2 seconds (`opcache.revalidate_freq = 2`).
 - stream_resolve_include_path: always 3% with `opcache.validate_timestamps = 0`. That's of course the best setting, but this adds the need for an opcache flush.

Never the less, that's a good patch to me :-)

Thanks!

sergeyklay:
Thank you for so good research

Issue #11029, opened 2015-10-17
[2.1.x] PHP 5.5.9-1ubuntu4.13 Trouble binding to closure when the source of the function was a static class method


dschissler:
* Phalcon - `7fdc8831d7bb80e84b17b9be6eefec809dc70e28` - latest
* Zephir - `3386a95bedd1e27e0cef0683e1e91224089aebac` - latest
* PHP - PHP 5.5.9-1ubuntu4.13

`PHP Warning:  Cannot bind an instance to a static closure in ...`

This is in my private project but I based it off of publicly accessible code in my Webird project, so it should be the same issue.

I'm using a static method in the base Module class to create the `view` service function.  Then each module calls this static method that is then set with module specific view paths (in addition to the common view paths).  This was working great in Phalcon 1.3 and 2.0 and it allows me to pre-compile all Volt templates for the build process.  The method needs to be public static because I make the typical assumption that only a single module will be called during execution cycle [and the CLI script to create the production build needs to be able to access all of the module view functions] .

* The [view function](https://github.com/perchlayer/webird/blob/master/app/phalcon/common/library/Module.php#L27) in the base Module class
* The web module [setting the service](https://github.com/perchlayer/webird/blob/master/app/phalcon/modules/web/Module.php#L49)

So you can see that function returned is not static and it should appear just as a normal function.  This seems like some kind of regression in all of the cool closure stuff that appeared in PHP 5.4 and 5.5.

I have no idea why it works like this.  I suspect that PHP 5.5 closure stuff is still just internal hacks to make it seem like their is pure support.  Zephir or Phalcon are perhaps interpreting something too strictly.

Would this work in PHP 5.6 or 7.0?

dschissler:
I inlined the static function into the called module class and it stopped complaining about it at that location but it still fails at `echo $application->handle()->getContent();`

So again, is this a new convention that I need to adhere to or is this a bug?

dschissler:
I recompiled Phalcon `2.0.8` with Zephir `0.9a-dev` and my code works again.  So, its a Phalcon issue.

dschissler:
Also I'll add that the service closure wasn't able to bind the DI as `$this` in the closure.  I suspect that code is an issue.  The function comes from a static function but is itself just a anonymous function assigned to a variable and then passed in.  This is interfering with the DI being able to bind to it.

What version of PHP are the devs using?  PHP 5.4+ is supposed to be supported in this version of Phalcon.

[...]

PR #13395, opened 2018-06-01
Add static closure to Service resolution

sergeyklay:
Refs: #11709, #11710, #11029


virgofx:
Will this impact performance? Reflection has always been slow ...

sergeyklay:
@virgofx I'll research this question

sergeyklay:
I did some research and I wasn't able to identify any problems described in the previous issues  #11709, #11710 and #11029. In my tests I used all the code samples that were provided in the issues. I didn't use PHP 5 due to the fact that this version is no longer supported by Phalcon 4.x. Therefore, I have to close this PR. I'll open this PR again if needed arises.

/cc @ogarbe

PR #13413, opened 2018-06-21
3.4.x Fix performance of getRouteByName/Id

Jurigag:
Hello!

* Type: bug fix
* Link to issue:

Small description of change: Actually i would consider it as a hotfix because currently it's pretty big performance hit for big applications which are generating many urls using phalcon services and have many routes. In our app we have 725 routes and we have 50 links generation for single page. It was taking 270ms for doing this. With similar changes like here now it's taking 0.133ms(cached) or 18ms(not cached).

There are already tests for getRouteByName/getRouteById, it won't change anyhow logic of this, well the only case is if we after adding two routes we will swap names of them after already using getRouteByName for their names - but im not sure even why do this and this is a very rare case.

Thanks

### DIFF
--- a/CHANGELOG-3.4.md
+++ b/CHANGELOG-3.4.md
@@ -1,5 +1,6 @@
 # [3.4.1](https://github.com/phalcon/cphalcon/releases/tag/v3.4.1) (2018-XX-XX)
 - Fixed `Phalcon\Validaiton\Validator\Uniqueness::isUniquenessModel` to properly get value of primary key when it has different name in column map [#13398](https://github.com/phalcon/cphalcon/issues/13398)
+- Fixed bad performance for repeated `Phalcon\Mvc\Router::getRouteByName` and `Phalcon\Mvc\Router::getRouteById` calls for applications with many routes
 
 # [3.4.0](https://github.com/phalcon/cphalcon/releases/tag/v3.4.0) (2018-05-28)
 - Added `Phalcon\Mvc\Router::attach` to add `Route` object directly into `Router` [#13326](https://github.com/phalcon/cphalcon/issues/13326)
--- a/phalcon/mvc/router.zep
+++ b/phalcon/mvc/router.zep
@@ -95,6 +95,10 @@ class Router implements InjectionAwareInterface, RouterInterface, EventsAwareInt
 
 	protected _notFoundPaths;
 
+	protected _keyRouteNames = [] { get, set };
+
+	protected _idRouteNames = [] { get, set };
+
 	const URI_SOURCE_GET_URL = 0;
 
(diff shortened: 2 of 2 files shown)


sergeyklay (approved):

michanismus:
Did you actually test your code?
I get a lot of exceptions!

First of all, why use "_idRouteNames"? You don't even fill it, you fill "_keyRouteNames". There is no performance boost because you always refill "_keyRouteNames" with every call of getRouteById().

Second, and much bigger bug. You use "let size = count(this->_routes);" and later iterate through "range(0, size)" which will en in "undefined index error" because key != size it is actually size-1 for the last key.

michanismus:
There should also be a check if the route names are already collected. Otherwise the name and id cache is useless because it will iterate through all routes if id or name is not available!

For getRouteByName() there should be a check for the existence of a name. The name is optional.  Otherwise there will be an empty key in _keyRouteNames.

michanismus:
my working fixed code:

```
    protected _keyRouteNames { get, set };

    protected _idRouteNames { get, set };


    /**
     * Returns a route object by its id
     */
    public function getRouteById(var id) -> <RouteInterface> | boolean
    {
        var route, routeId, key;

        if empty this->_routes {
            return false;
        }

        if typeof this->_idRouteNames == "array" {
            if fetch key, this->_idRouteNames[id] {
                return this->_routes[key];
            }

            return false;
        }

        let this->_idRouteNames = [];

        for key, route in this->_routes {
            let routeId = route->getRouteId();
            let this->_idRouteNames[routeId] = key;
        }

        return this->getRouteById(id);
    }

    /**
     * Returns a route object by its name
     */
    public function getRouteByName(string! name) -> <RouteInterface> | boolean
    {
        var route, routeName, key;

        if empty this->_routes {
            return false;
        }

        if typeof this->_keyRouteNames == "array" {
            if fetch key, this->_keyRouteNames[name] {
                return this->_routes[key];
            }

            return false;
        }

        let this->_keyRouteNames = [];

        for key, route in this->_routes {
            let routeName = route->getName();
            if !empty routeName {
                let this->_keyRouteNames[routeName] = key;
            }
        }

        return this->getRouteByName(name);
    }
```

michanismus:
What about routes which are named later? Unlikely but possible at the moment.
The route itself should notice the router that there is a name now for the route to invalidate the whole cache.
Cache also needs to be flushed after a route was added or group was mounted!

My opinion after reviewing all these id/name cache stuff: **Revert it!** 
Try another approch with v4 and keep everything untouched in 3.x!

Jurigag:
Oh, you are right, i missed there a line.

PR #13420, opened 2018-06-28
3.4.x fix getRouteById and add test

Jurigag:
Hello!

* Link to issue:

Small description of change:

Thanks

### DIFF
--- a/phalcon/mvc/router.zep
+++ b/phalcon/mvc/router.zep
@@ -973,10 +973,10 @@ class Router implements InjectionAwareInterface, RouterInterface, EventsAwareInt
 
 		let size = count(this->_routes);
 
-		for key in range(0, size) {
+		for key in range(0, size - 1) {
 			let route = this->_routes[key];
 			let routeId = route->getRouteId();
-			let this->_keyRouteNames[routeId] = key;
+			let this->_idRouteNames[routeId] = key;
 
 			if routeId == id {
 				return route;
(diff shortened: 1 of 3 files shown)


Jurigag:
@michanismus @sergeyklay

michanismus:
@Jurigag Why do you use the range() function? If there is no route the function will return an array with the elements [0, -1] what will end in an undefined index error again.
And again, you don't check for a route name in getRouteByName().

My suggestion, use my code I posted:

```
    protected _keyRouteNames = [] { get, set };

    protected _keyRouteIds = [] { get, set };

    /**
     * Returns a route object by its id
     */
    public function getRouteById(var id) -> <RouteInterface> | boolean
    {
        var route, routeId, key;

        if empty this->_routes {
            return false;
        }

        if fetch key, this->_keyRouteIds[id] {
            return this->_routes[key];
        }

        for key, route in this->_routes {
            let routeId = route->getRouteId();
            let this->_keyRouteIds[routeId] = key;

            if routeId == id {
                return route;
            }
        }

        return false;
    }

    /**
     * Returns a route object by its name
     */
    public function getRouteByName(string! name) -> <RouteInterface> | boolean
    {
        var route, routeName, key;

        if empty this->_routes {
            return false;
        }

        if fetch key, this->_keyRouteNames[name] {
            return this->_routes[key];
        }

        for key, route in this->_routes {
            let routeName = route->getName();
            if !empty routeName {
                let this->_keyRouteNames[routeName] = key;

                if routeName == name {
                    return route;
                }
            }
        }

        return false;
    }

```

Jurigag:
Well, `Phalcon\Mvc\Router::getRouteById` is kind of pointless since `Phalcon\Mvc\Router\Route` doesn't have `getId()` method or i miss something?

Jurigag:
@michanismus i guess it;'s okay now?

sergeyklay (approved):

sergeyklay:
Thank you

Jurigag:
Well actually you are right about range, i forgot about this.

Issue #13450, opened 2018-07-25
[BUG] Numericality validator does not work in some locales


mervick:
### Expected and Actual Behavior

Numericality validator returns validation error 'Field :field does not have a valid numeric format' for valid float number when php uses locales which uses coma `,` in locality number format, for example `ru_RU.utf8` locale

The main problem is that Numericality validator uses regexp:
https://github.com/phalcon/cphalcon/blob/adfcef56c06ab26b41dfb4620c8a21f843c5f8b2/phalcon/validation/validator/numericality.zep#L74

Code below perfectly demonstrates this bug
(make sure that you installed both `en_US.utf8` and `ru_RU.utf8` locales in your system)
```php
$regexp = '/^-?\d+\.?\d*$/';
$number = 123.12;

$locale = setlocale(LC_ALL, 'en_US.utf8');
echo $locale; // en_US.utf8
echo $number; // 123.12
var_dump(preg_match($regexp, $number)); // int(1)

$locale = setlocale(LC_ALL, 'ru_RU.utf8');
echo $locale; // ru_RU.utf8
echo $number; // 123,12
var_dump(preg_match($regexp, $number)); // int(0)
```

> Provide minimal script to reproduce the issue

```php
class TestForm extends \Phalcon\Forms\Form {
    public function initialize()
    {
        $el = new \Phalcon\Forms\Element\Hidden('price');
        $el->addValidator(new \Phalcon\Validation\Validator\Numericality());
        $this->add($el);
    }
}

function testNumericalityWithLocale($locale)
{
    $setlocale = setlocale(LC_ALL, $locale);

    if ($setlocale !== $locale) {
        throw new \Exception(sprintf('Looks like you need to install %s locale on your OS', $locale));
    }

    $price = 123.12;

    $form = new TestForm();
    $data = ['price' => $price];

    echo 'Price is ' . $price . PHP_EOL;

    if (!$form->isValid($data)) {
        throw new \Exception($form->getMessagesFor('price')[0]->getMessage());
    }

    echo 'OK', PHP_EOL;
}

testNumericalityWithLocale('en_US.utf8');
testNumericalityWithLocale('ru_RU.utf8');
```

Output:
```
php > testNumericalityWithLocale('en_US.utf8');
Price is 123.12
OK

php > testNumericalityWithLocale('ru_RU.utf8');
Price is 123,12
PHP Warning:  Uncaught Exception: Field price does not have a valid numeric format in php shell code:18
Stack trace:
[...]

PR #13457, opened 2018-08-02
Fix issue with validation on locales where comma used

mervick:
fix https://github.com/phalcon/cphalcon/issues/13450

Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/13450

Thanks

### DIFF
--- a/phalcon/validation/validator/numericality.zep
+++ b/phalcon/validation/validator/numericality.zep
@@ -71,7 +71,7 @@ class Numericality extends Validator
 
 		let value = validation->getValue(field);
 
-		if !preg_match("/^-?\d+\.?\d*$/", value) {
+		if !preg_match("/^-?\d+(?:[\.,]\d+)?$/", value) || !is_numeric(value) {
 			let label = this->prepareLabel(validation, field),
 				message = this->prepareMessage(validation, field, "Numericality"),
 				code = this->prepareCode(field);
--- a/CHANGELOG-3.4.md
+++ b/CHANGELOG-3.4.md
@@ -1,3 +1,6 @@
+# [3.4.2](https://github.com/phalcon/cphalcon/releases/tag/v3.4.2) (2018-XX-XX)
+- Fixed `Phalcon\Validation\Validator\Numericality` to accept float numbers on locales with comma decimal point [#13450](https://github.com/phalcon/cphalcon/issues/13450)
+
 # [3.4.1](https://github.com/phalcon/cphalcon/releases/tag/v3.4.1) (2018-08-04)
 - Changed `Phalcon\Cache\Backend\Redis` to support connection timeout parameter
 - Fixed `Phalcon\Validaiton\Validator\Uniqueness::isUniquenessModel` to properly get value of primary key when it has different name in column map [#13398](https://github.com/phalcon/cphalcon/issues/13398)
--- a/tests/unit/Validation/Validator/NumericalityTest.php
+++ b/tests/unit/Validation/Validator/NumericalityTest.php
@@ -10,11 +10,12 @@ use Phalcon\Validation\Validator\Numericality;
  * \Phalcon\Test\Unit\Validation\Validator\NumericalityTest
  * Tests the \Phalcon\Validation\Validator\Numericality component
  *
- * @copyright (c) 2011-2017 Phalcon Team
+ * @copyright (c) 2011-2018 Phalcon Team
  * @link      http://www.phalconphp.com
  * @author    Andres Gutierrez <andres@phalconphp.com>
  * @author    Nikolaos Dimopoulos <nikos@phalconphp.com>
  * @author    Wojciech Ślawski <jurigag@gmail.com>
+ * @author    Andrey Izman <izmanw@gmail.com>
  * @package   Phalcon\Test\Unit\Validation\Validator
  *
  * The contents of this file are subject to the New BSD License that is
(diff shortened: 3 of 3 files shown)


phalcon/validation/validator/numericality.zep
```diff
@@ -71,7 +71,7 @@ class Numericality extends Validator
 
 		let value = validation->getValue(field);
 
-		if !preg_match("/^-?\d+\.?\d*$/", value) {
+		if !preg_match("/^-?\d+[\.,]?\d*$/", value) || !is_numeric(value) {
```
sergeyklay: In general looks good but I fear for already working applications that expects that `99,99` is not numeric. Also as I know the `is_numeric` function rely on locale setup. Finally, introducing `is_numeric` will also "say" that `1.45e99` is number (previously not).

mervick: It will test first is `99.99` or `99,99`, so `is_numeric` used only for checks if `99.` or `99,` not used because it's not a right numeric format. 

>  I fear for already working applications that expects that 99,99 is not numeric.

It exactly a problem, it's a bug. On some locales it must be a proper number format.
Becides `is_numeric` uses locale format, so for `en_US.utf8` number `99,99` is still not valid, but it will be valid for `ru_RU.utf8` which now is not working

mervick: It test both regexp and `is_numeric`, so in case if one of them returns false - it false,
so it does not except `1.45e99` and will except only valid numbers for current locale

Jurigag:
I think we need some more extensive tests with locale changing etc and also update changelog.

sergeyklay:
What do you think guys about introducing an _optional parameter_ to set a regular expression to validate number? For example it may looks like this:
```php
$validator->add(
    [
        "price",
        "amount",
    ],
    new Numericality(
        [
            "message" => [
                "price"  => "price is not numeric",
                "amount" => "amount is not numeric",
            ],
            "regexp" => [
                "price"  => "/^-?\d+(\.?\d*(e[-+]?\d*))?$/", // 1.45e-99
                "amount" => "/^-?\d+(\.\d{1,})?$/",          // 1123123.3
            ],
        ]
    )
);
```

Also I agree with @Jurigag about tests and change log. @mervick could you please amend tests and update change log aw well?

mervick:
> Also I agree with @Jurigag about tests and change log. @mervick could you please amend tests and update change log aw well?

Yes, I can amend tests, and not sure about logs

mervick:
> What do you think guys about introducing an optional parameter to set a regular expression to validate number? For example it may looks like this:

As for me I think for this reason there is [RegExp](https://github.com/phalcon/cphalcon/blob/3.4.x/phalcon/validation/validator/regex.zep) validator

sergeyklay:
Well, okay. Let's cover this by tests and add these changes to the [CHANGELOG-3.4.md](https://github.com/phalcon/cphalcon/blob/3.4.x/CHANGELOG-3.4.md)

sergeyklay (dismissed):
@mervick LGTM. Could you add tests covers your changes?

mervick:
@sergeyklay, sorry but I can't get it work on travis-ci, changing locales works wrong for number formatting.
Tests works fine on my localhost, but it does not works on travis, somehow it works for converting to string using `sprintf` but does not work when convert to string.

Have no idea what is wrong,
example:
```php
$num = 123.12;

setlocale(LC_ALL, 'en_US.utf8');
$str1 = sprintf('%s', $num);
$str2 = $num . '';

setlocale(LC_ALL, 'fr_FR.utf8');
$str3 = sprintf('%s', $num); // works as expected both travis and on localhost
$str4 = $num . ''; // does not works on travis-ci

print_r([$str1, $str2, $str3, $str4]);

// output on localhost
Array
(
    [0] => 123.12
    [1] => 123.12
    [2] => 123,12
    [3] => 123,12
)

// output on travis-ci
Array
(
    [0] => 123.12
    [1] => 123.12
    [2] => 123,12
    [3] => 123.12 // wrong
)

```

Code of test here:

https://github.com/mervick/cphalcon/blob/3.4.x-test/tests/unit/Validation/Validator/NumericalityTest.php#L90-L117

phalcon/validation/validator/numericality.zep
virgofx: I wonder if we should just switch to internally use PHP's `is_numeric(value)` only. That way we can avoid any bias that we've implemented. Instead we should be more of a wrapper around existing and well thought out methods that exist in the core.

Thus, I'm actually in favor of just replacing with: `if !is_numeric(value) {`

virgofx: It appears that `is_numeric()` covers everything in the regular expression and more. So yeah .. we should just straight proxy ... I think that's the best solution.

mervick:
Added tests, and it perfectly works on localhost.
But it does not works on travis-ci.

Travis ignore `LC_NUMERIC` locale settings.

The same test on localhost:
selection_141


Any help are welcome

sergeyklay:
@mervick 
> But it does not works on travis-ci.

That's why even such a small change needs to be tested.  Now let's imagine that this isn't Travis, but the real system in the production environment. Also note that the tests pass successfully for PHP 7.2

mervick:
@sergeyklay travis-ci have a lot of issues with locales, if someone knows how to change locale on travis-ci please advise.
This change on phalcon does not have an issue, test have an issue because expected locale does not have a valid format.

mervick:
Fixed issue with test.

I did some research and found what numeric formatting with comma in PHP is very old issue, and I think `Numericality` validator must not accept strings with comma, but must accepts float numbers. The current issue was what `Numericality` converts float numbers to string with comma when it checks regexp, but with additional `is_numeric` we can check valid float numbers and ignore strings with comma which is not float numbers.

sergeyklay (approved):

virgofx:
I don't have an issue with the changes ... as it does what's intended ...

virgofx (approved):
I don't have an issue with the changes ... as it does what's intended. Again, my personal preference was to just wrap the native function but as mentioned above it seems that misses a few things ... so this seems reasonable.

PR #13480, opened 2018-09-03
Fixes #11605: Parameters are stored in _bindParams when calling query execute

### LINKED ISSUE #11605: [2.1.x][Bug] PHQL subquery cant bind params for it- only for "array" type
Hi there,
I think i found a issue with sub-queries. Let get look that example: 

``` php
  $prepare = [
    'categories' => [1, 2, 3, 4, 5, 6, 10, 11]
  ];

  $phql = "SELECT * FROM Models\Teachers WHERE id IN (
    SELECT teacher_id FROM Models\CategoriesTeachers WHERE category_id IN ({categories:array})
  )";

  $query = $this->modelsManager->createQuery($phql);
  $elements = $query->execute($prepare);
```

So... Phalcon returns an error for me: "**Bind value is required for array type placeholder: categories**" - but it is prepared. I try this in older versions of phalcon- still make an error. If i am doing something wrong, please, correct me.
Just one more detail- if i use not array as param, but string or int `category_id = {category:int}` there is no problem- issue is only with array type.

Thank you!

CameronHall:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/11605

Small description of change: The parameters aren't stored in `this->_bindParams` when calling the `execute` function. This raises a problem when using array placeholders as there is a check [here](https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/model/query.zep#L671) to see if the binded parameter is in _bindParams. Because `execute` function doesn't store the `mergedParams` in `this->_bindParams`, it creates the problem above.

Thanks


niden:
@CameronHall Small request: Could you please add a line to the `CHANGELOG` for 3.4 so that we do not lose track of this?

Thank you!

CameronHall:
@niden I have updated `CHANGELOG-3.4.md` :)

niden:
Hmm a couple of tests are complaining, related to this PR. Could you have a look please? I need to go to bed.

CameronHall:
All fixed. I've squashed all three commits as well.

niden (approved):

niden:
@CameronHall Thank you very much for the contribution!!

PR #13502, opened 2018-09-22
Update route.zep

legale:
Type: documentation

Small description of change:
example how to deep parse URL with convert() function


legale:
unsuccessful attempt.

PR #13511, opened 2018-09-30
Add PHP 7.3 support

sergeyklay:
Hello!

* Type: ci improvements
* Link to issue: https://github.com/travis-ci/travis-ci/issues/9717

Small description of change:

- Add PHP 7.3 support to the build matrix on Travis CI
- Removed PHP 5.5 from build matrix (no longer supported by Travis CI)
- Do not run legacy tests (due to fantastic complexity associated with the satisfaction of dependencies from a past era)
- Reorganize and refactor some unit tests
- Refactor Travis CI build/testing stages

Thanks


lubber-de:
As this is merged into the 3.4.x branch already: Any idea when you will be able to provide windows-dlls for Phalcon 3.4.x and PHP 7.3.x. We cannot upgrade to Php 7.3 until then 😢

sergeyklay:
@lubber-de This PR about just an ability to build/test Phalcon using PHP 7.3 on CI but not about production ready for PHP 7.3. All necessary changes was made in the 4.0.x branch. The 3.x version no longer active developed.

lubber-de:
@sergeyklay Thanks for clarification. But as phalcon 4 won't be stable soon and more important won't be backward compatible to 3.4, I highly suggest that you at least think about making 3.4.2 php 7.3 compatible without any other fixes or features (that's what phalcon 4 is for) . Otherwise many teams will be either stuck on php 7.2 or have to migrate their existing and working projects to the phalcon 4 which is probably a lot of effort also because much stuff has been removed, so it's not only renaming some method calls. The main point is that php 7.3 is stable and released while phalcon 4 is not. So currently there is no combination of both libraries in stable condition. You should support the current php version for the 3.4.x branch at least until phalcon 4 is stable IMHO. 
Many people might consider using/migrating to Phalcon 4 only when it gets stable and probably waiting for some more first Bugfix releases. Until this happens my personal guess is, there will already be  php 7.3.10+ around.
My boss would tell me to drop phalcon then (a new project to migrate to phalcon 4 someday had to be made anyway, so my boss would argue to switch back to a raw php framework if that's the only way to be able to use the latest php version. 

Don't get me wrong, I really like what you are doing and you are doing it great! Just think about it again 😉

Anyway, I read somewhere somebody got a working 3.4.2 windows DLL for php 7.3. So, even if phalcon is not officially supporting it, I would love if you could point me to a download link. Then my concerns might be completely gone and lolling forward to a phalcon 4 upgrade can be planned in. More relaxed way 
. I currently had no luck by simply googling it.. 🙂

sergeyklay:
Let's continue in the #13847

PR #13516, opened 2018-10-08
Fixes #13501: getJsonRawBody() handles empty body and JSON decode errors

### LINKED ISSUE #13501: [BUG] getJsonRawBody() never returns false
The condition in the `getJsonRawBody()` method looks like this:
```php
// (...)
let rawBody = this->getRawBody();
if typeof rawBody != "string" {
	return false;
}
// (...)
```
`getRawBody()` method looks like this:
```php
// (...)
let contents = file_get_contents("php://input");
// (...)
```

`file_get_contents("php://input")` always return string, so the condition `if typeof rawBody != "string"` is never satisfied.

I think the `getJsonRawBody()` method should look like this:
```php
public function getJsonRawBody(boolean associative = false) -> <\stdClass> | array | boolean
{
	var rawBody, data;

	let rawBody = this->getRawBody();
	if rawBody === "" {
		return false;
	}

	let data = json_decode(rawBody, associative);

	if json_last_error() !== JSON_ERROR_NONE {
		return false;
	}

	return data;
}
```

### Details
* Phalcon version: master/0a24d8de4c3f52eee2558b344f7b49077cc79a2f

CameronHall:
Hello!

* Type: bug fix
* Link to issue: [#13501](https://github.com/phalcon/cphalcon/issues/13501)

Small description of change: As suggested by @michalzielanski we check if the request is empty instead of not a string type, to decide whether we continue or return false. JSON decode errors are also handled politely now as well.

Thanks


niden (approved):

niden:
Thank you!!

PR #13563, opened 2018-11-01
PR-11608 Modify redis backend

niden:
Hello!

* Type: enhancement
* Link to issue: https://github.com/phalcon/cphalcon/pull/11608 https://github.com/phalcon/cphalcon/issues/10905 https://github.com/phalcon/cphalcon/pull/11452

Made the `specialKey` (`_PHCR`) optional for the `Phalcon\Cache\Backend\Redis` adapter. This is based on the work of [@pantaovay](https://github.com/pantaovay)

Thanks


tests/unit/Cache/Backend/RedisCest.php
```diff
@@ -44,11 +44,15 @@ public function increment(UnitTester $I)
...
-            'port'  => env('TEST_RS_PORT', 6379),
-            'index' => env('TEST_RS_DB', 0),
-        ]);
+        $cache = new Redis(
+            new Data(['lifetime' => 20]),
+            [
+                'host'     => env('TEST_RS_HOST', '127.0.0.1'),
+                'port'     => env('TEST_RS_PORT', 6379),
+                'index'    => env('TEST_RS_DB', 0),
+                'statsKey' => '_PHCR',
```
sergeyklay: Well, all these changes will now require the user to pass `statsKey` key more often, but previously not.

I think from the point of view of the Phalcon user this changelog entry "Made the `specialKey` (`_PHCR`) optional" is wrong because now he is forced to play with this key almost always (or at least more often)

phalcon/cache/backend/redis.zep
```diff
@@ -188,18 +188,19 @@ class Redis extends Backend
...
 	public function save(keyName = null, content = null, lifetime = null, boolean stopBuffer = true) -> boolean
 	{
-		var nonePrefixedKey, lastKey, frontend, redis, cachedContent, preparedContent,
+		var prefixedKey, frontend, redis, lastKey, cachedContent, preparedContent,
 			tmp, ttl, success, options, specialKey, isBuffering;
 
 		if keyName === null {
-			let lastKey = this->_lastKey;
-			let nonePrefixedKey = substr(lastKey, 5 + strlen(this->_prefix));
+			let prefixedKey = substr(lastKey, 5),
```
sergeyklay: Now the number 5 is more than meaningless. The stats key may be `_PHCR` (5) or just `_` (1) right?

tests/unit/Cache/Backend/RedisCest.php
```diff
@@ -413,7 +492,7 @@ public function queryKeysWithStatsKeyAndPrefix(UnitTester $I)
         $keys = $cache->queryKeys();
         sort($keys);
 
-        $I->assertEquals(['a', 'b'], $keys);
+        $I->assertEquals(['phalcon-a', 'phalcon-b'], $keys);
```
sergeyklay: Is the business logic was changed? Previously we'd expect here result w/o prefix. Should we notice about this in the blogpost to alert users?

phalcon/cache/backend/redis.zep
sergeyklay: On other hand there is still at least 2 places where we require exact `_PHCR`

sergeyklay (changes requested):

tests/unit/Cache/Backend/RedisCest.php
niden: I disagree. The reason all these `statsKey` elements appear in the tests is because we assumed that the key would always be there. So all the existing tests must have that in the initialization of the `Redis` component. 

For an real life application, this initialization would happen mostly in the DI container and there will be no more changes.

tests/unit/Cache/Backend/RedisCest.php
niden: No the business logic remains the same. The test was incorrect. The logic is:

```
$cache->get('a'); // 'a' (no statsKey, no prefix)
$cache->get('a'); // 'phalcon_a' (no statsKey, prefix "phalcon_")
$cache->get('a'); // '_PHCRphalcon_a' (statsKey "_PHCR", prefix "phalcon_")
```

phalcon/cache/backend/redis.zep
niden: Made changes to take into account the length of the `statsKey`

sergeyklay:
@niden Could you please rebase

sergeyklay (approved):
LGTM. Relevant documentation changes are necessary.

PR #13590, opened 2018-11-11
4.0.x Remove name property from Phalcon\Di\Service so that it can be used as a better closure wrapper

dschissler:
The current design of the Service class doesn't make it well suited for acting as a closure wrapper and the name property is unnecessary.  There is only one place where the name property is used and that is in the DI and it already has access to the name of the service due to having just having received the name via a method parameter.  I added a single exception class that is to be thrown when this one situation occurs.  I think that this is a good use of an exception since it is an exceptional event.  Otherwise the design of Phalcon suffers only to satisfy this one moment.

Accepting this pull request will allow frameworks to use one service per file designs.

```php
return function() {
    // This is a service closure
}
```

In addition to the following two options;

```php
    return new Service(function() {

    });
```

and

```php
    return new Service(function() {

    }, true);
```

and then later by an additional pull request:

```php
    return new SharedService(function() {

    });
```

If this pull request is not accepted then it must be done like:

```php
    return new Service("this_is_bullshit", function() {

    });
```

and then hoping to not create a collision.  However for purposes of sanity this entire use of `Service` is essentially unavailable and for no good reason

This pull request is derived from an old pull request that I pulled down from before when Phalcon was in its neglected state.  I relied on the actual PR to run the tests for me and if I missed something here or something changed then I'll fix it and then if the fix or fixes become too messy then I will resubmit with a local copy-paste of the completed passing code.  Obviously that isn't an ideal development process and is already being addressed elsewhere by the Team.

### DIFF
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -34,6 +34,7 @@
 - Refactored `Phalcon\Db\Adapter\Pdo::query` to use PDO's prepare and execute. `Phalcon\Db\Adapter::fetchAll` to use PDO's fetchAll
 - Fixed  `\Phalcon\Http\Response::setFileToSend` filename last much _ 
 - Changed `Phalcon\Tag::getTitle()`. It returns only the text. It accepts `prepend`, `append` booleans to prepend or append the relevant text to the title. [#13547](https://github.com/phalcon/cphalcon/issues/13547) 
+- Changed `Phalcon\Di\Service` constructor to no longer takes the name of the service.
 
 ## Removed
 - PHP < 7.0 no longer supported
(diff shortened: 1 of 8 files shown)


niden:
The code looks solid. 

We will have to rework this component for PSR implementation later on.

niden (dismissed):

niden:
@dschissler Can you add a line to the `CHANGELOG` please before I merge this?

dschissler:
Damn typo in the changelog but its just an extra "s".

niden (approved):

niden:
@dschissler Thank you!

PR #13589, opened 2018-11-10
fix all interface definitions

Izopi4a:
Hello!

* Type: bug fix | code quality 

**In raising this pull request, I confirm the following (please check boxes):**

- [x] I have read and understood the [Contributing Guidelines](https://github.com/phalcon/cphalcon/blob/master/CONTRIBUTING.md)?
- [x] I have checked that another pull request for this purpose does not exist.
- [ ] I wrote some tests for this PR.

Small description of change:

this fixes all the fatal errors of bad function parameters and type hints after zephir 0.11 fixes type hints
Even tho its said that it should be done in 4.0.x after digging around I don't think there will be broken things.

there are a lot of wrong interfaces, i do think we should merge this asap ( if things don't get broken )

Session AdapterInterface was really strange, but i changed in the way to keep backwards compatibility. 

What is left to be decided

- [x] Session destroy interface is garbage but it works at the moment without breaking anything
- [ ] phalcon/db/adapterinterface.zep ( limit changes ) should we change mysql adapter to `var sqlQuery` 
https://github.com/phalcon/cphalcon/pull/13589/files#diff-34b0990a92ee5f99eb0c587c4e0c4b3b

- [ ] mysql postgres adapter can return array | string when it actually returns only strings https://github.com/phalcon/cphalcon/blob/9227ff106254777b2046878fce1ea3af6ba0efe7/phalcon/db/dialect/postgresql.zep#L341

- [ ]  `offsetUnset` `offsetGet` `offsetSet` `offsetExists` `unserialize` `Seekable::seek` are complaining because of the way internal php interfaces are defined ( i think ) or at least the case for me  
- [x] `hasFiles` in http/request will be renamed in the `near feature`
- [x] Di::attempt needs to be refactored `some day` to always return service

```
$ php -v
PHP 7.2.10-0ubuntu0.18.04.1 (cli) (built: Sep 13 2018 13:45:02) ( NTS )
Copyright (c) 1997-2018 The PHP Group
```

links

https://github.com/phalcon/cphalcon/issues/13444
https://github.com/phalcon/cphalcon/issues/13465
https://github.com/phalcon/zephir/issues/1656

Thanks

PR #13598, opened 2018-11-13
continueing fixing interfaces

Izopi4a:
Hello!

Small description of change:

continue fixing interfaces for building with zephir 0.11

Thanks

### DIFF
--- a/phalcon/acl/adapter.zep
+++ b/phalcon/acl/adapter.zep
@@ -49,19 +49,19 @@ abstract class Adapter implements AdapterInterface, EventsAwareInterface
 
 	/**
 	 * Role which the list is checking if it's allowed to certain resource/access
-	 * @var mixed
+	 * @var string
 	 */
 	protected _activeRole { get };
 
 	/**
 	 * Resource which the list is checking if some role can access it
-	 * @var mixed
+	 * @var string
 	 */
 	protected _activeResource { get };
 
 	/**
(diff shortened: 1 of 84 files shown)


Izopi4a:
merging this and/or previous PR changes nothing. It only allows people manually compiling phalcon via latest zephir.

i think people soon will upgrade their zephir version to the latest and than they will be ... "It doesn't work.." 

at least what I think. I am 99% sure if released with zephir 0.10 no one will notice the difference, its only for the other people who might try to fix/add something, try to compile .. and nothing. 

If you decide not to merge this one we should revert the previous one as well, and leave those changes for 4.0.x

sergeyklay:
I propose keep this open for a few days to ato take a fresh look at it later. But at the moment I agree with @niden.

sergeyklay:
@Izopi4a Since we are a small team, we have to be careful on where to focus in order to keep the framework moving forward. Because of that, we cannot introduce code in a branch that is/will no longer be supported and with that, introduce errors that will divert us from the work we do for `4.0.x`. Phalcon `3.4.x` has only one release left and then we will no longer release anything for it, unless it is security related. Also, Phalcon `3.4.x` will never be compiled using Zephir `0.11.x` and as a result these changes will not matter in this branch but are essential to the `4.x` branch. 

I also reverted this PR: https://github.com/phalcon/cphalcon/pull/13589. Thank you for contributing, and for helping us make Phalcon better!

Issue #12705, opened 2017-03-12
View partials do not properly maintain scope


jesseforrest:
### Expected and Actual Behavior

Typically in MVC frameworks, each partial view gets variables in an encapsulated scope.  Therefore if a variable is changed it will not affect any other view partial that used that same variable name.  I'm noticing different behavior with Phalcon.  There may be performance reasons for doing it this way, but unfortunately larger teams can not work this way since they are dirtying other view partials.  I think there should be an option (if one doesn't already exist) to make sure each partial gets its own scope of variables.

For example, let's say we have the following controller:

```php
class ExampleController extends Phalcon\Mvc\Controller
{
   public function exampleAction()
   {
      $this->view->partial('example/index', ['count' => 1]);
   }
}
```

Now lets say we have the follow `example/index.phtml` view:

```php
<?= $count; ?>
<?php $this->view->partial('example/subIndex', ['items' => ['x', 'y', 'z']]); ?>
<?= $count; ?>
```

Let's then say this is the contents of `example/subIndex.phtml`:

```php
<?php
$count = count($items);
echo $count;
```

I would expect that `subIndex` would have it's own scope, so that `$count` does not override the parent `$count` variable.  

The desired output should be:
```
131
```

The actual output is:
```
133
```

### Details

* Phalcon version: 3.0.3
* PHP Version: 7.0.16
* Operating System: Linux
* Server: Apache

Jurigag:
I think there were actually created the same issue here not long ago.

Jurigag:
Oh there it is - https://github.com/phalcon/cphalcon/issues/12648 similar issue

sergeyklay:
I hope we managed to sort out with this issue finally here https://github.com/phalcon/cphalcon/pull/13606. All necessary changes are already in the appropriate branch. We will try to release the next version in the coming days.  Thank you for your patience, the report, and for helping us make Phalcon better.

Issue #12385, opened 2016-11-01
the variables in partials globalize after upgrade to 3.0.1 (volt)


odat:
Hello,
the variables in partials globalize, how i can set some of variable on some partials and don't defined on another one on the same view 

for example in controller i pass id on view by 
`$this->view->id = '1' ; `

in the view template :   

```twig
{{ id }}   {# now the id is 1 #} 

{{ partial("some/partial",['id':'2','another':3]) }}

{{ id }}   {# now the id is 2 #} 

{{ another}}   {# and i can use another now, it's defined!! #} 
```

when i defined the some partial with another id , the id is overwritten on the old one on the page

Thanks,

Jurigag:
You have php 7 ? I think this is why it's happening, not because of 3.0.1

odat:
Yes php 7, but can i get some way to fix this issue

Jurigag:
Then do PR if you have any solution or it's just workaround ?

seliver:
Has this issue been fixed? Workaround maybe?

seliver:
This is due ro zephir incompatibilities with php7: https://forum.phalconphp.com/discussion/14497/view-variables-are-surprising-bug-or-feature
@phalcon Is there a workaround for this?

Jurigag:
I don't think there is right now @andresgutierrez any plans to fix this?

alivex:
I can confirm the bug occurring on php7 alone. Especially, when I can getRender(). I'll post my code, if it helps fixing the issue

```php
//User has firstName, lastName, email, and more...
$content = $this->view->getRender("mail", "newPass",
                            [
                                "user" => $user->firstName . ' ' . $user->lastName,                               
                                "base_url" => 'https://' . $_SERVER['SERVER_NAME'] . ':' . $_SERVER['SERVER_PORT']
                            ]
                        );
//After this part, User has only firstName and lastName.

```

emiliodeg:
I found a new super behavior
```volt
    {{ robot.getPart() }} //part object #....
    {{ partial("_partials/show-part", ['part':robot.getPart()]) }} //render fine
    {{ robot.getPart() }} //robot is null!!!!
```
pain solution

```volt
    {% set part = robot.getPart() %} //part object #....
    {{ partial("_partials/show-part", ['part':part]) }} //render fine
    {{ robot.getPart() }} //robot is fine
```
[...]

PR #13606, opened 2018-11-18
Full support of PHP 7.3

sergeyklay:
- Full support of PHP 7.3
- Fixes: #13288, #12176, #12385, #12648, #12705

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,6 @@ matrix:
   fast_finish: true
   allow_failures:
     - php: 'master'
-    - php: '7.3'
 
 cache:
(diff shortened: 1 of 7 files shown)


lubber-de:
Could you please provide the windows-dlls for PHP 7.3.x and Phalcon 3.4.2, or just release a 3.4.3 for that, if this is merged already? 
We cannot upgrade to PHP 7.3 before and Phalcon 4 wont be stable soon and wont be backward compatible, so we have to use phalcon 3.4.x for a while, but have to upgrade php to 7.3.x

PR #13634, opened 2018-12-06
T13438 psr 3 implementation 2

niden:
@sergeyklay 

Progress check and comments so far.

Note this is on a different branch which will be deleted when this work is completed.


niden:
New PR will be sent

Issue #13006, opened 2017-08-04
Column doesn't belong to any of the selected models (1)


mbrostami:
### Expected and Actual Behavior
Related to this issues: #1731 , #10934
When I want to use subquery in where conditions and having beside that I got this error: 
```php
Column 'SomethingNotInModels' doesn't belong to any of the selected models (1), when preparing: SELECT "aaa" AS [SomethingNotInModels] FROM [Trumpet\Models\Ad] AS [a] WHERE a.IDCategory IN (SELECT c.ID from Trumpet\Models\Category c) HAVING SomethingNotInModels = "aaa"

#0 [internal function]: Phalcon\Mvc\Model\Query->_getQualified(Array)
#1 [internal function]: Phalcon\Mvc\Model\Query->_getExpression(Array, true)
#2 [internal function]: Phalcon\Mvc\Model\Query->_getExpression(Array)
#3 [internal function]: Phalcon\Mvc\Model\Query->_prepareSelect()
#4 [internal function]: Phalcon\Mvc\Model\Query->parse()

```

> Reproduce the issue:
```php
$builder->columns('"aaa" as SomethingNotInModels')
            ->from(['a' => 'Trumpet\Models\Ad'])
            ->where('a.IDCategory IN (SELECT c.ID from Trumpet\Models\Category c)')
            ->having('SomethingNotInModels = "aaa"');
$builder->getQuery()->execute();
```
If I remove 'where' OR 'having' it doesn't trigger error.
### Details

* Phalcon version: 3.1.2
* PHP Version: 5.6.30-12~ubuntu14.04.1+deb.sury.org+1
* Operating System: Ubuntu
* Installation type: installing via package manager
* Server: Nginx
* Other related info (Database, table schema): 
```
create table Ad (ID INT(20), IDCategory INT(20));
create table Category (ID INT(20));
```

alexbusu:
I have this issue as well.
So, to reproduce you just have to 
1. select a column with alias
2. use a sub-query in `WHERE` clause
3. use the alias further in `HAVING` OR `ORDER BY` (which in SQL will apply to selected result set, of course)
In my case the error trace is 
```
#0 [internal function]: Phalcon\Mvc\Model\Query->_getQualified(Array)
#1 [internal function]: Phalcon\Mvc\Model\Query->_getExpression(Array)
#2 [internal function]: Phalcon\Mvc\Model\Query->_getOrderClause(Array)
[...]

Issue #12548, opened 2017-01-13
HAVING after nested SELECT doesn't work correctly in PHQL


nsossonko:
```sql
SELECT grandParent.id, SUM(parent.total) AS total
FROM Parents AS parent
INNER JOIN GrandParents AS grandParent
ON parent.grandParent = parent.id
WHERE NOT EXISTS (
  SELECT * FROM Childs AS child
  WHERE parent.id = child.parent
)
HAVING total > 1
```

That causes the PHQL parser to throw an exception that "total is not a column". If I switch to:
"HAVING SUM(parent.total) > 1" it works fine.

PHP 7 with Phalcon 3.0 (master).

Jurigag:
Im guessing PHQL is just following **SQL standard** which **doesn't allow** aliases in columns. Other engines like for example MySQL can allow it, but this is not part of sql standard.

nsossonko:
It allows using the alias when not using a subquery...

Jurigag:
Oh okay then.

Issue #1731, opened 2013-12-23
Uncaught exception: Column 'cate_id' doesn't belong to any of the selected models (1)


dancebear:
I use the following statement to cache metadata

```
$di->set ( 'modelsMetadata', function () use($config) {
    if (! $config->application->debug && isset ( $config->metadata )) {
        $metaDataConfig = $config->metadata;
        $metadataAdapter = '\Phalcon\Mvc\Model\Metadata\\' . $metaDataConfig->adapter;
        $metaData = new $metadataAdapter ( $config->metadata->toArray () );
    } else {
        $metaData = new \Phalcon\Mvc\Model\MetaData\Memory ();
    }
            $metaData->setStrategy ( new \Core\Model\AnnotationsMetaDataInitializer () );
    return $metaData;
}, true );
```

class AnnotationsMetaDataInitializer

```
<?php
namespace Core\Model;

use Phalcon\Mvc\ModelInterface,
    Phalcon\DiInterface,
    Phalcon\Mvc\Model\MetaData,
    Phalcon\Db\Column;

class AnnotationsMetaDataInitializer
{

    /**
     * 初始化模型的元数据
     *
     * @param \Phalcon\Mvc\ModelInterface $model
     * @param \Phalcon\DiInterface        $di
     *
     * @return array
     */
    public function getMetaData(ModelInterface $model, DiInterface $di)
    {
    $reflection = $di['annotations']->get($model);
    $properties = $reflection->getPropertiesAnnotations();

    $attributes = array();
    $nullables = array();
    $dataTypes = array();
    $dataTypesBind = array();
    $numericTypes = array();
    $primaryKeys = array();
    $nonPrimaryKeys = array();
    $identity = null;

    foreach ($properties as $name => $collection) {

        if ($collection->has('Column')) {

        $arguments = $collection->get('Column')->getArguments();

        /**
         * Get the column's name
         */
        if (isset($arguments['column'])) {
            $columnName = $arguments['column'];
        } else {
            $columnName = $name;
        }

        /**
         * 处理列注释的类型参数
         */
        if (isset($arguments['type'])) {
            switch ($arguments['type']) {
            case 'integer':
                $dataTypes[$columnName] = Column::TYPE_INTEGER;
                $dataTypesBind[$columnName] = Column::BIND_PARAM_INT;
                $numericTypes[$columnName] = true;
                break;
            case 'string':
                $dataTypes[$columnName] = Column::TYPE_VARCHAR;
[...]

PR #13637, opened 2018-12-07
Prevent overwrite _sqlColumnAliases by sub-queries, in Phalcon\Mvc\Model\Query

alexbusu:
Hello!

* Type: bug fix
* Link to issue: 
  * https://github.com/phalcon/cphalcon/issues/13006
  * https://github.com/phalcon/cphalcon/issues/12548
  * https://github.com/phalcon/cphalcon/issues/1731

Small description of change:

In order to prevent the SQL Column Aliases override in _root_ select, by the sub-queries, I have scoped the sub-queries aliases into another entries in `_sqlColumnAliases`, which is now and array of set-of-aliases entries. The `_sqlColumnAliases[0]` contains the aliases of the root query, the `_sqlColumnAliases[1]` contains the set of aliases of the _last_ parsed sub-query on level 1, and so on.

Thanks

### DIFF
--- a/phalcon/mvc/model/query.zep
+++ b/phalcon/mvc/model/query.zep
@@ -120,7 +120,7 @@ class Query implements QueryInterface, InjectionAwareInterface
 
 	protected _sqlAliasesModelsInstances;
 
-	protected _sqlColumnAliases;
+	protected _sqlColumnAliases = [];
 
 	protected _modelsInstances;
 
@@ -137,6 +137,8 @@ class Query implements QueryInterface, InjectionAwareInterface
 	protected _enableImplicitJoins;
 
 	protected _sharedLock;
+
+	protected _nestingLevel = -1;
 
 	/**
 	 * TransactionInterface so that the query can wrap a transaction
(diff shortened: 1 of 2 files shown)


phalcon/mvc/model/query.zep
```diff
@@ -178,6 +180,9 @@ class Query implements QueryInterface, InjectionAwareInterface
 		} else {
 			let this->_enableImplicitJoins = globals_get("orm.enable_implicit_joins");
 		}
+
+		let this->_nestingLevel = -1;
```
CameronHall: How about setting `-1` as the default value when you define the property on the class?

phalcon/mvc/model/query.zep
```diff
@@ -178,6 +180,9 @@ class Query implements QueryInterface, InjectionAwareInterface
...
+		let this->_sqlColumnAliases = [];
```
CameronHall: Same here, set `[]` as the default property on the class.

CameronHall (commented):
Hey @alexbusu,

Love your work! I've suggested two minor changes if you wanted to take a quick look :)

phalcon/mvc/model/query.zep
alexbusu: Absolutely!

phalcon/mvc/model/query.zep
alexbusu: Good point :)

CameronHall (approved):

CameronHall (changes requested):
Can you please rebase and target 4.0.x and squash your commits? 3.4.2 is the final revision of 3.4.x unless there's a critical bug or security vulnerability.

And don't forget to update the changelog.

CameronHall (dismissed):

niden (approved):

PR #13659, opened 2018-12-12
4.0.x Fixing interefaces

niden:
Hello!

* Type: bug fix
* Link to issue:

Thanks


virgofx:
While you're at it can the Request/Response interfaces be updated? I'm sure there are more issues but one of our static type check blockers is that `getURI()` is not in `RequestInterface`.

niden:
@virgofx If you don't mind can you add an issue or if there is one already assign it to me and cc me?

I also have to fix the session interfaces ...

PR #13671, opened 2018-12-18
Update the nanobox config to mongo 4.0

npfedwards:
Hello!

* Type: code quality

Updating the included nanobox config to use the latest 4.0.x mongodb version (using the offical docker image) under the assumption that we should be testing in environments that include the latest mongo version rather than an out of date version.

Thanks

### DIFF
--- a/tests/_ci/nanobox/boxfile.7.2.yml
+++ b/tests/_ci/nanobox/boxfile.7.2.yml
@@ -89,7 +89,7 @@ data.memcached:
   image: nanobox/memcached:1.4
 
 data.mongodb:
-  image: nanobox/mongodb:3.0
+  image: mongo:4.0
 
 data.mysql:
   image: nanobox/mysql:5.7
--- a/tests/_ci/nanobox/boxfile.7.3.yml
+++ b/tests/_ci/nanobox/boxfile.7.3.yml
@@ -89,7 +89,7 @@ data.memcached:
   image: nanobox/memcached:1.4
 
 data.mongodb:
-  image: nanobox/mongodb:3.0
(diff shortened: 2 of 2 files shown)


sergeyklay (approved):

niden:
@npfedwards Thank you!!

danhunsaker:
This may have some issues in actual use, especially if users decide to use the testing configs to deploy their apps, since the official Mongo image doesn't include the Nanobox hooks for configuration, environment, migration (used for scaling), update, backup/restore, or clean startup/shutdown. This should work fine for testing, but that's about it.

npfedwards:
@danhunsaker Is adding a new version to nanobox's images literally changing the numbers in the dockerfile? If so I'll submit a pull request and we can use that. I ask because you seem to be the expert and have contributed to that repo.

danhunsaker:
I used to work for Nanobox, so I had a lot more of an effect on how well things got done back then. That's no longer the case.

It shouldn't cause any issues in the short term. It's just a caveat for now. We'll want to make it clear that the testing boxfiles aren't meant for production use, and that some adjustments may be needed to get them working in production.

niden:
@danhunsaker I know about the hooks (we talked about this before) but I think we are OK in this aspect.

This "version" if you like of nanobox is designed to run the testing suite only locally - no deployment anywhere, so adding the container will help the suite.

Now if someone wants to deploy this setup to a provider using nanobox.io then they have to use the nanobox supplied images.

And yes we need to add a warning that these boxfiles are only for local use not prod.

sergeyklay:
+1 for warning/notice

Issue #12295, opened 2016-10-06
Container interface (PSR-11)


sergeyklay:
Phalcon's DI must be fully compatible with [Container interface (PSR-11)](https://www.php-fig.org/psr/psr-11/) for v4.0.0

---

Refs:
- https://github.com/phalcon/cphalcon/pull/13437

virgofx:
How would this be implemented? Create similar container interface or would Phalcon somehow import/reference Psr* packages?

sergeyklay:
I need research best way

Jurigag:
@virgofx the same question is for psr request i guess

SidRoberts:
I personally find this PSR a bit vague (it only has two methods) so I don't think it's absolutely essential that Phalcon should actually implement the interface. I think just following the spec should be enough.

Two things I'd like to recommend:

* Be able to specify dependencies/services in classes instead of anonymous functions (see https://github.com/SidRoberts/container for a working example).

* Inject things via the constructor instead of the accessing the DI/container inside the code (see https://github.com/phalcon/cphalcon/blob/master/phalcon/tag.zep#L205). I think all Phalcon components be completely independent of the DI/container **although this would involve changing how many classes operate**.

sergeyklay:
@dragoonis Just assigned to you.

Idrinth:
would likely be easiest for users if phalcon would just contain the psr-interfaces - having a similar one is not really going to help out much, since their implementations and the interfaces themselves would not be easily exchangeable.

dschissler:
@Idrinth Sergei was doing a lot of work on the PSR extension and I'm sure that this was the idea.

niden:
@Idrinth That is the package we are using for the interfaces (v4)

niden:
Implemented

PR #13681, opened 2018-12-22
T12295 psr 11

niden:
Hello!

* Type: new feature
* Link to issue: #12295 

Created a new "proxy" class `Phalcon\Container`. This class implements PSR-11 and can be used instead of the `DI` container. Note that you still have to create the DI container and set all services into it and then inject that container in this container (I got dizzy writing this).

This is not the full implementation, a "proxy" for now - more on that in a future version. 

Thanks


niden:
cc @sergeyklay

tests/unit/Container/ConstructCest.php
```diff
@@ -0,0 +1,42 @@
...
+     * @since  2018-11-13
+     */
+    public function containerConstruct(UnitTester $I)
+    {
+        $I->wantToTest('Container - __construct()');
+
+        $container = new Di();
+        $class     = ContainerInterface::class;
+        $actual    = new Container($container);
+        $I->assertInstanceOf($class, $actual);
```
sergeyklay: Can this be simplified, eg
```php
$I->assertInstanceOf(ContainerInterface::class, new Container(new Di));
```

phalcon/container.zep
```diff
@@ -0,0 +1,46 @@
...
+ * For the full copyright and license information, please view the LICENSE.txt
+ * file that was distributed with this source code.
+ */
+
+namespace Phalcon;
+
+use Psr\Container\ContainerInterface;
+use Phalcon\DiInterface;
+
+final class Container implements ContainerInterface
```
sergeyklay: Do we really need this as final

sergeyklay (dismissed):
LGTM

tests/unit/Container/ConstructCest.php
niden: I like to keep things separate in the test; It makes reading a bit easier and discourages people writing an assertion that is long and difficult to understand.

phalcon/container.zep
niden: No, Let me correct that

PR #13693, opened 2018-12-24
Fixes #13464: httpOnly is no longer initialised with a value

### LINKED ISSUE #13464: [BUG] httpOnly flag is ignored in set() for cookies
System:
```
$ phalcon info

Phalcon DevTools (3.4.0)

Environment:
  OS: Linux vmv 4.9.0-6-amd64 #1 SMP Debian 4.9.88-1+deb9u1 (2018-05-07) x86_64
  PHP Version: 7.0.30-0+deb9u1
  PHP SAPI: cli
  PHP Bin: /usr/bin/php7.0
  PHP Extension Dir: /usr/lib/php/20151012
  PHP Bin Dir: /usr/bin
  Loaded PHP config: /etc/php/7.0/cli/php.ini
Versions:
  Phalcon DevTools Version: 3.4.0
  Phalcon Version: 3.4.0
  AdminLTE Version: 2.3.6
```

By default `httpOnly` seems to be active ([see](https://github.com/phalcon/cphalcon/blob/master/phalcon/http/cookie.zep#L58)). In [API documentation](https://docs.phalconphp.com/de/latest/api/Phalcon_Http_Response_Cookies) one can see, that with the last parameter of method `set` this flag can be set.
But, regardless of value I'm using, the cookie is sent with active `httpOnly` to browser, always.

Ex. in controller:
```
$this->cookies->set(
    'uid', // name
    $id, // value
    time() + 86400, // expire
    '/', // path
    false, // secure
    $domain, // domain
    false // http only
);
```

The only way I get `httpOnly` removed is by calling following statements right after `set()`:

```
$cookie = $this->cookies->get('uid');
$cookie->setHttpOnly(false);
```

So the flag in `set()` seems to be ignored (all other parameters are working correctly, only `httpOnly` is set to `true` always). That's not very comfortable. This is urgent to me, so to be sure I fall back to native `setcookie()` now.
What I do not understand is, that the flag `httpOnly` is optional and declared with `null` when calling `set()`. But as property in class `Cookie` it's initialized with `true`, see [here](https://github.com/phalcon/cphalcon/blob/master/phalcon/http/cookie.zep#L60).

CameronHall:
Hello!

* Type: bug fix
* Link to issue: #13464

-  [x] I have read and understood the [Contributing Guidelines](https://github.com/phalcon/cphalcon/blob/master/CONTRIBUTING.md)?

Small description of change: httpOnly is no longer initialised with a value

Thanks

### DIFF
--- a/phalcon/http/cookie.zep
+++ b/phalcon/http/cookie.zep
@@ -48,7 +48,7 @@ class Cookie implements CookieInterface, InjectionAwareInterface
 
 	protected _secure;
 
-	protected _httpOnly;
+	protected _httpOnly = false;
 
 	/**
 	 * The cookie's sign key.
--- a/phalcon/http/cookie.zep
+++ b/phalcon/http/cookie.zep
@@ -48,7 +48,7 @@ class Cookie implements CookieInterface, InjectionAwareInterface
 
 	protected _secure;
 
-	protected _httpOnly = true;
+	protected _httpOnly;
 
 	/**
 	 * The cookie's sign key.
--- a/tests/unit/Http/Cookie/CookieCest.php
+++ b/tests/unit/Http/Cookie/CookieCest.php
@@ -153,6 +153,7 @@ class CookieCest extends HttpBase
     public function testIssue13464(UnitTester $I)
     {
         $I->wantToTest("Issue #13464");
+        $I->checkExtensionIsLoaded('xdebug');
         
         $this->setDiCrypt();
         $container = $this->getDi();
--- a/tests/unit/Http/Cookie/CookieCest.php
+++ b/tests/unit/Http/Cookie/CookieCest.php
@@ -9,7 +9,7 @@
  * file that was distributed with this source code.
  */
 
-namespace Phalcon\Test\Unit\Http;
+namespace Phalcon\Test\Unit\Http\Cookie;
 
 use Phalcon\Http\Cookie;
 use Phalcon\Http\Cookie\Exception;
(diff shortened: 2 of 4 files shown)


tests/unit/Http/Cookie/CookieCest.php
```diff
@@ -149,4 +149,22 @@ public function shouldWorkWithoutInitializeInternalCookiesProperty(UnitTester $I
         $actual  = $cookies->send();
         $I->assertTrue($actual);
     }
+
+    public function testIssue13464(UnitTester $I)
+    {
+        $I->wantToTest("Issue #13464");
```
sergeyklay:
```suggestion
        $I->wantToTest("Issue #13464");
        $I->checkExtensionIsLoaded('xdebug');
```

sergeyklay (changes requested):

sergeyklay (dismissed):

sergeyklay (dismissed):

phalcon/http/cookie.zep
```diff
@@ -48,7 +48,7 @@ class Cookie implements CookieInterface, InjectionAwareInterface
 
 	protected _secure;
 
-	protected _httpOnly = true;
+	protected _httpOnly;
```
niden: Can we keep this to be initialized with false?

niden (changes requested):

phalcon/http/cookie.zep
sergeyklay: Yeah. Otherwise it is just NULL.

CameronHall: I did this when I updated the tests :| I'm unsure where it went, I'll sort it out shortly.

CameronHall:
@niden I made the changes :)

niden (approved):

niden:
Thank you @CameronHall

Issue #13007, opened 2017-08-04
Missing translation in NativeArray should (optionaly) raise a warning


dugwood:
### Expected and Actual Behavior

When using NativeArray (https://github.com/phalcon/cphalcon/blob/master/phalcon/translate/adapter/nativearray.zep) on a missing key, the key itself is returned if there's no translation for this key.

A warning should be raised to deal with such a situation, avoiding displaying keys on the production website. It's not a security issue, but that should help displaying what's intended.

### Details

* Phalcon version: since 2.x
* PHP Version: not relevant
* Operating System: not relevant
* Installation type: compilation (Linux)
* Zephir version (if any): not relevant
* Server: not relevant

I can create a pull request for this, just tell me if you have a specific implementation to follow: throw an exception (I see that you have `Phalcon\Translate\Exception`), passing a function...
And option would be `exceptionOnMissingKey` or anything you'd prefer, defaulting to false of course.

dugwood:
An exception may be too difficult to deal with, it should be a function:
`new NativeArray(['content' => [], 'onMissingKey' => $myfunction]`

The function may return the key, overriding the code:
https://github.com/phalcon/cphalcon/blob/master/phalcon/translate/adapter/nativearray.zep#L63

Example of function to pass:
```php
function missingKey($key) {
    trigger_error('Can\'t find translation key: '.$key);
    return $key; // or better, return nothing with «return '';»
}
```

niden:
This has been addressed

PR #13730, opened 2019-01-04
T13007 nativearray missing

niden:
Hello!

* Type: new feature
* Link to issue: #13007 

- Added a new parameter in the constructor array `triggerError`. If set, then the component will trigger an error `E_USER_ERROR` if the key is not found.
- Added  `notFound()` method which returns the key requested if not found. The method can be overriden when extending the class, returning what the developer needs
Thanks


niden:
cc @sergeyklay

sergeyklay (dismissed):

Issue #13771, opened 2019-01-19
[NFR] Service Locator


niden:
Introduce a service locator that can be extended and used to store services from the DI to any other component that requires it.

- Lazy loading
- Quick location of services
- Ground work for auto wiring

niden:
Addressed

PR #13772, opened 2019-01-19
T13771 service locator

niden:
Hello!

* Type: new feature
* Link to issue: #13771 

Introduced service locator; Added tag locator for `Phalcon\Html\`

Thanks


sergeyklay (changes requested):
-

phalcon/html/helper/abstracthelper.zep
```diff
@@ -0,0 +1,134 @@
...
+
+use Phalcon\Html\Exception;
+use Phalcon\EscaperInterface;
+
+/**
+ * Phalcon\Html\Helper\AbstractHelper
+ *
+ * Abstract class for all html helpers
+ */
+abstract class AbstractHelper
```
sergeyklay: This class should implement an interface with the __invoke method

phalcon/html/helper/abstracthelper.zep
```diff
@@ -0,0 +1,134 @@
...
+	public function __construct(<EscaperInterface> escaper)
+	{
+		let this->escaper = escaper;
+	}
+
+	/**
+	 * Keeps all the attributes sorted - same order all the tome
+	 *
+	 * @var array  attributes Any attribute overrides
+	 * @var array  attributes Any additional attributes
```
sergeyklay: We don need annotation here

phalcon/html/helper/abstracthelper.zep
```diff
@@ -0,0 +1,134 @@
...
+	 * @var array  attributes Attributes to render
+	 */
+	protected function renderAttributes(array attributes) -> string
+	{
+		var key, result, value;
+
+		let result = "";
+		for key, value in attributes {
+			if typeof key == "string" && value !== null {
+				if typeof value == "array" || typeof value == "resource" {
```
sergeyklay: how about object

sergeyklay (changes requested):

phalcon/html/helper/abstracthelper.zep
niden: We can't put `__invoke()` there because it is not the same `__invoke` for every helper. For instance one might get instantiated with 2 parameters another with 3. See `Anchor` (3) vs `Button` (2)

sergeyklay: Best practice is to use an interface to specify the _contract_ and an abstract class as just one _implementation_ thereof. That abstract class can fill in a lot of the boilerplate so you can create an implementation by just overriding what you need to or want to without forcing you to use a particular implementation.

phalcon/html/helper/abstracthelper.zep
niden: Doen

phalcon/html/helper/abstracthelper.zep
niden: How can you override the parameters of a method (number or type)? There is no interface that I can abstract from the helpers since they all call `__invoke` with different parameters based on the type. The abstract helper class is there to offer common functionality for the helpers.We can easily rename it to `Base` or something if Abstract is not ok.

sergeyklay: Well, if the design implies that the methods may have different signatures and this cannot be circumvented, then I cannot argue

sergeyklay (approved):

sergeyklay:
@niden Could you please update change log?

PR #13792, opened 2019-02-02
Camel-case the action name (before call in handler)

### LINKED ISSUE #13555: Route action and controller parameter inconsistencies
```php
$router->add('...', [
    'controller' => 'my_controller' // MyController
    'action' => 'my_action' // my_action
]);
```

alexbusu:
Hello!

* Type: bug fix
* Link to issue: this aims to fix #13555

Small description of change:

I have added `lcfirst(camelize())` for `_actionName` in `Phalcon\dispatcher::getActiveMethod()`.

Thanks


niden:
@alexbusu Can you please either cherry pick d6f8a6696192fe098a529d9cdc83b1c4ad6207f1 from my fork (niden/cphalcon) or open the `tests/_ci/.env.default` and remove the commas after the double quote (MySQL area). That is what makes the build fail.

Thank you so much for this work. Could we also add a test or two for this?

Thanks again

niden:
@alexbusu you can rebase your work from 4.0. The change I did to fix the tests is in there, so your tests will run just fine.

alexbusu:
@niden It seems that `camelize` transforms `thisCase` into `Thiscase`. Is this the appropriate behaviour? If yes, then I guess this function is not suitable in `Dispatcher::getActiveMethod()`. What do you think?

alexbusu:
@niden I have adjusted the code. Now the tests do pass.

sergeyklay (dismissed):

niden:
I haven't thought about that @virgofx . Thanks. @alexbusu Can you have a look if we can move that logic to the setter?

alexbusu:
Thanks for feedback @virgofx .
I thought about that and I checked where `Dispatcher::getActiveMethod()` is [called internally, within the Framework](https://github.com/phalcon/cphalcon/search?q=getActiveMethod&unscoped_q=getActiveMethod). There is only one call, in `Dispatcher::dispatch()` method. So per request the formatting will happen **only once**.
Following your concern, you are actually talking about the active handler, which is not the object of the current PR.
Anyway, formatting the method name in setter means **every route entry will have to format the action name**, despite the fact that only one route will match at most. Clearly there will be no performance gain at all.
I hope you do agree with me.
@niden , what do you think?

virgofx:
Well the method is public which means the UI logic can call it. Depending on how your controllers are setup - views/partials can reference an active state by checking the handler ($dispatcher->getActiveHandler()) or the method (e.g. a controller that has multiple actions -- /:controller/:action) using $dispatcher->getActionName().

So I do think it's still applicable. I think we're only ever using the setter once via the main Application ... and it's less likely to recall `setActionName()` via events/UI/partials; whereas it's much more frequent on the getter side.

So yeah ... couldn't we just move the logic to the setter?

alexbusu:
@virgofx I got your point, will have a look. Thanks

alexbusu:
@virgofx Yes, you're right!
This can be actually improved. The confusion I made was between the dispatcher and the router components for [setActionName](https://github.com/phalcon/cphalcon/blob/master/phalcon/mvc/application.zep#L282) (which initially I thought you are confusing :man_facepalming:).
So yeah, good point.

phalcon/dispatcher.zep
```diff
@@ -198,7 +198,7 @@ abstract class Dispatcher implements DispatcherInterface, InjectionAwareInterfac
 	 */
 	public function setActionName(string actionName) -> void
 	{
-		let this->_actionName = actionName;
+		let this->_actionName = lcfirst(join("", array_map("ucfirst", preg_split("/[_-]+/", actionName))));
```
niden: I got dizzy reading this :D :D :D

niden (approved):

niden:
@axxeld Thank you!

niden:
Resolves #13555

sergeyklay:
@niden Could you please update the change log

Issue #13143, opened 2017-11-02
trap invalid opcode .... in phalcon.so


ruudboon:
I'm trying to run a phalcon docker build on coreOS using docker swarm.
My dockerfile is based from phalconphp/php:alpine-3-php7

Locally the image works fine (Mac).

I'm seeing the following errors: 

```
[02-Nov-2017 18:39:12] WARNING: [pool www] child 86 exited on signal 4 (SIGILL - core dumped) after 315.152352 seconds from start
```

```
[522081.446459] traps: php-fpm[18560] trap invalid opcode ip:7fd1ddb83590 sp:7ffcd30b2b20 error:0 in phalcon.so[7fd1dd857000+686000]

```

```
/ # php --ri phalcon

phalcon


Web framework delivered as a C-extension for PHP
phalcon => enabled
Author => Phalcon Team and contributors
Version => 3.2.2
Build Date => Aug 18 2017 13:56:35
Powered by Zephir => Version 0.9.9-868cb1f92b

Directive => Local Value => Master Value
phalcon.db.escape_identifiers => On => On
phalcon.db.force_casting => Off => Off
phalcon.orm.events => On => On
phalcon.orm.virtual_foreign_keys => On => On
phalcon.orm.column_renaming => On => On
phalcon.orm.not_null_validations => On => On
phalcon.orm.exception_on_failed_save => Off => Off
phalcon.orm.enable_literals => On => On
phalcon.orm.late_state_binding => Off => Off
phalcon.orm.enable_implicit_joins => On => On
phalcon.orm.cast_on_hydrate => Off => Off
phalcon.orm.ignore_unknown_columns => Off => Off
phalcon.orm.update_snapshot_on_save => On => On
phalcon.orm.disable_assign_setters => Off => Off
```
```
/ # php -v
PHP 7.1.5 (cli) (built: May 13 2017 10:49:18) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.1.5, Copyright (c) 1999-2017, by Zend Technologies
    with Xdebug v2.5.3, Copyright (c) 2002-2017, by Derick Rethans
```

ruudboon:
Upgraded to php 7.1.11 and phalcon 3.2.4, still same issue.

ruudboon:
Pin pointed the issue to the router.
```
<?php
[...]

PR #13800, opened 2019-02-04
Removed -march

ruudboon:
Removed march to be cross cpu compatible. Mtune isn't used because of -march
Solves #13143

Hello!

* Link to issue:

Small description of change:

Thanks

### DIFF
--- a/build/install
+++ b/build/install
@@ -24,7 +24,7 @@
 
 # Check best compilation flags for GCC
 export CC="gcc"
-export CFLAGS="-march=native -mtune=native -O2 -fomit-frame-pointer"
+export CFLAGS="-mtune=native -O2 -fomit-frame-pointer"
 export CPPFLAGS="-DPHALCON_RELEASE"
 
 # Set defaults


sergeyklay:
@ruudboon The best approach is to "ask" the compiler which flags it supports and use only them. This flag is very useful. It allows the extension to be "closer to the hardware". However, this has some limitations, of course. We used this flag in cases where we just wanted to install the Phalcon using the command line. I want to invite you to do a little research and make it more elegant :) Feel free to start from https://wiki.gentoo.org/wiki/GCC_optimization

sergeyklay:
Also you're doing PR to the master branch

ruudboon:
Sorry, this was indeed in the wrong branch. Will fix that if needed.
I'm not a C expert at all and was already happy that I could pinpoint the issue. To me it feels like we should build as compatible as possible. Then if you want to fine tune it for your environment you can always do a custom build.
With what flags are we building our binary packages at this point? Again, I'm not an expert but if we build with this settings probably there will be issues if you run it on a different CPU.

ruudboon:
Closing this. If needed will create a new pull on the correct branch.

PR #13821, opened 2019-02-12
preventing call to undefined method isSent

Idrinth:
Hello!

* Type: bug fix
* Link to issue: -

Small description of change:

Currently isSent is only implemented by `Response` and not part of the `ResponseInterface`. Therefore calling it on instances of the interfaces is not a reasonable solution.
Adjusting the interface would break backwards compatibility, so adding the check seems like the more sensible option.

### DIFF
--- a/phalcon/mvc/micro.zep
+++ b/phalcon/mvc/micro.zep
@@ -1024,8 +1024,8 @@ class Micro extends Injectable implements \ArrayAccess
 			/**
 			 * Automatically send the response
 			 */
-			 if !returnedValue->isSent() {
-				returnedValue->send();
+			 if !response->isSent() {
+				response->send();
 			 }
 		}
 	}
--- a/phalcon/mvc/micro.zep
+++ b/phalcon/mvc/micro.zep
@@ -24,6 +24,7 @@ use Phalcon\Mvc\Micro\Exception;
 use Phalcon\Di\ServiceInterface;
 use Phalcon\Mvc\Micro\Collection;
 use Phalcon\Mvc\Micro\LazyLoader;
+use Phalcon\Http\Response;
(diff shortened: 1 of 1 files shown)


phalcon/mvc/micro.zep
```diff
@@ -1008,17 +1008,27 @@ class Micro extends Injectable implements \ArrayAccess
 		 */
 		if typeof returnedValue == "object" {
 			if returnedValue instanceof ResponseInterface {
-				/**
-				 * Automatically send the response
-				 */
-				 if !returnedValue->isSent() {
-				 	returnedValue->send();
-				 }
+				this->sendIfUnsent(returnedValue);
```
Jeckerson: You make same check twice.

```suggestion
if method_exists(response, "isSent") && !returnedValue->isSent()
```

Idrinth: may need to reintegrate the method, completely overlooked that check, my bad.

Idrinth:
I'm not sure how `CacheTest::testDataFileCacheIncrement Failed asserting that 7 matches expected null.` could be related. Do you have any idea where I could dig for the connection between these changes and the failure, @Jeckerson ?

CameronHall:
@Idrinth Thank you for your contribution!

We are no longer accepting pull requests to `3.4.x`. If this change is still applicable to the `4.0.x` branch can you please target that.

Idrinth:
still an issue in master and 4.0.x, but with 4 being breaking it might be a better choice to add the missing method to the interface, wouldn't it @CameronHall ?

CameronHall:
Absolutely @Idrinth!

Idrinth:
replaced with https://github.com/phalcon/cphalcon/pull/13836 to solve issue in 4.0.x

Issue #13610, opened 2018-11-20
Discussion if Phalcon\Config::merge should also accept a basic array


dschissler:
This is the current merge code of `Phalcon\Config`.  Is there a reason why `::merge` is unable to take a basic array?

```zephir
	public function merge(<Config> config) -> <Config>
	{
		return this->_merge(config);
        }
```

```zephir
	protected final function _merge(<Config> config, var instance = null) -> <Config>
        {
```

  I've found myself overloading this `::merge` on a few occasions and I'm wondering if it shouldn't be relaxed to:

```zephir
	public function merge(var config) -> <Config>
```

and then with a simple `switch typeof config` to do the conversion if its an array.  Is there any reason why this is a bad idea, has noticeably reduced performance?

dschissler:
What about for Phalcon 4 using a "iterable pseudo-type"?  While writing my own extension and reviewing the Phalcon code I often feel that much of it has been written around the limitations of Zephir and of course earlier PHP versions.

http://php.net/manual/en/migration71.new-features.php#migration71.new-features.iterable-pseudo-type

I've already promoted the idea that Phalcon 4 Zephir should require > PHP 7.1.  This would be a nice feature that would likely be free.

dschissler:
@CameronHall Is there a math or CS term for "merging" that touches on set theory, etc?  I'm always for that if it is keeping with established terms, but otherwise having a "merge" and "append" that do very similar things but differing in a way that requires reading the docs just seems kind of funky.

CameronHall:
Not that I'm aware of. You're right, I retract what I said. It would be kind of funky and really unnecessary to add another layer of complexity. I don't see any reason why we can't just make `Phalcon\Config::merge` accept arrays.

dschissler:
Code was merged.

PR #13828, opened 2019-02-14
#13610 Allow Config::merge to accept array.

dschissler:
new feature
Issue #13610

If this code is acceptable then I will create a change log entry as well as possibly a test.

I modified the Config::merge method to allow arrays.  I first tried the following but I decided to go with a ternary:

```zephir
let config = is_array(config) ? new Config(configParam);
```

Another way of doing it but it seems too much:

```zephir
	public function merge(var configParam) -> <Config>
	{
		var config;

		switch typeof configParam {
			case "array":
				let config = new Config(configParam);
				break;
			case "object":
				let config = configParam;
				break;
			default:
				throw new \Exception("Invalid data type for merge.");
		}

		return this->_merge(config);
	}
```

### DIFF
--- a/phalcon/config.zep
+++ b/phalcon/config.zep
@@ -201,8 +201,13 @@ class Config implements \ArrayAccess, \Countable
 	 * $globalConfig->merge($appConfig);
 	 *</code>
 	 */
-	public function merge(<Config> config) -> <Config>
+	public function merge(var configParam) -> <Config>
 	{
+		var config;
+
+		// Allow merge to accept array.
+		let config = is_array(configParam) ? new Config(configParam) : configParam;
+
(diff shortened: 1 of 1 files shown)


niden:
@dschissler I prefer the second option. It sees a bit of an overkill but it gives more information as to what happened (exception).

Please add a test if you can and update the changelog

Thanks!

dschissler:
@niden The reason that I went with this option is that `_merge` already checks for `<Config>` in the signature.  If you still feel that way I'll do the switch version, but sometime in the evening.

niden:
@dschissler Why did you close this?

niden:
Never mind I saw the other PR

Issue #13593, opened 2018-11-13
PSR-11 4.0.x Use DI Presets instead of clunky factory defaults


dschissler:
I've found the Phalcon DI factory default classes to be very clunky and inflexible and so I designed a new method of creating boring default services.  These di presets are more flexible because they are both a DI that can be created and used directly and they also have the ability to extend an existing DI.  Note that I wrote it this way since at the time I couldn't find a better way to accomplish the task while working around Zephir's inflexibility.

```zephir
namespace Phalcon\Di;

use Phalcon\DiInterface;

interface PresetInterface
{

    public static function configure(<DiInterface> di);
}
```

```zephir
namespace Phalcon\Di\Preset;

use Phalcon\Di\PresetInterface;
use Phalcon\Di;
use Phalcon\DiInterface;

class Cli extends Di implements PresetInterface
{

    public static function configure(<DiInterface> di)
    {
        var serviceName, className, services;

        let services = [
            "dispatcher"           : "Phalcon\\Cli\\Dispatcher",
            // ...
        ];

        for serviceName, className in services {
            di->set(serviceName, className, true);
        }
    }

    public function __construct()
    {
        parent::__construct();

        self::configure(this);
    }
}
```

If a no-clobber setting option was added to `::configure` then a DI (from preset or vanilla) can be configured using multiple presets.  This allows extensions/packages to provide a preset to automatically configure service classes that they include in however mechanism that they want to.

So for example.

```php
$di = new Phalcon\Di;
Phalcon\Di\Preset\Cli::configure($di);
MyExtension\Di\Preset\Dev::configure($di);
```

and then later in some other bootstrap setup after the composer autoload has been brought in:

```php
Dschissler\Di\Preset\BigStuff::configure($di);
```

Also for simpler projects the standard di factory default can be instantiated directly and that is that.

```php
$di = new Phalcon\Di\Preset\Cli;
```

Now keep in mind that I'm doing things differently in my project but just to demonstrate the flexibility of this approach.

Additionally, I'm quite offended by the web-centric naming of the current factory defaults and if this proposal is not to be accepted on its technical merits then I demand that it be done so to remedy this quite offensive and incendiary language so that web and cli can exist without users of either run mode being discriminated against.

As an example of this wanton Phalcon language just observe the following:

```php
Phalcon\Di\FactoryDefault\Cli;
Phalcon\Di\FactoryDefault;
```

Do you see!  There is no web there for the web factory default.

[...]

PR #13837, opened 2019-02-18
Add DI presets, a new take on DI factory defaults.

dschissler:
Meh!

* Type: new feature
* Link to issue: [#13593](https://github.com/phalcon/cphalcon/issues/13593)

DI Presets are a new take on the DI factory defaults.  You can instantiate a DI preset class directly or instead use the `configure` method on an existing DI.  The default for `configure` is to clobber any existing services but if `false` is passed as the second argument then the existing services can be preserved.  This is especially useful to allow third party developers to provide a more complete experience.  For example a plugin provided by composer could then be expected to provide a service which could be provided to the DI through the call `Dschissler\MyPlugin\Di\Preset::configure($di, false)`

The Presets at a glance.

```zephir
namespace Phalcon\Di;

use Phalcon\DiInterface;

interface PresetInterface
{
    public static function configure(<DiInterface> di, bool clobber = true);
}
```

```zephir
class Web extends Di implements PresetInterface
{

	public static function configure(<DiInterface> di, bool clobber = true)
	{
		let services = [
			"annotations":        new Service("Phalcon\\Annotations\\Adapter\\Memory", true),
			"assets":             new Service("Phalcon\\Assets\\Manager", true),
			// and so on
		];

		for serviceName, className in services {
			if clobber {
				di->set(serviceName, className, true);
			} else {
				di->attempt(serviceName, className, true);
			}
		}
	}

	public function __construct()
	{
		parent::__construct();

		self::configure(this);
	}
}
```

There is potential room for optimization by allowing the default `clobber` option to set all of the services directly or at least in one pass without calling an expensive function each time.  Other than potentially negligible performance impact (which can be mitigated) the presets are simply superior.

**NOTICE:** This is the first commit to this tech for seeking comments.  There is no change log entry yet because that would complicate merging and I don't expect this to be a painless change.  The tests are more involved, especially since the CLI factory default inherits from the base FD (which I don't believe is correct).

@niden I'm now PRing from private branches so that I can have several of these going at once.  As it nears completion I'll push in the conflicting bits.  I already have the changelog entry done which I'm keeping separately.

- Added DI Presets, remove DI factory defaults. [#13593](https://github.com/phalcon/cphalcon/issues/13593)

### DIFF
--- a/phalcon/di/presetinterface.zep
+++ b/phalcon/di/presetinterface.zep
@@ -1,0 +1,15 @@
+namespace Phalcon\Di;
+
+use Phalcon\DiInterface;
+
+/**
+ *
+ */
+interface PresetInterface
+{
+
+    /**
+     *
+     */
+    public static function configure(<DiInterface> di, bool clobber = true);
+}
--- a/phalcon/di/presets/cli.zep
+++ b/phalcon/di/presets/cli.zep
@@ -1,0 +1,74 @@
+
+/**
+ * This file is part of the Phalcon Framework.
+ *
+ * (c) Phalcon Team <team@phalconphp.com>
+ *
+ * For the full copyright and license information, please view the LICENSE.txt
+ * file that was distributed with this source code.
+ */
+
+namespace Phalcon\Di\Presets;
+
+use Phalcon\Di;
+use Phalcon\Di\PresetInterface;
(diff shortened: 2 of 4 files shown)


dschissler:
DAMNIT! I'll have to close this because my commits got cross wired.

PR #13838, opened 2019-02-18
[#13593] Add DI Presets.

dschissler:
Meh! 

* Type: new feature
* Link to issue: [#13593](https://github.com/phalcon/cphalcon/issues/13593)

- [ ] Tests will come in later commit or PR.

DI Presets are a new take on the DI factory defaults.  You can instantiate a DI preset class directly or instead use the `configure` method on an existing DI.  The default for `configure` is to clobber any existing services but if `false` is passed as the second argument then the existing services can be preserved.  This is especially useful to allow third party developers to provide a more complete experience.  For example a plugin provided by composer could then be expected to provide a service which could be provided to the DI through the call `Dschissler\MyPlugin\Di\Preset::configure($di, false)`

The Presets at a glance.

```zephir
namespace Phalcon\Di;

use Phalcon\DiInterface;

interface PresetInterface
{
    public static function configure(<DiInterface> di, bool clobber = true);
}
```

```zephir
class Web extends Di implements PresetInterface
{

	public static function configure(<DiInterface> di, bool clobber = true)
	{
		let services = [
			"annotations":        new Service("Phalcon\\Annotations\\Adapter\\Memory", true),
			"assets":             new Service("Phalcon\\Assets\\Manager", true),
			// and so on
		];

		for serviceName, className in services {
			if clobber {
				di->set(serviceName, className, true);
			} else {
				di->attempt(serviceName, className, true);
			}
		}
	}

	public function __construct()
	{
		parent::__construct();

		self::configure(this);
	}
}
```

There is potential room for optimization by allowing the default `clobber` option to set all of the services directly or at least in one pass without calling an expensive function each time.  Other than potentially negligible performance impact (which can be mitigated) the presets are simply superior.

As an example I allow services to be set up many ways in my framework.  Here is a yaml snippet.

I start with a vanilla DI and then load over it.  The presets are loaded with force/clobber off.  `cli.php` returns a function that is basically the same as a service provider class.  The directories ending with `/` have one service per file.  Do as you want.

```yaml
service:
- Perch\Di\Preset\Cli
- ./CliServiceProvider.php
- ./cli.php
- ./common/
- ./cli/
```

It could easily look like this.

```yaml
service:
- Perch\Di\Preset\Cli
- ./CliServiceProvider.php
- Bobs\Extension\Di\Preset\User
- Larrys\Extension\Di\Preset\Security
```

So Bob's and Larry's presets from a package are run non-clobber/non-force.  You still get a chance to set them up manually but the package dudes can offer an easy path.  The trick is that for the non-zephir provided presets that you first need to run the composer autoload.  That can't be helped except to bake that earlier into your bootstrap.

**NOTICE:** This is the first commit to this tech for seeking comments.  There is no change log entry yet because that would complicate merging and I don't expect this to be a painless change.  The tests are more involved, especially since the CLI factory default inherits from the base FD (which I don't believe is correct).

@niden I'm now PRing from forked branches so that I can have several of these going at once.  As it nears completion I'll push in the conflicting bits.  I already have the changelog entry done which I'm keeping separately.

- Added DI Presets, remove DI factory defaults. [#13593](https://github.com/phalcon/cphalcon/issues/13593)


niden (dismissed):

niden:
@dschissler 

This is solid. I would however like to change the name of `clobber` I had to google it to find what it is :)

Let me know and we will merge this.

dschissler:
Ok but clobber is quite a well used word.  Force I guess but clobber just does it.  I didn't make that shit up for example.

dschissler:
Also I think that due to how disruptive it is that we just get this into the tree and then make a good strategy of removing the factory defaults.  Later there will be opportunity to optimize the presets through tricks.  At the moment the FD tests are a bit intimidating.  If you get this in then I can make another PR later to complete the transition.

niden:
I speak English and had to google the word to understand what it actually meant and looking at the code I understood. One cannot expect all developers to be proficient in English. A more obvious/descriptive word for the variable will definitely help.

_Also I think that due to how disruptive it is that we just get this into the tree and then make a good strategy of removing the factory defaults_

Not sure I get this. You mean that the goal is to remove completely the factory defaults? If so, and correct me if I am wrong here, it was my understanding that this PR would offer a way to set your own factory defaults using a config method. Am I on the right track here?

Finally, I have seen this in other issues, as well as this one, kindly tone down the colorful  comments. 

Thanks

dschissler:
> Not sure I get this. You mean that the goal is to remove completely the factory defaults? If so, and correct me if I am wrong here, it was my understanding that this PR would offer a way to set your own factory defaults using a config method. Am I on the right track here?

My goal is to first replace the cli and web related factory defaults with these.  It might need another look because the current CLI FD inherits over the normal "web" FD.  I think that they should be just completely separate.  So firstly the current FD role must be assumed by the presets.  This inheritance makes the 

Then as an addon users can create their own preset classes provided by a composer package.  Then anyone using their tech would load the preset with clobber/force off before or after setting up their services.  In the case of my Perch Framework I would use `Phalcon\Di\PresetInterface` instead of `Perch\Di\PresetInterface`.

It may be that over the course of the Phalcon v4 lifecycle that we discover that a different preset API would be better, but as a first step this is vastly better than the current factory defaults.

dschissler:
@niden As an example I allow services to be set up many ways in my framework.  Here is a yaml snippet.

I start with a vanilla DI and then load over it.  The presets are loaded with force/clobber off.  `cli.php` returns a function.  The directories ending with `/` have one service per file.  Do as you want.

```yaml
service:
- Perch\Di\Preset\Cli
- ./CliServiceProvider.php
- ./cli.php
- ./common/
- ./cli/
```

It could easily look like this.

```yaml
service:
- Perch\Di\Preset\Cli
- ./CliServiceProvider.php
- Bobs\Extension\Di\Preset\User
- Larrys\Extension\Di\Preset\Security
```

So Bob's and Larry's presets from a package are run non-clobber/non-force.  You still get a chance to set them up manually but the package dudes can offer an easy path.  The trick is that for the non-zephir provided presets that you first need to run the composer autoload.  That can't be helped except to bake that earlier into your bootstrap.

niden:
> My  goal is to first replace the cli and web related factory defaults with these. 

Again not sure I fully understand this. From what I see this is pretty much done, other than the variable name. It offers a way for someone to create presets for their DI container and not use the Phalcon supplied ones in the FactoryDefault. Was there consideration to remove the `FactoryDefault` classes all together?

In all honesty one could do the same thing by extending the `FactoryDefault` class but this is another option that definitely helps developers.

dschissler:
@niden the factory defaults don't have `::configure` with option of force/clobber or not.  With this you can use more than more default setup for the DI.  You can instantiate directly or call `::configure` on your already partially setup DI.  It always framework and package developers to offer more of a drop in

niden:
@dschissler My point is that as things currently stand, one can easily create their own FactoryDefault with the services they need and utilize that vs the one supplied by the framework. That is customization as things are now.

This PR offers an additional way of customizing your DI services, which is great. 

My question is whether there is more after this PR. Maybe I misunderstood but you mentioned this:

> My goal is to first replace the cli and web related factory defaults with these.

What are you going to replace? If you add the `configure` which this PR does, you "move" the service definitions to the `configure` method which can be overriden by a custom class that extends the DI one. 

Is it your goal to completely remove the `FactoryDefault` class and instead use Web/Cli as per this PR?

dschissler:
> Is it your goal to completely remove the `FactoryDefault` class and instead use Web/Cli as per this PR?

Yes.

> My point is that as things currently stand, one can easily create their own FactoryDefault with the services they need and utilize that vs the one supplied by the framework.

Again they currently can't use one or many factory defaults in an additive way with non-clobber/non-force.  This will also add an interface so that people can know what to expect across projects and third-party support.

Today or tomorrow I will begin to redo the tests and to remove the DI factory defaults.  This will break a great many projects but the fix will be easy by just changing the class name.  I think that for early adopters before an upgrade guide I will add a warning through Phalcon 4 beta 1 or 2 in the factory defaults that they are going away.

david-duncan:
> > Is it your goal to completely remove the `FactoryDefault` class and instead use Web/Cli as per this PR?

> Yes.

I'd prefer not. The factoryDefault is nice and slim, and does not have iteration like your requested changes do.

I like minimizing TTFB. All of this configuration has a cost for no real use case.

Heads up - none of your website links work @dschissler

niden:
Removing the `FactoryDefault` is not an option. This is why I was asking repeatedly what is your vision about this. 

Switching the code around so that `FactoryDefault` can use the `configure` and all default declarations are there is perfectly fine. 

> Again they currently can't use one or many factory defaults in an additive way with non-clobber/non-force. This will also add an interface so that people can know what to expect across projects and third-party support.

On the above, without the `configure` all one had to do is extend the `FactoryDefault` and replace the `__construct` method with whatever one wants. As to add or remove stuff from the container, that happens once in the application bootstrap so I fail to see the need for doing this. The `configure` however does offer one more way for people to setup their DI.

As for your comment about the interface, I really don't see the purpose of it, The container is just that, a container and it implements already an interface that has services in it (which in turn implement the service interface). I cannot envision the interconnection that you mention. Do you have an example to help with that? Unless you are thinking about other applications utilizing the DI container of Phalcon as a plugin to request services loaded there.

dschissler:
@niden 
I have two examples.  In the first example the `::configure` method doesn't need it to implement the preset interface.  In the second method it basically does need it since there are multiple ways to define services.  I have it setup to call the static `configure` on classes which implement the preset interface.  Why?  Because I will be calling `::configure` on it and that is better than trying to call configure on a random full class name.

```php
// load your composer stuff

$di = new Phalcon\Di;
Phalcon\Di\Preset\Cli::configure($di);

// Now addons can provide required services in force and non-force.
MyCode\Di\Preset\Dev::configure($di, false);
```

This is the service section of each "boot mode" (app mode).  I have all of the stuff in there now to be an example.  I've written about this much here in the PR and the issue.

```yaml
service:
- Perch\Di\Preset\Cli
- ./CliServiceProvider.php
- ./cli.php
- ./common/
- ./cli/
- MyCode\Di\Preset\Dev
```

Again, MyCode can provide more things out of the box than before.  I'll try to address @david-duncan 's efficiency concerns but at this point I'm just getting tired explaining myself as to the basic premis.  I'm going to try to make it more efficient for some cases but other than that I don't need this to be a part of Phalcon in order to have my glory.  Since this can be used separately then I'll likely end up doing that and instead focus on broken Phalcon things that I don't want to have to live with for a few years.  I'm overall happy that I've been able to get through so many other PRs and I never expected to win them all.

Issue #13815, opened 2019-02-09
NFR Rename Volt "compiled"


v00v:
Request to remove "compiled" from ``setOptions`` and simply use it like this;
Because eveyone who uses Volt know that this ``setOptions`` are "compiled" :-)

```php
$di->set('view', function() {
  $view = new View();
      $view->setViewsDir('../app/views/');
      $view->registerEngines([
        '.volt' => function($view, $di) {
        $volt = new Volt($view, $di);
          $volt->setOptions([
            'path' => '../app/cache/',  // "compiledPath"
            'extension' => '.html',  // "compiledExtension"
            'separator' => '_',  // "compiledSeparator"
            'stat' => true,
            'always' => true,  // "compileAlways"
          ]);
          return $volt;
        }]);
  return $view;
});
```

I think by now you can guys see that LESS is MORE :-) Its better to have less typed alias, commands and functions. Because who wants to type ``someFunctionNameAgainAndAgain()``

To have backward compability you can simply create alias for this. So everyone who uses "compiled" it will still work, or simply note this in Phalcon changelog that it has been changed like few recent renames of functions you have done :-) Love it!

dschissler:
I like it without the alias.  Phalcon v3 was around for a very long time.  I think that it bough the project some leeway to break stuff in v4.  If anything I would emit a warning if the "compiled" key exists and then just remove that warning later in the v4 cycle.

If people can agree that removing the "compiled" level is good then I could do this PR.  I find that what Phalcon really lacks in its issues is a "Sounds good, lets do it".  So issues just sit there maybe in agreement but then nobody does it.  So @niden it would be great if you could get in the habit of "in theory" approving changes.  Of course that doesn't mean you are bound to the implementation.  I have no idea if you would actually accept something or if you are just being kind and tagging it.  So I pretty much just only do stuff that I feel that I can't live without.  Nothing else is worth the uncertainty.

niden:
@dschissler I always try to add labels to issues. Aside from the glaring obvious `not a bug` or won't fix`, the RFC one is where we need comments from the community. Anything else pretty much gets a NFR/Enhancement or Bug.

Any of these that fit the workload of the current version are added in the v4 project here:

https://github.com/phalcon/cphalcon/projects/3

Anything that is not, will be introduced in v4.1 etc. I cannot develop more than what the current project has, otherwise we will end up with an alpha version that lasts a while year. The tasks on the project are pretty much heavy with PSR and all that.

[...]

PR #13840, opened 2019-02-18
Changed Volt compiler options.

dschissler:
Hello!

* Type: new feature
* Link to issue: #13815

Small description of change: 

Changed Volt compiler options with to `compiled` or `compile` prefix to be deprecated. Now use the key without the prefix. I'm simplifying the API.

**NOTICE:** I'm trying something different by using a different forked branch for each PR.  If this change is acceptable then I'll make the change log entry to get this all of the way through.  In order to avoid merge conflicts this should be done as late as possible.  I hope to be able to have several of these going at once for higher throughput.

### DIFF
--- a/phalcon/mvc/view/engine/volt/compiler.zep
+++ b/phalcon/mvc/view/engine/volt/compiler.zep
@@ -2504,9 +2504,9 @@ class Compiler implements InjectionAwareInterface
 	 */
 	public function compile(string! templatePath, bool extendsMode = false)
 	{
-		var stat, compileAlways, prefix, compiledPath, compiledSeparator, blocksCode,
-			compiledExtension, compilation, options, realCompiledPath,
-			compiledTemplatePath, templateSepPath;
+		var blocksCode, compilation, compileAlways, compiledExtension, compiledPath,
+			compiledSeparator, compiledTemplatePath, optionKey, options, prefix,
+			realCompiledPath, stat, templateSepPath;
 
 		/**
 		 * Re-initialize some properties already initialized when the object is cloned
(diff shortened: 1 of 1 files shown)


sergeyklay (approved):
LGTM. 

/cc @phalcon/framework-team

dschissler:
I looked up "LGTM" and its "literally garbage to me" :(((.  Then I saw that its a PR system.

@sergeyklay Hey so for these two PRs that you approved.  Let me then just add the change log entry one at a time.  I don't know the best way to get these all through.  I don't want to have a PR blocking the rest of my PRs (as you can see I have many).

sergeyklay:
@dschissler I will keep this open for a day or two to allow other contributors a chance to speak up, and to take a fresh look at it later. But at the moment it seems good to me.

CameronHall (approved):

niden:
I would have liked a couple of tests on this but it's perfectly fine. Merging

Thank you @dschissler

sergeyklay:
@dschissler @niden Question: Should we introduce deprecations for a brand new version? Can we just remove old config param?

dschissler:
I think that for things like this that the deprecations aren't needed but it was asked for.

niden:
Yeah, now that I think about it, deprecation is not needed since this can be documented. We can just remove the `trigger_error`. 

Thanks guys

dschissler:
@niden So we would use both keys?  Then later in 4.x add them back?  It will also simplify the code a bit.

niden:
@dschissler no we can just use the non `compile` prefixed ones for v4. All we will have to do is ensure that the upgrade documentation reflects that change so that developers make that change in their code.

dschissler:
@niden Ha ok.  It will end up being almost a full revert with just replacing the keys.  I'll get that new PR to you guys soonish.

dschissler:
@niden @sergeyklay I have a good idea!  Why don't we leave the volt compiler keys as they are now and then remove the old keys later say after the first beta?  This will give early adopters some easy deprecation warnings.  I mean, the code is already there and it will help some people out.  It will be very easy to remove them later (before the 4.0 release).  I was only following a request by adding in the deprecation warnings but now it was determined the old keys aren't needed we might as well make the best out of the work that was done.

sergeyklay:
@dschissler I'm ok with removing he old keys later say after the first beta

sergeyklay:
@niden Could you please update change log

Issue #13647, opened 2018-12-10
Discusion 4.0.x Way of improving generic database setup


dschissler:
Here is an example of a typical generic database setup for a Phalcon 3 base project.  Notice that for Postgres the charset needs to be unset or apparently something bad will happen.  I'm thinking that it would be very nice if the builtin host based RDBSs didn't need this and would also default to utf8 by default.  I think that going forward about 99.5% of people want utf8 and fine if they don't then they can set it themselves.  It would be nice to do a bit of default handholding in Phalcon 4.  Does anyone else think that Phalcon should automatically set some PDO options and also to assume utf8 by default?

```php
    $class = $config->database->adapter;
    $params = [
        'host'     => $config->database->host,
        'username' => $config->database->username,
        'password' => $config->database->password,
        'dbname'   => $config->database->dbname,
        'charset'  => $config->database->charset,
    ];

    if ($class == Postgresql::class) {
        unset($params['charset']);
    }

    $connection = new $class($params);
```

CameronHall:
I agree! Worth noting that for the MySQL adapter we should set the default charset to `utf8mb4.

dschissler:
Oh right!  Good one.  Modern MySQL is pretty good but it sure has a funky history and PHP MySQL is even funkier.

CameronHall:
It's both. The character set can be defined on the DB, table and the connection. When I create my PDO connection I do this;
```php
PDO::MYSQL_ATTR_INIT_COMMAND => 'SET NAMES utf8mb4'
```

CameronHall:
I'd never actually thought much about it until now. I found a [StackOverflow](https://stackoverflow.com/questions/24356090/difference-between-database-table-column-collation) answer explaining the order of precedence. 

> Order of precedence with topmost being least precedence:
> 
> - Server collation
> - Connection-specific collation
> - Database collation
> - Table collation
> - Column collation
> - Query collation (using CAST or CONVERT)

So I suppose setting the character set on the connection doesn't really do a whole lot when it can be overriden by everything else anyway (and rightfully so).

dschissler:
[...]

PR #13841, opened 2019-02-18
[13647] - Allow default PDO DSN values.

dschissler:
Hello!

* Type: new feature
* Link to issue: [#13647](https://github.com/phalcon/cphalcon/issues/13647)

Small description of change:

This initial commit is an early proof-of-concept for providing automatic DSN values for each PDO adapter.  At the moment only the MySQL adapter is providing anything with `charset=utf8mb4`.  As far as the other included adapters postgres and sqlite don't seem to support that option.  The goal is to remove the need to specify the charset in PHP code for 99% of the users.  This will ease the learning curve as well as allow frameworks and skeletons to have an easier time.

**NOTICE:** I'm trying something different by using a different forked branch for each PR.  If this change is acceptable then I'll make the change log entry to get this all of the way through.  In order to avoid merge conflicts this should be done as late as possible.  I hope to be able to have several of these going at once for higher throughput.

### DIFF
--- a/phalcon/db/adapter/pdo.zep
+++ b/phalcon/db/adapter/pdo.zep
@@ -52,7 +52,7 @@ abstract class Pdo extends Adapter
 	protected _pdo;
 
 	/**
-	 *
+	 * Returns PDO adapter DSN defaults as a key-value map.
 	 */
 	abstract protected function getDsnDefaults() -> array;
 
--- a/phalcon/db/adapter/pdo.zep
+++ b/phalcon/db/adapter/pdo.zep
@@ -52,6 +52,11 @@ abstract class Pdo extends Adapter
 	protected _pdo;
 
 	/**
+	 *
+	 */
+	abstract protected function getDsnDefaults() -> array;
+
+	/**
 	 * Constructor for Phalcon\Db\Adapter\Pdo
 	 */
 	public function __construct(array! descriptor)
@@ -75,7 +80,6 @@ abstract class Pdo extends Adapter
 	{
 		return this->_affectedRows;
 	}
-
 
 	/**
 	 * Starts a transaction in the connection
@@ -246,8 +250,8 @@ abstract class Pdo extends Adapter
 	 */
 	public function connect(array descriptor = null) -> bool
 	{
-		var username, password, dsnParts, dsnAttributes,
-			persistent, options, key, value;
+		var username, password, dsnDefaults, dsnDefaultValue, dsnParts,
+			dsnAttributes, persistent, options, key, value, pdoKey;
 
 		if empty descriptor {
 			let descriptor = (array) this->_descriptor;
@@ -283,12 +287,15 @@ abstract class Pdo extends Adapter
 		/**
 		 * Check for \PDO::XXX class constant aliases
 		 */
-        for key, value in options {
-            if typeof key == "string" && defined("\PDO::" . key->upper()) {
-                let options[constant("\PDO::" . key->upper())] = value;
-                unset options[key];
-            }
-        }
+		for key, value in options {
+			if typeof key == "string" {
+				let pdoKey = "\PDO::" . key->upper();
+				if defined(pdoKey) {
+					let options[constant(pdoKey)] = value;
+					unset options[key];
+				}
+			}
+		}
 
 		/**
 		 * Check if the connection must be persistent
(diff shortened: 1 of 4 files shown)


phalcon/db/adapter/pdo/sqlite.zep
```diff
@@ -43,6 +43,26 @@ class Sqlite extends PdoAdapter
...
+		return [];
+	}
+
+	/**
+	 * Constructor for Phalcon\Db\Adapter\Pdo\Sqlite
+	 */
+	public function __construct(array! descriptor)
+	{
+		if isset descriptor["charset"] {
+			trigger_error("Postgres does not allow the charset to be changed in the DSN.");
```
sergeyklay:
```suggestion
			trigger_error("Sqlite does not allow the charset to be changed in the DSN.");
```

phalcon/db/adapter/pdo.zep
```diff
@@ -51,6 +51,11 @@ abstract class Pdo extends Adapter
 	 */
 	protected _pdo;
 
+	/**
+	 *
+	 */
+	abstract protected function getDsnDefaults() -> array;
```
sergeyklay: Provide please documentation string

phalcon/db/adapter/pdo/postgresql.zep
```diff
@@ -45,6 +45,23 @@ class Postgresql extends PdoAdapter
 
 	protected _type = "pgsql";
 
+	protected function getDsnDefaults() -> array
```
sergeyklay: the same here

sergeyklay (changes requested):
I like the idea. But we need tests and please amend change log

phalcon/db/adapter/pdo/sqlite.zep
```diff
@@ -43,6 +43,26 @@ class Sqlite extends PdoAdapter
 
 	protected _type = "sqlite";
 
+	/**
+	 *
+	 */
+	protected function getDsnDefaults() -> array
```
sergeyklay: Provide please documentation string

phalcon/db/adapter/pdo.zep
dschissler: I think that the [Oracle Incubator](https://github.com/phalcon/incubator/blob/master/Library/Phalcon/Db/Adapter/Pdo/Oracle.php) will need to be updated shortly after this is merged.  So you'd need to start a 4.x incubator branch.

dschissler:
@sergeyklay 

> I like the idea. But we need tests and please amend change log

I will do the change log entry last to help prevent merge conflicts with my other branches.

CameronHall (approved):

phalcon/db/adapter/pdo.zep
```diff
@@ -283,12 +287,15 @@ abstract class Pdo extends Adapter
...
-        }
+		for key, value in options {
+			if typeof key == "string" {
+				let pdoKey = "\PDO::" . key->upper();
+				if defined(pdoKey) {
+					let options[constant(pdoKey)] = value;
+					unset options[key];
+				}
+			}
+		}
```
niden: Instead of trying to "guess" what the developer passed, why not leave it as simple `options` without trying to figure out what `\PDO` constant this is?

phalcon/db/adapter/pdo/mysql.zep
```diff
@@ -46,6 +46,17 @@ class Mysql extends PdoAdapter
...
 	protected _type = "mysql";
 
+	/**
+	 * Returns PDO adapter DSN defaults as a key-value map.
+	 */
+	protected function getDsnDefaults() -> array
+	{
+		// In modern MySQL the "utf8mb4" charset is more ideal than just "uf8".
+		return [
+			"charset" : "utf8mb4"
```
niden: :+1:

niden (changes requested):
This will definitely need tests. Other than a comment above solid work.

phalcon/db/adapter/pdo.zep
dschissler: @niden I believe that the class was already guessing in some way.  I'll have to get my head back into that but I tried to not disrupt it so much.  I'd be for a changing of the `options` API.  I also don't like the magic but the magic was already there.

niden: Since this is a big release and we _can break stuff_ so to speak, we can just have the options and let the developer pass whatever they want. I am also wondering if we even need to be checking for the `persistent` option since for PDO there is a particular option for it. 

Thoughts?

dschissler:
I'm closing this PR for now in favor of #13862.  I'll keep this branch alive so that we can go back to it if 13862 falls out of favour.

Issue #13440, opened 2018-07-16
[NFR]  Make shared instances removable


sergeyklay:
From @xueron 

> When I use Phalcon in a multi-process scene, I have to remove any DB connections copied from parent.

> so I add two methods: one check a shared instance for a service exists or not, and another to remove a shared instance. That we can recreate it if needed.

Refs: https://github.com/phalcon/cphalcon/pull/13285

sergeyklay:
@dragoonis Just assigned to you since you are currently working on ContainerInterface implementation in context of Phalcon's DI/ServiceLocator

victorgp89:
Any update about this?

dschissler:
I read the linked PR comments I would like to propose a whole new way of thinking about it.  I think that there is far too little cached short-circuiting around the [DI::get:226](https://github.com/phalcon/cphalcon/blob/b19d84a287477992cfabf258db8784486e140d1d/phalcon/di.zep#L226).  The [Service::resolve:120](https://github.com/phalcon/cphalcon/blob/b19d84a287477992cfabf258db8784486e140d1d/phalcon/di/service.zep#L120) is able to short-circuit from cache if its shared but then the DI doesn't use that state at all.

I'm proposing some logic like the following (it will be written for efficiency later):

```zephir
if service->isShared() && isset this->_sharedInstances[name] {
    return this->_sharedInstances[name];
}

try {
    let instance = service->resolve(parameters, this);
} catch ServiceResolutionException {
    throw new Exception("Service '" . name . "' cannot be resolved");
}

if instance->isShared() {
    let this->_sharedInstances[name] = instance
}
```

I'd like to point out that I made this DI+Service related commit for 4.0.x [right here](https://github.com/phalcon/cphalcon/commit/6766951396d2a097a80a7fba7ba2be0ff3741b4b).  I've been looking at this smelly code for a while now.  In hindsight its obvious as shit that the service name didn't need to be in the service constructor, but there it was.  My point is that `DI::get` shouldn't be treated as gospel and it is basically untouched since the moment it was ported from C to Zephir.  The core logic itself really isn't perfect.  Can we look there for a moment?

[...]

PR #13845, opened 2019-02-19
DI::get shared caching - small tweak

dschissler:
Hello!

This is a proof of concept for enhancing the DI::get for shared caching.  The behaviour in this version is slightly different because retrieving a shared instance with `::get` will not trigger the `di:afterServiceResolve` event.  I'm going to make another version that is a complete refactor of `::get`.

This is a proof of concept to help alleviate #13440 by creating a path for Phalcon to internally use `get` instead of `getShared`.  Its possible that there will be an edge case that will cause this new approach to not work out.

### DIFF
--- a/phalcon/di.zep
+++ b/phalcon/di.zep
@@ -203,9 +203,9 @@ class Di implements DiInterface
 	/**
 	 * Resolves the service based on its configuration
 	 */
-	public function get(string! name, parameters = null) -> var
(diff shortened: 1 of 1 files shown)


dschissler:
@sergeyklay I'm closing this issue for now.  I created two flavors of `DI:get` and this was the least impactful and complete.  I will keep the branch open for a while if this is ever needed.

Issue #13364, opened 2018-05-02
Remove Beanstalkd from Phalcon


sergeyklay:
I propose to remove [beanstalkd][1] from Phalcon and move this functionality to the incubator or other repo. This may be done for 4.0.0 version.

Actually beanstalkd isn't a popular queue system. In addition, beanstalk doesn't provides any unique feature (except a bit magic terminology e.g. `tube`, `peek-ready`, etc.). This project has [zero activity][2] for many years, unanswered questions like [Is this project still active][3], lack of any out of box HA tools, replication tools, cluster integration, etc.

In fact many of our users do not use `beanstalkd` and even did not use. I think that in the framework there should be truly top requested and most used tools.

[1]: https://github.com/kr/beanstalkd
[2]: https://github.com/kr/beanstalkd/pulse/monthly
[3]: https://github.com/kr/beanstalkd/issues/299

409H:
> In fact many of our users do not use beanstalkd and even did not use.

I'm curious, how did you poll this?

I'm actually a user that uses the `beanstalkd` functionality of Phalcon, though I don't see how moving it into Incubator will affect me much, but is it a task worth doing? What would be the pros of moving it out of the core and into Incubator?

niden:
@409H I too use `beanstalkd` on an environment with Phalcon. There have been quite a number of issues to maintain that environment, so we switched to AWS SQS to remove the burden. The maintenance cost for beanstalk became too much for us.

As for Phalcon, we are trying to keep the framework as slim as possible but as feature rich as possible. As mentioned by Serghei, the project has low activity (well zero actually) so we really don't want to support connection to a system that will potentially have bugs or is abandoned.

Moving the functionality to the incubator will have minimal impact to existing applications, since all you will have to do now is include the related composer package. 

HTH

Jurigag:
If we will add rabbitmq as replacement then fine.

davidtavarez:
rabbitmq is a good option

409H:
@niden I understand. Count me in favour of the proposal then.

ruudboon:
I’m using the beanstalkd a lot. The reason for me to go for beanstalk is the native integration with Phalcon.
+1 for me to keep it part of Phalcon

HanXHX:
Best way: a good refactoring with https://github.com/queue-interop/queue-interop

rpacker:
We use it, but either way, I'd vote to add some other queues such as RabbitMQ, and SQS.

[...]

PR #13850, opened 2019-02-22
T13364 beanstalk removal

niden:
Hello!

* Type: code removal
* Link to issue: #13364 

Removed Beanstalkd/Queue from Phalcon (new implementation will come next version)

Thanks


sergeyklay (dismissed):
LGTM. Please update change log

dschissler:
I think that it would be good to have something other than the Incubator to put stuff like this.  For example this is being removed from the Zephir port and will be possibly brought into PHP and the Incubator is advertised as a stepping stone to being a full Phalcon extension port.  It doesn't fit in Incubator but something more like "extras".  If that is too much work then I propose that the Incubator be renamed to something like "Extras".

Issue #13853, opened 2019-02-23
Using `spl_object_hash()` instead of `Di::isFreshInstance()` in the Dispatcher


dschissler:
I'm evaluating the use of [DI::wasFreshInstance()](https://github.com/phalcon/cphalcon/blob/b19d84a287477992cfabf258db8784486e140d1d/phalcon/dispatcher.zep#L495) in the `Phalcon\Dispatcher`.  I think that the `Di:wasFreshInstance` method is a bad design and that it could instead be achieved with a "handler" cache that stores whether a handler instance has been seen before.  If there is a miss on the handler cache using spl_object_hash then it could be assumed to be new and `wasFresh` could then be set to true.  Implementing something like this would simplify the DI.  Although I'm not positive on this I suspect that this design limitation goes way back to when Phalcon was changed to be HMVC.

niden:
This has been addressed

PR #13854, opened 2019-02-23
Dispatcher handlercache

dschissler:
…a handler hash cache.

Hello!  This is another proof of concept PR

* Type: code quality
* Link to issue: #13853

- [ ]  I did not write tests for this because a large refactor of the Dispatcher is needed.  This is only to remove one small wart.  Creating tests at this time would be premature.

This PR is attempting to solve the use of `DI::wasFreshInstance` in `Phalcon\Dispatcher`.  I think that `DI::wasFreshInstance` should be completely removed in 4.0.x since it is only used in this one place for this hack.  I'm attempting to see if a hash hit map of all handlers seen could be used instead of the use of `wasFreshInstance`.

I'm attempting to understand everything that would need to be changed to get the Dispatcher to run initialize first every time as first began by @virgofx but wow is it some involved code.  I'm not sure that I'm setup to change something so large at this point.  So I'm starting small.  This PR may never be merged or I could also grow it to be more encompassing.

### DIFF
--- a/phalcon/dispatcher.zep
+++ b/phalcon/dispatcher.zep
@@ -495,14 +495,6 @@ abstract class Dispatcher implements DispatcherInterface, InjectionAwareInterfac
 
 			let handler = dependencyInjector->getShared(handlerClass);
 
-			let handlerHash = spl_object_hash(handler);
-
-			// Ensure that the handler is new.
-			let wasFresh = isset this->_handlerHashes[handlerHash] ? false : true;
-			if wasFresh {
-				let this->_handlerHashes[handlerHash] = true;
-			}
-
 			// Handlers must be only objects
 			if typeof handler !== "object" {
 				let status = this->{"_throwDispatchException"}("Invalid handler returned from the services container", self::EXCEPTION_INVALID_HANDLER);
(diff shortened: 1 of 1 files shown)


dschissler:
It seems that it works.  I guess that it can be merged.

CameronHall (approved):

sergeyklay (approved):

niden:
Approving/merging this. Next time though we need tests to be 100% sure that nothing was broken.

Thank you @dschissler

dschissler:
@niden It really was not clear to me what kinds of tests would be needed to test this.  The dispatcher still needs an `initialize` refactor along the lines of what @virgofx proposed.  Perhaps it can be done then.  Something to ensure that initialize is only called once, at the right time.  Its already broken an funky, this is IMO just to remove DI `fresh` stuff.  With the other DI PR merged then I will make another PR to the DI to remove the fresh stuff (since only the dispatcher used it here) and to tighten it down.  This is a holistic removal of clunk and its difficult to do it all at once.

sergeyklay:
@dschissler I think you can just amend the description of the pull request with something like that "I did not provide tests for the following reasons ...". This is ok :) Anyway thank you for the contributting, and for helping us make Phalcon better.

dschissler:
@sergeyklay @niden I amended the description as to why no tests.  Now I will start the intimidating process of refactoring the dispatcher along what virgofx wrote about.  Its too complicated to just put on the cowboy hat.  I need to understand the scope better.

Issue #13847, opened 2019-02-20
Create DLLs for Phalcon 3.x and PHP 7.3


sergeyklay:
This issue was created from the issue's comment in #13511

@sergeyklay Thanks for clarification. But as phalcon 4 won't be stable soon and more important won't be backward compatible to 3.4, I highly suggest that you at least think about making 3.4.2 php 7.3 compatible without any other fixes or features (that's what phalcon 4 is for) . Otherwise many teams will be either stuck on php 7.2 or have to migrate their existing and working projects to the phalcon 4 which is probably a lot of effort also because much stuff has been removed, so it's not only renaming some method calls. The main point is that php 7.3 is stable and released while phalcon 4 is not. So currently there is no combination of both libraries in stable condition. You should support the current php version for the 3.4.x branch at least until phalcon 4 is stable IMHO. 
Many people might consider using/migrating to Phalcon 4 only when it gets stable and probably waiting for some more first Bugfix releases. Until this happens my personal guess is, there will already be  php 7.3.10+ around.
My boss would tell me to drop phalcon then (a new project to migrate to phalcon 4 someday had to be made anyway, so my boss would argue to switch back to a raw php framework if that's the only way to be able to use the latest php version. 

Don't get me wrong, I really like what you are doing and you are doing it great! Just think about it again 😉

Anyway, I read somewhere somebody got a working 3.4.2 windows DLL for php 7.3. So, even if phalcon is not officially supporting it, I would love if you could point me to a download link. Then my concerns might be completely gone and looking forward to a phalcon 4 upgrade can be planned in a more relaxed way. 
. I currently had no luck by simply googling it.. 🙂

_Originally posted by @lubber-de in https://github.com/phalcon/cphalcon/pull/13511#issuecomment-465325295_

sergeyklay:
cc @phalcon/framework-team

ruudboon:
Is it aready clear what needs to be done to support 7.3 under 3.4.2?

sergeyklay:
In general no. l'll research on that.

david-duncan:
Just wanted to chime in -

I'd like to see 3.x support 7.4 and 8.x

From what I can tell, the 4.x branch has a lot of unnecessary churn without providing anything new of value. I understand technical debt and the need to reduce it, but I don't see anything additive or worthwhile to spend the time migrating from 3.x to 4.x.

[...]

PR #13860, opened 2019-02-24
Provide the ability to build Phalcon using PHP 7.3

sergeyklay:
Provided PHP 7.3 support. See: https://github.com/phalcon/cphalcon/issues/13847


virgofx:
I think this was the right call to bring full support to 3.4.x. Thanks 👍

PR #13861, opened 2019-02-24
Remove all 'fresh' related code from the DI and DIInterface

dschissler:
Hello!

* Type: code quality
* Link to issue: This is a continuation of a larger effort to clean up the DI.  This related to a few recently merged PRs:

* [PR 13846](https://github.com/phalcon/cphalcon/pull/13846)
* [PR 13854](https://github.com/phalcon/cphalcon/pull/13854)

- [ ] This PR deletes a public function in the DI which is no longer used internally within Phalcon.  Its possible that some third party code could rely on this in a plugin but I seriously doubt it.  Hopefully this is just dead code YAY!  After the Dispatcher is refactored according to the commented notes within that class then the creation of some new tests will need to be evaluated.

Through the PR mentioned in the top portion of this PR I have set the stage for this "was fresh" code to be removed.  This also tightens up the DI interface.  After I or someone else refactors the Dispatcher base class then its likely that some new clever tests will need to be created.  For the moment this is just the next step in cleaning up the code quality and API.


sergeyklay (approved):

niden:
Thank you @dschissler

sergeyklay:
@niden Interface has been changed. I added the Breaks BC label. Please  mention this at the release log as well as documentation.

dschissler:
@sergeyklay @niden It should also be noted that the DI behaviour has now changed for `DI::get` when the service was defined as shared and it already exists in the shared instances cache.  Before it would fire the events each time but now it only fires the events once when creating the object (just like `getShared`).  I feel that its unlikely to affect people and that the behaviour is now much more correct and useful, but they should be notified even if its just to tute our own horn about improvements.

PR #13862, opened 2019-02-24
Pdo radicaloptions

dschissler:
Hello!

* Type: new feature + code quality
* Link to issue: [#13647](https://github.com/phalcon/cphalcon/issues/13647)

This PR is a more aggressive alternative to [PR 13841](https://github.com/phalcon/cphalcon/pull/13841)

- [x] Another PR exists for this but its from me.  This is a more radical change to PDO::connect.

I created another branch and PR to try a more aggressive refactor to go along with my default DSN changes.  I didn't want to screw up my other PR since I wasn't sure how this was going to turn out.  @niden perhaps this will address some of the comments from the other PR.


dschissler:
Oh, so now the root level of `descriptor` won't be processed.  You need to now feed it through `descriptor["options"]` or `descriptor["dsn"]`.

dschissler:
@niden @sergeyklay I think that this new very refactored code is superior.

As far as this constant sniffing code below goes, I think that I like it because it makes working with configuration through a json or yaml file a lot easier.  If I had to guess the original idea probably came from someone who was struggling with just that. 

```zephir
		for key, value in options {
			if typeof key == "string" {
				let pdoKey = "\PDO::" . key->upper();
				if defined(pdoKey) {
					let options[constant(pdoKey)] = value;
					unset options[key];
				}
			}
		}
```

Next up if this gets merged is to add something like a `postOptions` setting.  I'm not 1000% sure but I found and read that Sqlite doesn't honour the option errors as exceptions when its set in the PDO constructor.

So this zephir line is not honored.

```zephir
let options[\Pdo::ATTR_ERRMODE] = \Pdo::ERRMODE_EXCEPTION;
```

I had to fix it in my service definiion using the following:

```php
$db =  new Sqlite([
    'dbname'  => $filePath,
]);

$db->getInternalHandler()
    ->setAttribute(\PDO::ATTR_ERRMODE, \PDO::ERRMODE_EXCEPTION);
```

So I'm thinking about another PR after this one for `postOptions`.  I feel that this is a low risk addition and if my reading and real world findings are true then Phalcon users would benefit tremendously from having errors as exceptions in **reality**.

I think that its too ambitious of me to attempt to call sql calls for things that just can't be setup otherwise like charset in postgres, etc.  Maybe Phalcon 5.  It would be great if we could get a few of these in and to have a framework for adding another weird setting in the future if it comes up.

dschissler:
I changed my mind about the PDO constant sniffing being useful for config files.  Only the keys would be brought in but the values wouldn't.  So I deleted it.

CameronHall (approved):

niden (approved):

niden:
Thank you @dschissler

PR #13872, opened 2019-03-01
Service removeresolved

dschissler:
I removed `Service::isResolved()` and wrote a new test.  The first PR got put on master somehow even though I clicked it.

* Type: code quality
* Link to issue: None

Small description of change:

I don't think that `Service::isResolved()` was serving a purpose.  I'm on a mission to simplify the Service class.


niden:
Why is this removed again? Sorry but the _I don't think it serves a purpose_ is not enough of a rationale

dschissler:
@niden 

> Why is this removed again?

What do you mean?

> Sorry but the _I don't think it serves a purpose_ is not enough of a rationale

`isRevolved` implies as though calling `resolve()` would have some lasting impact on the state of the Service object, but in reality all it does is set a flag for itself.  If the service definition was set as shared then it does but otherwise it is not shared then there is no effect on internal state of the service.  If its shared then you can call `isShared()` or just `resolve()` to get the shared instance from cache.  So I propose that its a useless bit of state that isn't used internally.

`Service::isResolved` is used literally nowhere in the core of Phalcon and only in a pretty weak test, which is IMO testing for a nothing feature which serves no actual purpose.  If "this doesn't do anything" isn't valid assessment please answer the question "What does it do?".  If the DI doesn't need it and the Service itself doesn't actually need that flag for its internal operation then what kind of funky use case is there for it?  I propose that its just cruft that built up as the original designers were figuring things out.

I already took out an entire pointless variable in the Service constructor in a previous commit and this has me thinking that in actuality the entire DI and Service design was far from perfect.  This nasty funk goes all of the way into the Dispatcher and that has been addressed by virgofx, although nobody has fixed that part yet.  I plan on working on that over the next month.  I expect it to take many hours of thinking first.  In the meantime I'm trying to fix up this Service funk and I don't believe that anyone with competency + courage + time has attempted to fix the Holy Zephir Code.  The code is fast but its not great and its a hell of a bureaucratic struggle to fix even the most obvious weak designs (as far as my eye goes).

Issue #12261, opened 2016-09-28
Assets Manager hard reference to \Phalcon\Tag, should use DI


kristofferingemansson:
Assets Manager has hard reference to \Phalcon\Tag when generating HTML.
Assets Manager should fetch tag instance from dependency injector instead to allow custom behavior.

``` php
class MyTag extends \Phalcon\Tag
{
    public static function stylesheetLink($parameters = null, $local = true)
    {
        die('bar');
    }
}

$di = new \Phalcon\Di\FactoryDefault();
$di->setShared('tag', MyTag::class);

/** @var \Phalcon\Assets\Manager $manager */
$manager = $di->get('assets');
$manager->addCss('test.css');
$manager->outputCss();
```

Actual

``` php
<link rel="stylesheet" type="text/css" href="/test.css" />
```

Expected

``` php
bar
```

https://github.com/phalcon/cphalcon/blob/v3.0.1/phalcon/assets/manager.zep#L775
https://github.com/phalcon/cphalcon/blob/v3.0.1/phalcon/assets/manager.zep#L811 (same for JavaScript)
### Details
- Phalcon version: 3.0.1
- PHP Version: 5.5.24
- Operating System: Windows 7
- Installation type: xampp

Jurigag:
If it's 3.1.0 then if service called tag doesn't exist it should still use Tag class in static way.

niden:
Implemented

PR #13894, opened 2019-03-12
Fix: assets manager hard reference to phalcon\tag. Use DI

ekmst:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/12261

- [x] I updated the CHANGELOG

Thanks

### DIFF
--- a/phalcon/assets/manager.zep
+++ b/phalcon/assets/manager.zep
@@ -18,14 +18,18 @@ use Phalcon\Assets\Asset\Js as AssetJs;
 use Phalcon\Assets\Asset\Css as AssetCss;
 use Phalcon\Assets\Inline\Css as InlineCss;
 use Phalcon\Assets\Inline\Js as InlineJs;
+use Phalcon\DiInterface;
+use Phalcon\Di\InjectionAwareInterface;
 
 /**
  * Phalcon\Assets\Manager
  *
  * Manages collections of CSS/Javascript assets
  */
-class Manager
+class Manager implements InjectionAwareInterface
 {
+
+	protected _dependencyInjector;
 
 	/**
 	 * Options configure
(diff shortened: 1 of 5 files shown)


phalcon/assets/manager.zep
```diff
@@ -763,15 +783,23 @@ class Manager
...
 		} else {
 			let collection = this->get(collectionName);
 		}
 
-		return this->output(collection, ["Phalcon\\Tag", "stylesheetLink"], "css");
+		let callback = ["Phalcon\\Tag", "stylesheetLink"];
+
+		let dependencyInjector = this->_dependencyInjector;
+		if typeof dependencyInjector == "object" && dependencyInjector->has("tag") {
+			let tag = dependencyInjector->getShared("tag");
```
niden: Can we add a check to see if the `tag` was in the DI container and valid and throw an exception if need be?

phalcon/assets/manager.zep
```diff
@@ -795,15 +823,23 @@ class Manager
...
 		} else {
 			let collection = this->get(collectionName);
 		}
 
-		return this->output(collection, ["Phalcon\\Tag", "javascriptInclude"], "js");
+		let callback = ["Phalcon\\Tag", "javascriptInclude"];
+
+		let dependencyInjector = this->_dependencyInjector;
+		if typeof dependencyInjector == "object" && dependencyInjector->has("tag") {
+			let tag = dependencyInjector->getShared("tag");
```
niden: Same here

niden (dismissed):

phalcon/assets/manager.zep
sergeyklay: The `dependencyInjector->has("tag")` already here, on line 797. But why we need throw an exception? It will work w/o 'tag'

sergeyklay (approved):

phalcon/assets/manager.zep
niden: Never mind.

niden:
Thank you @ekmst

Issue #13893, opened 2019-03-12
Implement Registry with Collection


niden:
Change `Phalcon\Registry` to use `Phalcon\Collection` to remove code duplication.

Change `Phalcon\Session\Bag` to also use the collection object

niden:
Implemented

PR #13895, opened 2019-03-13
T13893 registry collection

niden:
Hello!

* Link to issue: #13893 

- [x] I updated the CHANGELOG

Refactored `Registry` to use `Collection`;
Refactored `Session\Bag` to use `Collection`

Thanks


niden:
cc @sergeyklay

phalcon/collection.zep
```diff
@@ -86,12 +91,16 @@ class Collection implements \ArrayAccess, \Countable, \IteratorAggregate, \JsonS
...
 	 * Get the element from the collection
 	 */
-	public function get(string! element, var defaultValue = null) -> var | bool
+	public function get(string! element, var defaultValue = null, bool insensitive = true) -> var | bool
 	{
 		var value;
 
-		if likely fetch value, this->data[element] {
-			return value;
+		if insensitive {
```
sergeyklay:
```suggestion
		if likely insensitive {
```

sergeyklay: You have to use the branch prediction here

phalcon/collection.zep
```diff
@@ -108,44 +117,25 @@ class Collection implements \ArrayAccess, \Countable, \IteratorAggregate, \JsonS
...
-	/**
-	 * Set an element in the collection
-	 */
-	public function set(string! element, var value) -> void
-	{
-		array data;
-
-		let data          = this->data,
-			data[element] = value;
+		if insensitive {
```
sergeyklay: the same

phalcon/collection.zep
```diff
@@ -108,44 +117,25 @@ class Collection implements \ArrayAccess, \Countable, \IteratorAggregate, \JsonS
...
 
-		this->init(data);
+		return isset this->headers[element];
 	}
 
 	/**
 	 * Initialize internal array
 	 */
-	public function init(array! data = [])
+	public function init(array! data = []) -> void
```
sergeyklay: How about this?
```suggestion
	internal function init(array! data = []) -> void
```

sergeyklay: why do we need this in the API part?

phalcon/collection.zep
```diff
@@ -24,13 +24,18 @@ class Collection implements \ArrayAccess, \Countable, \IteratorAggregate, \JsonS
...
 	 * @var array
 	 */
-	private data = [];
+	protected data = [];
 
-	public function __construct(array! data = null) -> void
+	/**
+	 * @var array
+	 */
+	protected headers = [];
```
sergeyklay: Is it a common class for any other collections, right? Why do we need domain specific headers in the base class? I thought that `data` is enough for data storage. Why do we need `data` and `headers` here?

sergeyklay: How about HeadersBag or something like this?

sergeyklay (changes requested):

phalcon/collection.zep
niden: I want to offer the option to populate the collection after it was created. For instance you can reuse the same object to populate data from another array that came in later in the code flow..

phalcon/collection.zep
niden: This is not the `headers` as you think of it. This array holds the case insensitive header -> case sensitive header map for faster searching and locating. The name of the internal array/property can be anything./ I just named it `headers`.

I will change it to arrayKeys or something.

sergeyklay (dismissed):
Looks good to me!

PR #13902, opened 2019-03-18
Pdo postoptions

dschissler:
Hello!

* Type: new feature + bug fix
* Link to issue: This is a continuation of #13647 

- [ ] CHANGELOG: I'm waiting for reviews.

I created an options and postOptions defaults for the PDO adapters.  I specifically did this now so that Sqlite can use errors as exceptions.  So Mysql and Postgres store this in the options, while Sqlite stores it in the post options.  Sqlite ignores this PDO constructor option and it must be set after instantiation.

Additionally this creates the infrastructure for doing more advanced auto-initialization in the future, for say Phalcon 5.  With this we will be able to address many more PDO weird cases so that Phalcon can deliver a painless out-of-the-box configuration.

If the code review goes over well then I will create a CHANGELOG entry as well as fixing the CHANGELOG for the alpha3 release to be more in line with what happened during that time (Added instead of Changed, missing entries).


sergeyklay (dismissed):

ruudboon:
Do we need an extra method for all the adapters? We could overload the connect for the sqlite and set the options there.

dschissler:
@ruudboon But `connect` is public.

dschissler:
@ruudboon Now every adapter except for sqlite will have `options` set with the `\Pdo::ATTR_ERRMODE : \Pdo::ERRMODE_EXCEPTION`.  Only sqlite uses it in post options instead since it is ignored when used in normal constructor options with that adapter.  We would need to overload every single adapter since there is options at least.  Then DSN as well for Postgresql because of charset.  Its kind of a mess.  In short, every single adapter needs a custom setup.  Of the three Phalcon supported adapters not one of them has the same setup.

I tried writing out some overloaded `connect` and I didn't find them to be satisfactory.  Do you have a suggestion?

Also be sure to check out my [Zephir issue 1835](https://github.com/phalcon/zephir/issues/1835).  Do you see that the Team allowed me to merge buggy code and that nobody seems to understand the real situation with `null`?

ruudboon:
@dschissler I think we could use the DSN options from the constructor as post options after connecting. That way we only need changes for the sqlite adapter connect method and every adapter could work the same way from a users point of view. 
Don't think the Zephir issue relates to this issue.

dschissler:
@ruudboon I don't know what you are referring to.  If you could show me some psuedo code then maybe.

>  Don't think the Zephir issue relates to this issue.

This is the second PR on the issue that I made for improving PDO.  In the first merged PR I removed the null sets in the else condition of the fetch.  This is invalid since even though it logs like a null, it is not in fact a null.  It is a very bad idea to be passing invalid values along that would never happen normally in PHP.  I pointed this out that the Team even approved my PR because basically nobody is understanding this very bad situation around null.  You guys didn't catch it, probably everybody except possible maybe Serghei have no idea about it.

niden:
I don't really agree with the direction of this PR. This adds a bit more overhead (more methods) where overloading the connect could do the same thing.

As for the options that one passes to their objects, that is the responsibility of the user and not the framework. It is one thing to pass say a string instead of an integer - that is something a framework should at least hint. But if one passes element `abcdef` in an option array that is their problem. The framework cannot predict everything and anything that can be passed in a constructor or a connect method.

@dschissler  On the comment about the PR's that have been merged etc. I have been asking again and again what is the vision and never got an answer. The last time I asked about it the said PR was closed and nothing was done. So if this PR is the end game then it is evident that had I known this in the first place I would have never merged the first ones.

As for the situation with nulls, that will be addressed when Serghei can address it.

dschissler:
@niden I have no idea why you are thinking that there is so much prediction happening here or even generalized checking.  There is the very specific checking for "charset" in postgres because apparently it does bad stuff or at least is highly misleading and so problematic.  Just look at the Dev Tools templates `di` service ugliness for an example of how that plays out in real (virtual) life.  For sqlite I allowed people to use `dbname` as well as `dsn` to set the file.  That is it and nothing else is planned as far as that goes.

Other than those two things above, the current default settings (of the adapter) for mysql, postgres and sqlite only setup the minimum required settings for the adapters to work properly as expected.  That means that mysql and postgres use `options` for providing errors as exceptions and sqlite uses `postOptions`.  **I could have left the errors as exceptions key in `connect` for `options` but it is currently irrelevant for sqlite and so I'd rather not feed in worthless values.  Furthermore I'm trying to set this up for Phalcon 5 when more research can be done as to if and how its possible to create a new standard so that the static `Mvc::config` does not need to be used differently for every DB to achieve ideal results.**  This will allow each adapter to have special defaults for `options`, `postOptions` and `dsn` to overcome any unintuitive behaviour or weirdness.

@niden Over the issue and two PRs I've now explained this at least twice and possibly more in other fragments.

@niden @ruudboon I've asked for some pseudo code example of how you would like to see the connect method working with the overloading.  I scratched out some mockups and it didn't seem to work out.  If you guys could bang out any kind of example then I can try that, or I'll get around to it in the next few days.

dschissler:
Another option would be to create a single `getDefaults` method which returns the default `dsn`, `options` and `postOptions`.  I don't think that performance would be so terrible to access the three array keys.

dschissler:
I see that the Travis build succeeded.  Provide any feedback and if its fine with some changes then I'll fix it up, comment and do the change log.

PR #13921, opened 2019-03-29
Minor fixes

niden:
Corrected the `json_encode` flags and renamed application


SidRoberts:
@niden Why did you downgrade Zephir?

PR #13924, opened 2019-03-31
Simplified events.

SidRoberts:
Hello!

* Type: code quality

- [x] I updated the CHANGELOG

Previously, the Events Manager was able to store events in two different systems - SplPriorityQueue and arrays. The system depended on whether priorities were enabled or not.

A lof of the code look similar to this:

```zephir
if this->enablePriorities {
    // Handle SplPriorityQueue
} else {
    // Handle array
}
```

It doesn't make sense to duplicate the code functionality for two systems when just one can be used instead.

Now SplPriorityQueue is used regardless. If priorities are disabled, the default priority (new constant DEAULT_PRIORITY) is used instead so that everything has the same priority.


CameronHall (approved):

niden (approved):

niden:
Thank you @SidRoberts

sergeyklay:
Thank you

PR #13928, opened 2019-04-01
Refactored Html\TagLocatorFactory::getEscaper() method.

SidRoberts:
Hello!

* Link to issue:

- [-] I updated the CHANGELOG

This code was previously buggy - it checked if the DI contained the 'escaper' service. If `true`, it made a new instance; if `false`, it attempted to retrieve the non-existent service.

### DIFF
--- a/phalcon/Html/TagLocatorFactory.zep
+++ b/phalcon/Html/TagLocatorFactory.zep
@@ -87,16 +87,14 @@ class TagLocatorFactory implements LocatorFactoryInterface
      */
     private function getEscaper() -> <EscaperInterface>
     {
-        var container, escaper;
+        var container;
 
         let container = Di::getDefault();
 
-        if (true !== container->has("escaper")) {
-            let escaper = container->getService("escaper");
-        } else {
-            let escaper = new Escaper();
+        if !container->has("escaper") {
+            return new Escaper();
         }
 
-        return escaper;
+        return container->getService("escaper");
     }
 }


Jeckerson (approved):

phalcon/Html/TagLocatorFactory.zep
```diff
@@ -87,16 +87,14 @@ class TagLocatorFactory implements LocatorFactoryInterface
...
-        if (true !== container->has("escaper")) {
-            let escaper = container->getService("escaper");
-        } else {
-            let escaper = new Escaper();
+        if !container->has("escaper") {
+            return new Escaper();
         }
 
-        return escaper;
+        return container->getService("escaper");
```
sergeyklay: Isn't a side effect this? Looks like now this method does different things. Maybe it should be renamed, e.g `getEscaperAndSetItToContainer()`?

sergeyklay (approved):

niden (approved):

niden:
@SidRoberts Do you think we should add this particular piece of code in some sort of a "global" area, even the `DI` itself? It seems that we are checking for services in several components and using pretty much the same piece of code everywhere.

The way I see it we can have two methods (something like this):

```zep
public function getServiceOrNew(var container, string service)
{
    // check if the service exists in the container and 
    // return it otherwise return a new one
}
```
and the question is should we also set the new service in the DI if that service does not exist?

and if we require the user to set services
```zep
public function getServiceAndThrowException(var container, string service)
{
    // check if the service exists in the container and 
    // throw exception if it does not exist
}
```
The names above are just for demonstration purposes

Thoughts?

SidRoberts:
@niden I like the idea but a bigger problem is how do we get from the service string to an object if the DI isn't aware of it?

For example:

```zephir
private function getEscaper() -> <EscaperInterface>
{
    var container;

    let container = Di::getDefault();

    return container->getOrCreateService("escaper");
}
```

If 'escaper' isn't in the DI, how does it know to return an `EscaperInterface`?

phalcon/Html/TagLocatorFactory.zep
```diff
@@ -87,16 +87,14 @@ class TagLocatorFactory implements LocatorFactoryInterface
...
-        var container, escaper;
+        var container;
 
         let container = Di::getDefault();
 
-        if (true !== container->has("escaper")) {
-            let escaper = container->getService("escaper");
-        } else {
-            let escaper = new Escaper();
+        if !container->has("escaper") {
```
sergeyklay: @SidRoberts It seems `Di::getDefault` can return null. What do you think about this?

SidRoberts: I don't like it. :stuck_out_tongue_closed_eyes: I didn't even realise.

Personally, I think it should behave like this: 

```zephir
public static function getDefault() -> <DiInterface>
{
    if self::_default === null {
        let self::_default = new self();
    }

    return self::_default;
}
```

sergeyklay:
As a far as I know, most popular and modern Service Locators works so:

1. I want to get a service (dependency)
2. I call some `resolve`/`make`/`get`/etc method
3. The Service Locator lookups into its own in memory storage for the cached service
4. The service will be instantiated (say `new`) in case of its absence

The same way work our `Di::get`. So, IMO, we can do something like this:
```zep
    private function getEscaper() -> <EscaperInterface>
    {
        var container;

        let container = Di::getDefault();

        if !container->has("escaper") {
            return container->getService(Escaper::class);
        }

        return container->getService("escaper");
}
```

phalcon/Html/TagLocatorFactory.zep
sergeyklay: :+1:

niden:
Thank you @SidRoberts . We can talk about the container in a separate conversation. For now this one is solid

PR #13950, opened 2019-04-05
Model methods that extend Model Manager functionality are now final

SidRoberts:
Hello!

* Type: bug fix

* Link to issue:

- [-] I updated the CHANGELOG

`Mvc\Model::getSource()` is now final and should be replaced by using `setSource()` in `initialize()`:

```php
class Post extends \Phalcon\Mvc\Model
{
    public function initialize()
    {
        $this->setSource("custom_posts_source");
    }
}
```

Overriding `getSource()` meant that `Model::getSource()` and `Model\Manager::getModelSource()` would return different things.


SidRoberts:
Will wait until https://github.com/phalcon/cphalcon/pull/13951 is merged and then will rebase and remove the revert commit.

niden:
@SidRoberts 13951 is merged.

Jeckerson (commented):
Can you change inside all `$this->setSource();` from double quotes `"` into singles `'` ?

tests/integration/Mvc/Model/Refactor-ManagerCest.php
```diff
@@ -64,6 +65,29 @@ public function testShouldReturnSourceWithPrefix(IntegrationTester $I)
...
+     * @since  2019-04-06
+     */
+    public function testModelSourceShouldBeTheSame(IntegrationTester $I)
+    {
+        $manager = new Manager();
+
+        $robots = new People(null, null, $manager);
+
+        $I->assertEquals(
+            "personas",
```
Jeckerson:
```suggestion
            'personas',
```

tests/integration/Mvc/Model/Refactor-ManagerCest.php
```diff
@@ -64,6 +65,29 @@ public function testShouldReturnSourceWithPrefix(IntegrationTester $I)
...
+            $robots->getSource()
+        );
+
+        $I->assertEquals(
+            "personas",
```
Jeckerson:
```suggestion
            'personas',
```

sergeyklay (dismissed):

david-duncan:
I think this MR exposes the amount of pain upgrading is going to take. 51 files changed for something that has been canon for using Phalcon's ORM for years now.

We may want to look into codemods for changes like this.

niden:
@david-duncan I was too a bit concerned but then looked at the files changed and in reality it was only one file (model.zep) that had the significant work. The rest of the files were the models defined in the testing suite.

Since we marked this as `Breaks BC` users should be aware that this will require additional changes. In reality for an existing application all one will have to do is use `setSource` instead of `getSource`. For instance in my models I use this

```php
function getSource(): string
{
    return 'robots';
}
```
so I will have to remove that and add this in the initialize
```php
$this->setSource('robots');
```

Now I agree this is a change, but it will be more of a time consuming change that will have to happen once vs something more serious. After all, once models are set, they don't change that much, other than adding a bit more functionality here and there.

niden (approved):

niden:
Thank you @SidRoberts

niden:
@CameronHall @Jeckerson Can we add an issue to the devtools so that this becomes part of the template generating models? Thanks

CameronHall:
Done @niden https://github.com/phalcon/phalcon-devtools/issues/1297

PR #13972, opened 2019-04-14
Tidied up Volt compilation logic.

SidRoberts:
Hello!

* Type: code quality
* Link to issue:

- [-] I updated the CHANGELOG

Instead of dealing with every option individually, we should first check whether the compiled file exists. If it doesn't, nothing else matters - we need to compile it regardless! :wink:


sergeyklay (approved):

sergeyklay:
Thank you

Issue #13964, opened 2019-04-09
[BUG] Updating a model fails after fetching its related records


zsilbi:
When fetching a record (`Robots`) and getting it's related models (`RobotParts`) a `RuntimeException` is being thrown when trying to save/update afterwards.

My observations:
- It only happens when after accessing `hasMany` or `hasManyToMany` type of relations.
- I have tried a lot of different models (names, aliases) and relations
- When creating a new model the error not occurs
- It doesn't matter if there are any related records in `RobotParts` table
- It only occurs when accessing the related records as property (saving after `$robot->getRobotParts()` works fine)

Output error when using hasMany

```php
RuntimeException
Call to undefined method \x01::writeattribute() 
```
Output error when using  hasManyToMany

```php
RuntimeException
Call to undefined method \x01::save() 
```

Script to reproduce the issue (only the relevant parts)
```php
class Robots extends \Phalcon\Mvc\Model {
    public function initialize() {
        $this->hasMany('id', 'RobotParts', 'robotId', [
            'alias' => 'robotParts'
        ]);
    }
}

public function main() {
    $robot = Robots::findFirst();
    $robot->robotParts;
    $robot->save(); // or $robot->update();
}
```

Current workaround:
```php
public function main() {
    $robot = Robots::findFirst();
    $robot->getRobotParts();
    $robot->save();
}
```

It's a widely used feature and it's very easy to reproduce so I think - and hope - that it must be some kind of incompability, otherwise it would have been discovered previously.

### Details
* Phalcon version: 4.0.0-alpha.3
* PHP Version: 7.2.16-1
* Operating System: Ubuntu 16.04
* Installation type: Compiling from source
* Zephir version (if any): Version 0.11.11-b661a58
* Server: nginx
* Other related info (Database, table schema): MariaDB 10.1.37

zsilbi:
I wrote tests to confirm the issue and it failed in CI with a rather strange error message.

` [PHPUnit\Framework\Exception] Invalid callback , no array or string given`

niden:
Resolved

niden:
Actually my bad, this was not resolved 100%. We need to do a lot more regression testing with the proposed fix from the PR attached to this issue.

niden:
Resolved in https://github.com/phalcon/cphalcon/pull/14035

PR #13616, opened 2018-11-24
Method to check if relationship was loaded for 4.0.x

CameronHall:
Hello!

* Type: new feature
* Link to -issue- existing pull request: https://github.com/phalcon/cphalcon/pull/12772

**In raising this pull request, I confirm the following (please check boxes):**

- [x] I have read and understood the [Contributing Guidelines](https://github.com/phalcon/cphalcon/blob/master/CONTRIBUTING.md)?
- [x] I have checked that another pull request for this purpose does not exist.
- [x] I wrote some tests for this PR.

Small description of change: This pull request adds a method to check if a relationship is loaded. It also changes the way we cache relationships. For _some_ reason we only stored `hasOne` and `belongsTo` relationships (those that return a model). Now we store resultset as well, which enables caching for `hasMany`, `hasManyToMany` etc.

Co-authored-by: Erik Wiesenthal <erikforo@hotmail.com>

Thanks

Issue #13531, opened 2018-10-16
Model relation reusable=>false not working


nilsm:
### Expected and Actual Behavior

I have a model with a relation set to 'reusable' => false.
I would expect the relation to be fetched from the database every time it is accessed.
However the relation is cached after the first call, even though 'reusable' => false.
Using 'getRelated()' to access the relation works as intended.

Note: 'reusable' => false is the default setting, so models do not work as intended by default.

Here is some sample code:

```php
class Robots extends \Phalcon\Mvc\Model
{
    public $id;
    public $name;
    public function initialize()
    {
        $this->hasMany("id", "Parts", "robot_id", ['reusable' => false]);
    }
}

class Parts extends \Phalcon\Mvc\Model
{
    public $id;
    public $robot_id;
    public $name;
    public static function find($parameters = NULL)
    {
    	echo 'find';
    	return parent::find($parameters);
    }
    public function initialize()
    {
        $this->belongsTo("robot_id", "Robots", "id", ['alias' => 'robot']);
    }
}

// actual code to test
$robot = new Robot();
$robot->name = 'Bob';
$robot->save();

$robot->parts; // first query - echos 'find'
$robot->parts; // 2nd query - no echo, even though it should re-query
$robot->getRelated('parts'); // echo 'find', as intended
$robot->getRelated('parts'); // echo 'find', as intended
```

### Severity
Critical - this is stopping us from upgrading to Phalcon 3 + model relations do not work as documented.

### Details

* Phalcon version: 3.4.1
* PHP Version: 5.6
* Operating System: n/a
* Installation type: Compiling from source
* Zephir version (if any): n/a
* Server: n/a
* Other related info

```
CREATE TABLE `robots` (
    `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
    `name` varchar(70) NOT NULL,
    PRIMARY KEY (`id`)
);
CREATE TABLE `parts` (
    `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
    `robot_id` int(10) unsigned NOT NULL,
    `name` varchar(70) NOT NULL,
    PRIMARY KEY (`id`),
    FOREIGN KEY (robot_id) references robots(id)
);
```

nilsm:
I've dug around in the source code and am pretty sure that this is the cause of the problem:

https://github.com/phalcon/cphalcon/pull/11983/files

Once accessed, the related records are stored as an instance property and returned every time, regardless of what 'reusable' is set to.

[...]

PR #12772, opened 2017-04-06
Method to check if relationship was loaded

Surt:
Hello!

* Type: new feature
* Link to issue:

**In raising this pull request, I confirm the following (please check boxes):**

- [x] I have read and understood the [Contributing Guidelines](https://github.com/phalcon/cphalcon/blob/master/CONTRIBUTING.md)?
- [x] I have checked that another pull request for this purpose does not exist.
- [ ] I wrote some tests for this PR.

Method to check if a relationship was loaded. So we can check it without invoking the lazy loading.

Thanks

PR #13985, opened 2019-04-15
Fixed saving a model after fetching its related records

zsilbi:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/13964

- [x] I updated the CHANGELOG

Small description of change:

I've added some tests to check if this issue exists at all.
The tests failed, so added a fix in a previous PR (https://github.com/phalcon/cphalcon/pull/13970) but ran into a conflict so I made a new PR.

The problem is, that the related models are stored in the `_related` storage inside Model after accessing via `__get()`. It is not stored when accessing related records via `__call()`.

When the model is saved,` _postSaveRelatedRecords` is being called and it iterates through `_related` and tries to save the previously fetched related records.

For now, I just removed adding these records to the `_related` as it's not mandatory when you are only reading them.

Thanks

### DIFF
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -31,6 +31,7 @@
 - Query Builder's `GROUP BY` field is now always an array. [#13962](https://github.com/phalcon/cphalcon/pull/13962)
 - Renamed `Phalcon\Paginator\Adapter::getPaginate()` to `paginate()` in documentation/tests (originally renamed in 4.0.0-alpha.1). [#13973](https://github.com/phalcon/cphalcon/pull/13973)
 - Fixed the exception message in `Phalcon\Security::computeHmac()` by removing `"%s"` from output.
+- Fixed `Mvc\Model::save()` throwing an exception after fetching its related records (many/many-to-many) via `__get()` [#13985](https://github.com/phalcon/cphalcon/pull/13985)
 
 ## Removed
 - Removed `arrayHelpers` property from the Volt compiler. [#13925](https://github.com/phalcon/cphalcon/pull/13925)
--- a/phalcon/Mvc/Model.zep
+++ b/phalcon/Mvc/Model.zep
@@ -288,8 +288,11 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
 
                 /**
                  * We store relationship objects in the related bag
+                 *
+                 * Removed temporarily.
+                 * See https://github.com/phalcon/cphalcon/issues/13964
                  */
-                let this->related[lowerProperty] = result;
+                // let this->related[lowerProperty] = result;
             }
 
             return result;
--- a/tests/integration/Mvc/Model/Refactor-ModelsRelationsMagicCest.php
+++ b/tests/integration/Mvc/Model/Refactor-ModelsRelationsMagicCest.php
@@ -85,7 +85,8 @@ class ModelsRelationsMagicCest
         $I->assertEquals($originalAlbum->id, $album->id);
     }
 
-    private function executeSaveAfterQueryRelated(IntegrationTester $I) {
+    private function executeSaveAfterQueryRelated(IntegrationTester $I)
+    {
         $album = Albums::findFirst();
         $artist = $album->artist;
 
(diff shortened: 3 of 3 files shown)


sergeyklay:
@zsilbi Could you please provide a small example of code which leads to an error you're fixes

sergeyklay:
I will keep this open for a some time to allow other contributors a chance to speak up, and to take a fresh look at it later. But at the moment it seems good to me.

/cc @phalcon/framework-team

sergeyklay (dismissed):

niden:
@zsilbi can you update the changelog when you get a chance please and rebase.

Thanks!

virgofx:
We need another test as it looks like this removed functionality.

Need to query a related record multiple times and ensure the database is only hit once. In this case, it will be re-queried every time when you remove that line.

zsilbi:
> We need another test as it looks like this removed functionality. I'm not sure with regards to the run-time exception or what was generating ... but this is likely not the right fix as this will break caching.
> 
> Need to query a related record multiple times and ensure the database is only hit once. In this case, it will be re-queried every time when you remove that line.

1. In `Model::__get()`, relations are obtained via `Model/Manager::getRelationByAlias()`. `Manager` caches them properly.
2. Queried related objects are stored in properties of the `Model`. This functionality in not affected by this PR.

niden (approved):

niden:
Thank you @zsilbi

virgofx:
Well now `isRelationshipLoaded` won't return correctly. If you're going to refactor the caching of related items ... it needs to be done everywhere. If we somehow double cached (manager + model) that needs to be more conveyed.

This needs to not be merged @niden.

zsilbi:
> Well now `isRelationshipLoaded` won't return correctly. If you're going to refactor the caching of related items ... it needs to be done everywhere. If we somehow double cached (manager + model) that needs to be more conveyed and fixed in all locations. Commenting out one line breaks other areas. As I mentioned ... this still will re-query ... so it's broken as you haven't tested for that anywhere.
> 
> This needs to not be merged @niden.
> 
> If a runtime exception is being thrown somewhere ... the fix is to not UNCOMMENT a line .. it's to fix the issue where the exception is being thrown and explain it properly with tests.

@virgofx it's a different issue then because if you query the items via `Model::__call()` (for example `$robot->getRobotParts()`) the relation is never added to the `Model::related` array anyway. 

If you investigate further you will notice that this related storage is only used before and after updating a models related records. Only if this storage is empty `_preSaveRelatedRecords` and `_postSaveRelatedRecords` will never run during an update.

sergeyklay:
Just reverted this PR from 4.0.x branch

sergeyklay:
@zsilbi Feel free to discuss the issue here and re-send a new one, again if the need arises

virgofx:
@zsilbi I'm not disagreeing that the original implementation is poor. If we're going to remove the related ... it should be removed everywhere and we should have tests to account for what I described above. It's still being used when setting related items and checking if it's loaded. So if we're going to remove the `related` attribute for handling that .... all 3 things need to be taking care of:

1 - getting related
2 - setting related
3 - checking related
4 - tests - all 3 above +
   please add: 
      1 - multiple repeat getting related -- instantiate a new SQL logger of the queries to show only the first one is set.

zsilbi:
Ok guys, I've done some research to get the complete picture about the purpose of `Model::related[]`

In #12772 @Surt mentoined that the former `Model::_related[]` (now `related[]`) storage is only used for belongs-to relations. Maybe that's why it must've failed when it was populated with other type of relations.

I am not sure if @CameronHall had been aware of this in #13616 when he'd added that line that I uncommented here. When `Model::isRelationshipLoaded()` was implemented, it only returned true when the related records were accessed as a property (`$robot->robotParts`). It passed the tests that time because they were only written to cover this case.

Also, this related storage problem needs to be redone together with issue #13531, as storing objects only in properties as a cache and using `isset()` will make the `reusable` option uneffective.

Please advise.

virgofx:
Like I said ... the related records is very *finicky* and any changes will either be super minor ... or a refactor and need to cover the cases I've mentioned above. We have to be very careful because I have made changes in extending models and small changes like this cause issues where repeat queries kill production servers .... from an end-users application logic ... they may want to use    `$myModel->getFancyRelated()->getValue()` in numerous places throughout code.

CameronHall:
I agree `isRelationshipLoaded` is a silly method. I only added it because I was cleaning up the PR backlog. In hindsight, we should only be able to rely upon `__isset` to handle this.

PR #14033, opened 2019-05-02
Http status codes constants

ekmst:
Hello!

* Type: new feature
* Link to issue:

- [x] I updated the CHANGELOG

Small description of change:

Thanks


sergeyklay:
Whenever a C program (C extension) is executed some part of memory is allocated in the RAM for the program execution. This memory is used for storing the frequently executed code (binary data), program variables etc. Below memory segments talks about same:

- Local variables (also called as [automatic variables](https://en.wikipedia.org/wiki/Automatic_variable) in C)
- [Global variables](https://en.wikipedia.org/wiki/Global_variable#C_and_C++)
- [Static variables](https://en.wikipedia.org/wiki/Static_variable)
- You can have global static or local static variables but above 3 are parent types

This feature request has a side effect: for each constant memory will be allocated in **each request startup**. Basically, in the low level, a constant is stored in the allocated memory and created after launching your application, it always has an address in RAM. _Even if you never use this constant_.

C !== PHP

Depends on HTTP stack: HTTP status codes are already constants. It is unlikely that they ever will be changed. Probably never, because this is industrial standard. For example:

```php
if ($response->getStatusCode() === 403) {
}
```

the same as:
```php
if ($response->getStatusCode() === \Phalcon\Http\ResponseInterface::STATUS_FORBIDDEN) {
}
```

These two examples are identical. In both cases, the values are constants. No one will change 403 for something else because it will break half of the internet. But in the first case, the code is more concise.

sergeyklay (changes requested):
See my comment

ekmst:
@sergeyklay 
Got it!
Thanks for clarifying!

Issue #14021, opened 2019-04-27
Phalcon ORM writeAttribute when value is an associative array


wajdijurry:
When using ```$model->writeAttribute('field', ['a'=>'apply', 'b'=>'banana'])``` to write an associative array to an attribute in the model, the result is strange.

### Expected and Actual Behavior
Expected result:
```
$model->field = ['a'=>'apply', 'b'=>'banana']
```

Actual result:
```
$model->a = 'apple';
$model->b = 'banana';
```

This behavior happening when trying to write a new attribute to a model, and the value must be an associative array.

### Details

* Phalcon version: (`3.4.2`)
* PHP Version: (`7.2.15`)
* Operating System: Ubuntu 18.10
* Installation type: Docker

zsilbi:
I also noticed this yesterday when I was working on other Model related issues.
I have a work-in-progress PR for this in the 4.0.x branch here: https://github.com/phalcon/cphalcon/pull/14018

zsilbi:
I did some testing and this bug is confirmed, but only occurs when `field` in not defined as a property in `Model` because the problem is in `Model::__set()`.

For example:
```php
class Users extends Model
{
    public $id;
    public $name;
}
```

```php
$user = new Users();
$user->writeAttribute('whatEverUndefinedProperty', ['id' => 123, 'name' => 'My Name']);
// or
$user->whatEverUndefinedProperty = ['id' => 123, 'name' => 'My Name'];

print_r($user->toArray());
```
results in:
```php
Array
(
   [id] => 123
   [name] => My Name
)
```

sergeyklay:
Actually it is important _feature_ and we can not just remove this behavior from 3.x branch because:
- This behavior is documented
- Many developers rely on this behavior
- It will break existing applications

zsilbi:
@sergeyklay, I couldn't find this anywhere in the docs

sergeyklay:
As a far as I know for 4.x this should be changed in https://github.com/phalcon/cphalcon/pull/13677

zsilbi:
> As a far as I know for 4.x this should be changed in #13677

That PR did not solve this, the code on line 359 causes this issue.
https://github.com/borisdelev/cphalcon/blob/7b6cfe807674147e5186d2680277706766d95ac7/phalcon/mvc/model.zep#L359

sergeyklay:
Well, @wajdijurry pointed current issue for 
> Phalcon version: (`3.4.2`)

[...]

PR #14035, opened 2019-05-02
Mvc\Model::__get() / __set() and related storage

zsilbi:
Hello!

* Type: bug fix
* Link to issues: https://github.com/phalcon/cphalcon/issues/13531, https://github.com/phalcon/cphalcon/pull/13616, https://github.com/phalcon/cphalcon/issues/13964, https://github.com/phalcon/cphalcon/issues/14021

- [x] I updated the CHANGELOG

Added new related storages
Modified __set(), __get(), getRelated() to use these related storages
Modified save() to use these related storages with transaction handling
Modified _getRelatedRecords() to use getRelated() when possible
Reusable relations are cached and returned from the Model directly
Non-reusable relations are not cached
Fixed __set() to handle arrays correctly

Added tests for Mvc\Model::save()
Added tests for Mvc\Model::__set()
Added tests for Mvc\Model::writeAttribute()
More tests needed...

Thanks,
zsilbi


tests/integration/Mvc/Model/UnderscoreSetCest.php
```diff
@@ -12,20 +12,301 @@
...
+    {
+        $I->wantToTest('Tests Mvc\Model - __set() with has-one related record');
+
+        $customer = new Models\Customers();
+        $customer->user = new Models\Users();
+
+        $user = $customer->user;
+
+        $I->assertInstanceOf(Models\Users::class, $user);
+        $I->assertEquals($customer->getDirtyState(), Model::DIRTY_STATE_TRANSIENT);
```
SidRoberts: Assertions are `($expected, $actual)` so it should be:

```php
$I->assertEquals(
    Model::DIRTY_STATE_TRANSIENT,
    $customer->getDirtyState()
);
```

And space out the assertions on multiple lines to make it easier to read. :wink:

tests/integration/Mvc/Model/UnderscoreSetCest.php
```diff
@@ -12,20 +12,301 @@
...
+
+        $robot = new Models\Robots();
+        $robot->robotsParts = [
+            new Models\RobotsParts(),
+            new Models\RobotsParts()
+        ];
+
+        $robotsParts = $robot->robotsParts;
+
+        $I->assertTrue(is_array($robotsParts));
```
SidRoberts: https://phpunit.readthedocs.io/en/7.4/assertions.html#assertinternaltype

tests/integration/Mvc/Model/UnderscoreSetCest.php
zsilbi: Thank you, @SidRoberts :)

tests/integration/Mvc/Model/UnderscoreSetCest.php
zsilbi: Resolved, thanks!

zsilbi:
Is caching related records in `Model` worth it?
`Manager` handles it well, and I don't think it has any performance benefits to do the same here. 

Making `isRelationshipLoaded()` work would require this cache, but I think `Manager` could be used for this kind of tasks, but I honestly don't see any real-world use case for this function from my perspective.

Also, records queried with any extra arguments, like
```php
$mechanicalParts = $robot->getRelated('parts', ['type = "mechanical"');
```
will only be cached in `Manager`.

CameronHall:
@zsilbi I'm a little naive to SplObjectStorage and observing when there's a use case for it, but would that serve as a better option than storing the relationship objects in an array? Totally chill if you think otherwise, I'm just curious.

zsilbi:
@CameronHall: I checked SplObjectStorage but I dont't think it would be beneficial to use here.
The related records are indexed by their aliases, not by their content (hash).
On the other hand, I think in `Manager`, if you have to store any data for a given object, it could work.

niden:
@zsilbi can you please rebase this?

This looks great. Awesome work!

niden:
Thank you @zsilbi

Issue #13825, opened 2019-02-13
[NFR] Create a view method to simple render template to a string.


sergeyklay:
As a regular user I would like to have something like this:

```php
$html = $view->toString($template, $params);
mail("someone@example.com", "My subject", $html);
```

without setting any additional params eg base path, layout, base template, etc.

dschissler:
Since when are you a normal user?

Perhaps a third parameter for using a file path to read with `file_get_contents`.

```php
$html = $view->toString($template, $params);
$html = $view->toString($templatePath, $params, true);
```

Jeckerson:
This is actually useful. Not only for email templates, but for ajax responses when you need to return rendered html inside json.

JABirchall:
Wait, doesn't view->getRender() already do this?

https://github.com/phalcon/cphalcon/blob/v3.4.0/phalcon/mvc/view.zep#L1118-L1165

ViltusVilks:
As I remember it is not so trivial. GetRender triggers events and can break something at the end. Partial can be used.

Otherwise only forking can do safe render.

Jurigag:
@JABirchall you still have here controllerName and actionName. Right now this can be achieved by some hacky way - like simply create temporary new volt template in /tmp and then render it, but this is just workaround.

niden:
Resolved in https://github.com/phalcon/cphalcon/pull/14058

EcLaboratory:
Watching the changes in code it seems the only real change is not to fire events. Is it planned to make available a method to do the action an the begining of this blog exposed?

I mean

$html = $view->toString($template, $params);

where $template is an string with the volt template content and $params an array to use to render the template

Thanks in advance

niden:
@EcLaboratory This particular change is for the `View` which will take a template as a controller/action/view and will load it and send it back.

If you want a specific template to be parsed and returned, you can use the `View\Simple` component which does the same thing.

PR #14058, opened 2019-05-10
T13825 view tostring

niden:
Hello!

* Type: new feature
* Link to issue: #13825 

- [x] I updated the CHANGELOG

Added `Phalcon\Mvc\View::toString()` to return the rendered view as a string

Thanks


niden:
cc @sergeyklay

niden:
Made a mistake when creating the branch hence the extra files. I will sort out the ext folder in 4.x branch

SidRoberts:
How about adding '__toString()' as well?

niden:
@SidRoberts not sure if we can do that since we need some parameters for the `toString()`. From what I know `__toString()` does not accept any, unless I am mistaken.

sergeyklay (approved):

sergeyklay:
@SidRoberts I don't support the idea of adding `__toString` method. First of all `__toString` is not designed for humans. It acts as a magic method because for the most part it is don't get called explicitly but rather intercept operations. Unfortunately there are limits to its magic, specifically the only "context" the method is aware of is its design contract: to return a string. But its not clear what purpose that is. Should this be for some internal debugging or logging purposes? Secondly, you cannot throw an exception from within a `__toString` method. Doing so will result in a _fatal error_. It is by design. If you give users today `__toString`, then tomorrow they will start using it in situations where exceptions will theoretically occur. So it will yet another ability to get fatal error using Phalcon. Let's use principle of least astonishment as long as we can.

PR #14076, opened 2019-05-14
Tidied up Model::cloneResultmap() and ::cloneResultmapHydrate()

SidRoberts:
Hello!

* Link to issue:

- [x] I updated the CHANGELOG

Small description of change:

Thanks

### DIFF
--- a/phalcon/Mvc/Model.zep
+++ b/phalcon/Mvc/Model.zep
@@ -738,72 +738,72 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
         instance->setDirtyState(dirtyState);
 
         for key, value in data {
-            if typeof key == "string" {
-                // Only string keys in the data are valid
-                if typeof columnMap != "array" {
-                    let instance->{key} = value;
-
-                    continue;
-                }
-
-                // Every field must be part of the column map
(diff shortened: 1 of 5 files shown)


phalcon/Mvc/Model.zep
```diff
@@ -888,25 +885,17 @@ abstract class Model implements EntityInterface, ModelInterface, ResultInterface
...
         }
 
-        if hydrationMode == Resultset::HYDRATE_ARRAYS {
-            return hydrateArray;
+        if hydrationMode != Resultset::HYDRATE_ARRAYS {
+            return Arr::arrayToObject(hydrateArray);
         }
 
-        return hydrateObject;
+        return hydrateArray;
```
niden: This is actually slower than setting the object directly. With the current implementation we create a `stdClass` object and set its properties and return it in one loop while this implementation does two loops, one for the array and one to make it an object.

niden (changes requested):

niden (approved):

niden:
Thank you @SidRoberts

phalcon/Mvc/Model.zep
SidRoberts: Slightly slower but significantly less complicated.

PR #14077, opened 2019-05-15
Loading a Module (either MVC or CLI) now throws an exception properly

SidRoberts:
Hello!

* Type: bug fix
* Link to issue:

- [x] I updated the CHANGELOG

Previously, paths that were bad didn't matter if the class was already loaded. We can't assume that the class will already have been loaded or will be available by the regular loader.


niden (dismissed):

niden:
Thank you @SidRoberts

sergeyklay (dismissed):

Jeckerson (dismissed):

PR #14096, opened 2019-05-20
Tidied up tests.

SidRoberts:
Hello!

* Type: code quality
* Link to issue:

Small description of change:

Thanks


tests/integration/Mvc/Model/Behavior/NotifyCest.php
```diff
@@ -1,33 +0,0 @@
-<?php
```
niden: Why are these deleted?

tests/unit/Mvc/Collection/AggregateCest.php
```diff
@@ -1,31 +0,0 @@
-<?php
```
niden: Are these deleted in anticipation of the new mongo collection object?

tests/unit/Mvc/Dispatcher/CallActionMethodCest.php
```diff
@@ -1,31 +0,0 @@
-<?php
```
niden: Same why are these deleted?

niden (changes requested):

tests/integration/Mvc/Model/Behavior/NotifyCest.php
SidRoberts: It's an abstract class so we can't test it directly and it is covered by the subclasses SoftDelete and Timestampable anyway.

tests/unit/Mvc/Collection/AggregateCest.php
SidRoberts: All of the cests deleted from unit/Mvc/ also exist in integration/. As they require the DI so they should be in integration/ so these have been deleted.

tests/unit/Mvc/Dispatcher/CallActionMethodCest.php
SidRoberts: See Collection comment.

niden (approved):

niden:
Thank you @SidRoberts

Issue #14123, opened 2019-05-24
[NFR] Http\Payload 


niden:
Introduce `Http\Payload` class to handle different payloads sent to different parts of the application or even used as a return object for the view.

niden:
Resolved in https://github.com/phalcon/cphalcon/pull/14124

PR #14124, opened 2019-05-25
T13123 http payload

niden:
Hello!

* Type: new feature
* Link to issue: #14123 

Added `Phalcon\Http\Payload` to help in transferring messages between layers of the application. 
Added `Phalcon\Http\PayloadFactory`: Factory to create payload objects.

Thanks


niden:
cc @sergeyklay

sergeyklay (commented):
Looks awesome but I'm not sure about constants in the `Phalcon\Http\Payload\Status` class. Please take a look at https://github.com/phalcon/cphalcon/pull/14033#issuecomment-488651127

niden:
@sergeyklay Those constants need to be there because they are going to be used later on to work with headers and pointing to what has happened with the payload from the input to the domain to the response. So for instance the `ACCEPTED` or `AUTHENTICATED` tells the domain what happened with that response/payload

This is preparation for the future. I really don't know how better way to to this, if say I use different structure, i.e. numeric codes, Constants will still need to be there in any case.

You could say that we can leave that as a string and the developer will have to pass "ACCEPTED" in the code but that will not make things easy for the developers (syntax errors etc.) and we will have to check a lot of values in many `if` statements in the code to understand what was passed.

Issue #14126, opened 2019-05-25
[NFR] DB - Data Mapper Implementation


niden:
Introduce the Data Mapper pattern for the ORM

https://martinfowler.com/eaaCatalog/dataMapper.html

Related: https://github.com/phalcon/cphalcon/pull/13022

gnumoksha:
Do you have a roadmap for that? I can help with the planning and code.

At the moment I'm creating a Proof of Concept using the current Phalcon Active Record Models to manage a Plain Old PHP Object (POPO). It is very simple but can be a good start point.

niden:
@gnumoksha This will be in the 4.1 release maybe later depending on how much work we have. We might even release it in stages - not fully functional. For instance we can start by releasing the Table/Row classes first, then the mapper and then tie everything together. Not 100% sure since I have not started it.

Any help will be more than welcome.

PR #14167, opened 2019-06-07
Data Mapper

SidRoberts:
Hello!

* Type: new feature
* Link to issue: https://github.com/phalcon/cphalcon/issues/14126, https://github.com/phalcon/cphalcon/pull/13022

My plan:

1. Add the necessary functions to the Models Manager which simply call the method on the Model class.
2. Write / copy tests to prove it works.
3. Move the code over to the new Models Manager method and have the Model call the new Manager method instead.
4. Check all the tests still work.

All existing methods will continue to function as they do now to preserve BC. We can then consider removing them in version 5 whenever that may be. I'm expecting that, with this way, any BC-breaking work will be kept to a minimum when we start working on v5. Changes to interfaces will obviously have to wait until v5.

I'm planning to keep the method names the same as they are now but I'm open to comments.

### DIFF
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -12,6 +12,7 @@
 - Added `Phalcon\Http\PayloadFactory`: Factory to create payload objects. [#14123](https://github.com/phalcon/cphalcon/issues/14123)
 - Added `Phalcon\Http\Message\ServerRequestFactory::load`: Method to create a `ServerRequest` object from globals. [#14154](https://github.com/phalcon/cphalcon/pull/14154)
 - Added `Phalcon\Collection\ReadCollection`: Read only collection. [#14154](https://github.com/phalcon/cphalcon/pull/14154)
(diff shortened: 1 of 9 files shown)


niden:
This has to be a separate namespace not the same one. We cannot remove the current implementation it will really break everything for everyone. We need to have both implementations in the framework for one version at least to ensure that there is ample time for developers to adjust their applications.

There is a lot of spaghetti code in the framework that we need to clean up especially with the ORM and it is kind of a blessing to steer to the Data Mapper pattern because we can write the ORM from scratch.

Finally this is a 4.1+ - not releasing it with 4.0.

We can talk more about this @SidRoberts if you wish on Discord

SidRoberts:
I'll close this for now and reopen it when we have a 4.1.x branch. I'm going to use `Phalcon\Mvc\DataMapper` for the manager class and `Phalcon\Mvc\DataMapper\Model` if this is OK? I'm also going to start by copying classes and tests we have now and then massively modifying them as writing it from scratch may be too much work.

Issue #13768, opened 2019-01-17
Invalid logic of Phalcon\Config::merge()


zatrepalek:
Hi, I found this bug. Probably related to https://github.com/phalcon/cphalcon/issues/13201 (stale - closed).

### Expected and Actual Behavior

**If an array is passed to merge() method string (number-like) keys are not merged correctly**

Correct behaviour (for string-like keys):
```php
$params1 = ['a' => 'aaa', 'b' => ['bar' => 'rrr', 'baz' => 'zzz']];
$config = new \Phalcon\Config($params1);
$config->toArray()
// output
array (
  'a' => 'aaa',
  'b' => 
  array (
    'bar' => 'rrr',
    'baz' => 'zzz',
  ),
)

$params2 = ['c' => 'ccc', 'b' => ['baz' => 'xxx']];
$config->merge(new \Phalcon\Config($params2));
$config->toArray()
// correct merge
array (
  'a' => 'aaa',
  'b' => 
  array (
    'bar' => 'rrr',
    'baz' => 'xxx',
  ),
  'c' => 'ccc',
)
```

Wrong behaviour (for integer-like keys):
```php
$params1 = ['1' => 'aaa', '2' => ['11' => 'rrr', '12' => 'zzz']];
$config = new \Phalcon\Config($params1);
$config->toArray()
// output (string keys casted to integers)
array (
  1 => 'aaa',
  2 => 
  array (
    11 => 'rrr',
    12 => 'zzz',
  ),
)

$params2 = ['3' => 'ccc', '2' => ['12' => 'xxx']];
$config->merge(new \Phalcon\Config($params2));
$config->toArray()
// wrong merge (casted to integers and invalid merge logic for sub array)
array (
  1 => 'aaa',
  2 => 
  array (
    11 => 'rrr',
    12 => 'zzz',
    2 => 'xxx',
  ),
  3 => 'ccc',
)
```

### Details

* Phalcon version: 3.4.2
* PHP Version: 7.1.25
* Operating System: macOS 10.13.6 (host), Debian 9 (docker)
* Installation type: Compiling from source (build/php7/64bits)
* Zephir version: (Version 0.10.14-975ad02db4)
* Server: Nginx

TimurFlush:
It seems Phalcon Team wanted the behavior of merging mechanism as in the function array_merge().

You are must see http://php.net/manual/en/function.array-merge.php:

[...]

Issue #13201, opened 2017-12-08
Inconsistent Phalcon\Config::merge()


pentagonal:
Inconsistent config merge 

On some case of `Phalcon\Config::merge` that contain numeric (string/float/double) values possible lack of key name of config.

Because `Phalcon\Config` convert all keys into string, and when config doing merge process it will be validate the keys as numeric value not an integer (`string -> integer`).

Because on some case, config has been used to handle dynamic array values, eg: versioning.
Remove the `is_numeric` logic to `is_int` maybe better to make it more relevance.

## CASE (Logic numeric)

https://github.com/phalcon/cphalcon/blob/a0560cb3174b37d3b45a46986f5ac7aa1c7ffd81/phalcon/config.zep#L325

> Code Execution

```php
<?php
    use Phalcon\Config;
    $config        = new Config(['0.4' => 0.4]);
    $configToMerge = new Config(['1.1' => 1, '1.2' => 1.2, '2.8610229492188E-6' => 3]);
    $config->merge($configToMerge);
```
> Result

```text
Phalcon\Config Object
(
    [0.4] => 0.4
    [1] => 1
    [2] => 1.2
    [3] => 3
)

```

## CASE (arrayaccess increments)  -> offsetSet

And when config set as the increment also does not follow the `array` rules on method `offsetSet`

https://github.com/phalcon/cphalcon/blob/a0560cb3174b37d3b45a46986f5ac7aa1c7ffd81/phalcon/config.zep#L172

> Code

```php

$config   = new Config();
$config[] =  'a';
```
> Result
```text
Phalcon\Config Object
(
    [] => a
)
```

The increment if produce array

> Idea for increment key values (example code) -> like array reproduce of increment

```php
/**
 * php code start for increment
 *
 * @var \Phalcon\Config $this
 */
$array = get_object_vars($this);
// manipulate array this will be reproduce increment values
$array[] = true;
// move pointer into last array
end($array);
// real last increment
$increment = key($array);


```

the `offsetSet(index, value)` the index will be contains key as `null` if the increment set as empty array increment

niden:
Resolved in https://github.com/phalcon/cphalcon/pull/14186

PR #14186, opened 2019-06-18
Fixed Phalcon\Config::merge being inconsistent

CameronHall:
Hello!

* Type: bug fix
* Link to issue: #13201, #13768

Small description of change: 
- `Phalcon/Config` now extends `Phalcon\Collection\Collection`
- Removed __set_state method
- Merging is done properly
- Assignment works as expected _all_ the time

It is worth noting that in #13768 it talks about correctly interpreting integers defined as strings. This is simply not possible given that PHP will cast strings that look like integers to integers. 

We also need to maintain our own merge functionality because the native PHP `array_merge_recusive` will merge values together into an array if they have the same string keys. `array_replace_recursive` created its own issues whereby it would replace keys with the same numeric index instead of appending to array itself.

Thanks


phalcon/Config.zep
```diff
@@ -39,7 +39,7 @@ use Phalcon\Config\Exception;
  * );
  *</code>
  */
```
niden: Can we change `pathDelimiter` to a non static property (along with relevant calls and methods). No need to have it like that. `setPathDelimiter` can return self also so that will help with chaining

niden (changes requested):

niden (dismissed):

niden (dismissed):

niden:
This is solid. Will wait for @sergeyklay to give us any ideas on why the windows build fails. If something is wrong and we cannot fix it immediately we can always extend the `Collection` class vs. the `ArrayObject`

sergeyklay (dismissed):
Looks good to me

niden (dismissed):

niden:
Thank you @CameronHall

PR #14187, opened 2019-06-19
Updated Stale to stop closing NFR and Enhancements

CameronHall:
Hello,

I'd like to stop stale marking NFR or Enhancements as stale... I don't think we should be relying upon ticket activity to identify whether an NFR is worthwhile.


niden (approved):

niden:
Thank you @CameronHall

PR #14189, opened 2019-06-19
Simplified .travis.yml

SidRoberts:
Hello!

* Type: code quality
* Link to issue:

I'm having a few issues with setting up Travis on my account (despite it previously working fine). I'm trying to simplify the setup process to make it easier to debug. Everything is pretty much the same as before. ~~The only difference is that I have removed the shell script that installs MySQL 5.7 as it is already installed by Travis.~~


.travis.yml
```diff
@@ -63,25 +58,15 @@ before_install:
 
     git config --global advice.detachedHead false
 
-    # Export tests environment variables
```
sergeyklay: Could we move this to `.ci` directory which is at the project root. I plan to migrate all provision/build files from `tests/_ci` to `.ci`

.travis.yml
```diff
@@ -49,11 +49,6 @@ env:
 
 before_install:
   - |
-    # General settings
-    set -e
-    stty cols 120
-    ulimit -c unlimited -S || true
```
sergeyklay: Could you return back 
```
ulimit -c unlimited -S || true
```

and
```
set -e
```

This first one is needed to generate core dumps, and `set -e`  to exit immediately if a command exits with a non-zero status.

.travis.yml
```diff
@@ -63,25 +58,15 @@ before_install:
...
+  - tests/_ci/export-variables.sh
 
 install:
-  - .ci/install-mysql-5.7.sh
   - tests/_ci/setup-dbs.sh
   - tests/_ci/pear-setup.sh
-  - .ci/install-prereqs.sh
   - tests/_ci/install-re2c.sh
+  - tests/_ci/install-zephir.sh
+  - tests/_ci/install-php-extensions.sh
```
sergeyklay: I wold like avoid using `tests/_ci` directory. Could you move scripts to `.ci` instead

.travis.yml
```diff
@@ -63,25 +58,15 @@ before_install:
...
-    # Export tests environment variables
-    source tests/_ci/environment
-    export $(cut -d= -f1 tests/_ci/environment)
-
-    # Export build environment variables
-    export PHP_MAJOR="$(php -r 'echo phpversion();' | cut -d '.' -f 1)"
-    export PHP_MINOR="$(php -r 'echo phpversion();' | cut -d '.' -f 2)"
-    export PHP_VERNUM="$(php-config --vernum)"
-    export PHP_PEAR_PHP_BIN=$(phpenv which php)
+  - tests/_ci/export-variables.sh
```
sergeyklay: You're missed `source` keyword:

```suggestion
  - source tests/_ci/export-variables.sh
```

Previously we used export in current shell. Now you're use subshell thus you have to _source_ this file.

tests/_ci/install-php-extensions.sh
```diff
@@ -73,10 +54,3 @@ if [[ "$(php -m | grep redis | wc -l)" = "0" ]] && [[ -f "${redis_ext}" ]];
 then
 	echo 'extension="redis.so"' > "$(phpenv root)/versions/$(phpenv version-name)/etc/conf.d/redis.ini"
 fi
-
-# Local variables:
-# tab-width: 4
-# c-basic-offset: 4
-# End:
-# vim600: noet sw=4 ts=4
-# vim<600: noet sw=4 ts=4
```
sergeyklay: Please return this back.

sergeyklay (changes requested):

.ci/install-php-extensions.sh
```diff
@@ -74,8 +55,6 @@ then
 	echo 'extension="redis.so"' > "$(phpenv root)/versions/$(phpenv version-name)/etc/conf.d/redis.ini"
 fi
 
-# Local variables:
-# tab-width: 4
```
sergeyklay: Return please this back

.ci/install-zephir.sh
```diff
@@ -0,0 +1,27 @@
...
+$(phpenv which phpize)
+./configure --silent --with-php-config=$(phpenv which php-config) --enable-zephir_parser
+make --silent -j"$(getconf _NPROCESSORS_ONLN)"
+make --silent install
+echo 'extension="zephir_parser.so"' > "$(phpenv root)/versions/$(phpenv version-name)/etc/conf.d/zephir_parser.ini"
+
+wget --no-clobber -O $HOME/bin/zephir https://github.com/phalcon/zephir/releases/download/${ZEPHIR_VERSION}/zephir.phar
+chmod +x $HOME/bin/zephir
+
+# c-basic-offset: 4
```
sergeyklay: Missed 
```
# Local variables:
# tab-width: 4
```

.travis.yml
```diff
@@ -49,9 +49,7 @@ env:
 
 before_install:
   - |
-    # General settings
     set -e
-    stty cols 120
```
sergeyklay: Well, `set -e` added by mistake. Could you remove this line too ?

sergeyklay (changes requested):

SidRoberts:
It turns out that, despite testing the same commits, phalcon's Travis uses Ubuntu 14.04 and my Travis uses Ubuntu 16.04 :face_with_head_bandage: I've explicitly set the distribution to trusty (14.04) to avoid this.

sergeyklay (approved):

sergeyklay:
Thank you

PR #14194, opened 2019-06-21
Added PHP 7.4 to Travis as allowed failure

CameronHall:
Hello!

I've updated travis to test the nightly version so we can see how we perform on upcoming PHP versions (e.g. PHP 7.4)

Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,9 @@ sudo: false
 
 language: php
 php:
-  # PHP 7.4 is not ready yet
-  # - 'master'
-  - '7.3'
-  - '7.2'
+  - nightly
+  - 7.3
+  - 7.2
 
 dist: trusty
 
@@ -21,9 +20,8 @@ addons:
 
 matrix:
   fast_finish: true
-  # PHP 7.4 is not ready yet
-  # allow_failures:
-  #  - php: 'master'
+  allow_failures:
+    - php: nightly
 
 cache:
   apt: true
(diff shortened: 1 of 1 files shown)


CameronHall:
> This package requires php >=7.2 <8.0 but your PHP version (8.0.0-dev) does not satisfy that requirement.

Please hold.

.travis.yml
```diff
@@ -2,10 +2,10 @@ sudo: false
 
 language: php
 php:
-  # PHP 7.4 is not ready yet
-  # - 'master'
-  - '7.3'
-  - '7.2'
+  - nightly
```
sergeyklay: It is 8.0 actually. I'm not sure that we need it here in the next half year

sergeyklay (changes requested):
Also disable REPORT_COVERAGE for PHP 7.4

.travis.yml
CameronHall: Yeah I agree. I'll remove that one as it's dependent on Zephir anyway.

sergeyklay:
@CameronHall AppVeyor supports skip directive in commit message like `[skip appveyor]`. 
For more see https://www.appveyor.com/docs/how-to/filtering-commits/#skip-directive-in-commit-message

CameronHall:
> Also disable REPORT_COVERAGE for PHP 7.4

How do I do that?

sergeyklay:
```yml
before_install:
  - |

    # ...
    # ...
    # ...

    if [ "$(php-config --vernum)" -ge "70400" ]; then
      export REPORT_COVERAGE=0
    fi
```

CameronHall:
@sergeyklay all done

sergeyklay:
Thank you

PR #14207, opened 2019-06-25
Update form.zep

scrnjakovic:
Form shouldn't create object variables, it should write to entity only if the property exists

* Type: bug fix


sergeyklay:
The `3.4.x` branch is not longer supported.

ruudboon:
I tried integrating this in V4 but it will break too much a requires some discussion first. 
@scrnjakovic I will close this now. If you still would like to see this happening could you open an issue and describe to problems you're having now?

PR #14215, opened 2019-06-26
Remove HAVE_SPL

petk:
Hello!

* Type: bug fix
* Link to issue: /

Small description of change:

The HAVE_SPL symbol is defined in PHP to indicate the presence of the spl extension. Since PHP 5.3 the SPL extension is always availabe and since PHP-7.4 the HAVE_SPL symbol has also been removed.

Thanks

### DIFF
--- a/ext/kernel/main.c
+++ b/ext/kernel/main.c
@@ -106,9 +106,7 @@ void zephir_fast_count(zval *result, zval *value TSRMLS_DC) {
 
 	if (Z_TYPE_P(value) == IS_OBJECT) {
 
-		#ifdef HAVE_SPL
 		zval *retval = NULL;
-		#endif
 
 		if (Z_OBJ_HT_P(value)->count_elements) {
 			ZVAL_LONG(result, 1);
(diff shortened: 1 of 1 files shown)


niden:
@petk Thank you for this but I will have to decline it. Here is why

These files are automatically generated when we run `zephir build` As such, these changes will be overwritten the next time we release a version and thus generate the build

However you can send the same PR in the Zephir repository and that will propagate here also.

Have a look at this:

https://github.com/phalcon/zephir/search?q=HAVE_SPL&unscoped_q=HAVE_SPL

These are the files/templates that the `HAVE_SPL` is defined.

Thanks for this !!

petk:
Got it.. Ok. more coming up tomorrow then.

sergeyklay:
Done:
```
$ ag --cc HAVE_SPL ext/ | wc -l
0
$ ag --cc HAVE_SPL build/ | wc -l
0
```

Thank you for the report, and for helping us make Phalcon better!

PR #14227, opened 2019-07-01
Refactored the storage adapters

ekmst:
Hello!

* Type: code quality
* Link to issue:

Small description of change:
Refactored the storage adapters

Thanks

### DIFF
--- a/phalcon/Storage/Adapter/Redis.zep
+++ b/phalcon/Storage/Adapter/Redis.zep
@@ -17,8 +17,6 @@ use Phalcon\Storage\SerializerFactory;
 use Phalcon\Storage\Serializer\SerializerInterface;
 
 /**
- * Phalcon\Storage\Adapter\Redis
- *
  * Redis adapter
  */
 class Redis extends AbstractAdapter
--- a/phalcon/Storage/Adapter/Stream.zep
+++ b/phalcon/Storage/Adapter/Stream.zep
@@ -10,12 +10,16 @@
 
 namespace Phalcon\Storage\Adapter;
 
+use FilesystemIterator;
+use Iterator;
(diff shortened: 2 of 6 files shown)


phalcon/Storage/Adapter/Redis.zep
```diff
@@ -15,10 +15,9 @@ use Phalcon\Storage\Adapter\AbstractAdapter;
 use Phalcon\Storage\Exception;
 use Phalcon\Storage\SerializerFactory;
 use Phalcon\Storage\Serializer\SerializerInterface;
+use Redis;
```
niden: I don't think this will work since the actual class is called Redis. Perhaps add an alias here? This was the reason that I left the `\` in the connection class below.

phalcon/Storage/Adapter/Stream.zep
```diff
@@ -290,6 +295,17 @@ class Stream extends AbstractAdapter
         return this->getDir(key) . str_replace(this->prefix, "", key, 1);
     }
 
+    private function getIterator(string! dir) -> <Iterator>
```
niden: I am guessing this is working now with 0.12.0 Zephir. It was not working for the earlier version hence the `rglob` implementation.

niden (changes requested):

phalcon/Storage/Adapter/Stream.zep
ekmst: Offer to return `rglob`?

niden: No no, If this implementation works leave it. It is more modern and cleaner :)

niden:
Thank you @ekmst

Issue #13697, opened 2018-12-24
Remove Mongo from Phalcon


niden:
Since there is no support for the existing driver for PHP, we have to remove Mongo from this release.

There is a plan to re-introduce it in the future as a storage adapter and with that we can re-introduce the `Mvc\Collection` class or something similar

ViltusVilks:
What do You mean by removing Mongo? Remove direct "usage" of mongo in .zep and also remove ODM completely in 4.0 release?

niden:
@ViltusVilks Remove the current driver (Mongo driver) that Phalcon uses to interact with Mongo because it has been deprecated. That will mean that the ODM will not work for a period of time until the new driver is introduced.

If we can do this in one go that would be great.

ViltusVilks:
Huh, deprecated? What is this? https://pecl.php.net/package/mongodb
Or I do not understand what is deprecated (Phalcon support for ODM/Mongo compatibility, or Mongo decided to drop support for MongoDB native driver (NETLIB)?)

I am working on dokerized Phalcon-ODM-only project to make slim, fast and scalable image. Without ODM project is dead :(

niden:
@ViltusVilks It is the driver itself. Phalcon has been using this driver:

http://php.net/manual/en/book.mongo.php

and we need to change it to this driver

http://php.net/manual/en/set.mongodb.php

niden:
It is a matter of time and resources really. We have not "removed" anything with regards to the Mongo functionality. If you have the first driver installed then it will work I think. I really don't know the changes that 7.2 brought in that aspect of PHP.

Removing mongo functionality including the ODM in one release and reintroducing it properly tested and upgraded is the plan. It has not materialized yet since we have not finished the release - still in alpha

ViltusVilks:
Ah... this is old one driver! 
Just for information - I am using **[incubator](https://github.com/phalcon/incubator)** github project, which overrides collection and uses new driver. I also utilize GridFS features. 

And everything works fine php7.2.13 + 3.4.2 on Alpine 3.8 :)
Have You checked that?

Shortcut - base collection class which is replacing Mvc\Collection. (i am using in Models namespace)
https://github.com/phalcon/incubator/blob/master/Library/Phalcon/Mvc/MongoCollection.php

If that stuff can be "ported" to .zep ... on re-implemented /cleaned.

wajdijurry:
I am using Phalcon MongoCollection and Client component to connect to
mongo. It works as a charm.
[...]

PR #14089, opened 2019-05-17
T14044 model findfirst null

niden:
Hello!

* Type: enhancement
* Link to issue: #14044 

**In raising this pull request, I confirm the following:**

- [x] I have read and understood the [Contributing Guidelines](https://github.com/phalcon/cphalcon/blob/master/CONTRIBUTING.md)
- [x] I have checked that another pull request for this purpose does not exist
- [x] I wrote some tests for this PR
- [x] I updated the CHANGELOG

 Changed `Phalcon\Mvc\Model::findFirst()` now returns `null`. `Phalcon\Mvc\Model::getRelated()` for one to one relationships returns `null` 

Thanks

Issue #14044, opened 2019-05-07
[RFC] Change return type of Model::findFirst or throw exception


Jurigag:
Simple as that. It feels weird that `Phalcon\Mvc\Model::findFirst` returns object or false, why false exactly, i never undestood this really. Null makes more sense and also it works with nullable types in PHP 7 and above.

I propose to add it to v4 while we still can.

virgofx:
Agree. I believe the return <tt>false</tt> is somewhat reminiscent of legacy/older PHP functions that often return <tt>false</tt>. It's a bad practice. 

I would be in favor of actually moving everything to use exceptions. I think we should use proper returns everywhere. Thus the return type would be `@return static` with an annotation for `@throws \Phalcon\Model\NotFoundException` or something. We should use exceptions everywhere and drop the global/orm option that defaulted this to false in 4.0

Jurigag:
I guess that's even better idea.

virgofx:
I would argue that `find()` should always return a `ResultSet` instance/interface 100% of the time such that static analysis can work. One of the interfaces from Complex/Simple. Even if it's empty. 

And I would say that we really should drop the ORM option with respect to `findFirst()` It makes it harder for example and what not. It would be better to have a consistent examples, documentation ...  

Best practice is the developer should handle their exceptions as necessary. Catch and return custom JSON, arrays, custom view logic, or simply allow the error to  bubble and be handled by default exception/error logic. It will automatically lead to better programming and cases of hidden bugs where people attempt to operate on a returned result of `null|false` (even if it's less common).

crlikcngroup:
I agree that false value can be bad practice.

But in our case it was quite useful. In caches for example. When attribute is null, you can tell entity was not queried. When false, you can discriminate between already queried but does not exist.
But a lot of programmers has a problem with this. So if it was handled via exceptions for example, cool.

scrnjakovic:
I disagree on couple of points:

1) Why should `Model::findFirst` throw an exception if nothing is found, and `Model::find` should return ResultSet?

2) Having exceptions being thrown at you for doing a simple query is wrong and costly. I don't think exceptions are meant to be used this way. It implies that record is always expected to exist in the database, which is really wrong. Exceptions should be used to signal serious errors from which a method is unable to recover. Fetching something if it exists, only to find out it doesn't, isn't an exceptional state.

Xplouder:
[...]

PR #13779, opened 2019-01-21
add missing method

fergeben:
Hello!

* Type: bug fix
* Link to issue: none

**In raising this pull request, I confirm the following (please check boxes):**

- [X] I have read and understood the [Contributing Guidelines](https://github.com/phalcon/cphalcon/blob/master/CONTRIBUTING.md)?
- [X] I have checked that another pull request for this purpose does not exist.
- [ ] I wrote some tests for this PR.

Small description of change:
Implement a missing method on Mvc\Collection.
Even if the mongo driver is deprecated, this is a bug.

Thanks

PR #14230, opened 2019-07-03
Feature/new mongodb driver

tacxou:
Hello!

* Type: breaks bc | new feature
* Link to issue: https://github.com/phalcon/cphalcon/issues/13697

* Added `Phalcon\Mvc\Collection\Document::__construct()`
* Added `Phalcon\Mvc\Collection\Document::count()`
* Changed `Phalcon\Mvc\Collection\Document`  
  * Now implements `Countable`
* Changed `Phalcon\Mvc\Collection::findFirst()`
  * Now returns a `null` or `CollectionInterface` 
* Changed `Phalcon\Mvc\Collection::getId()`
  * Now returns a `\MongoDB\BSON\ObjectId` 
* Depreciated `Phalcon\Mvc\Collection::createIfNotExist()`
* Deleted `Phalcon\Mvc\Collection::summatory()`
  * This method to group records no longer exists in the new driver. It can be easily replaced by the MongoDB [$group](https://docs.mongodb.com/manual/reference/operator/aggregation/group/) in [aggregation framework](https://docs.mongodb.com/manual/aggregation/), you can use `Phalcon\Mvc\Collection::aggregate()` method. 
* Changed `Phalcon\Mvc\Collection\Manager::getCustomEventsManager()`
  * Now return a `null` or `Phalcon\Events\ManagerInterface`
* Changed `Phalcon\Mvc\Collection\Manager::getConnection()`
  * Now return a `\MongoDB\Database` or `Phalcon\Db\Adapter\AdapterInterface`
* Refactoring `Phalcon\Mvc\Collection\Manager`
  * Events triggered in collections are now prefixed with `collection:` instead of `model:`
  * Modification of variable names to match collections (`var model` to `var collection`)

- Refactoring `Phalcon\Mvc\Collection`
  - Modification of variables of type `MongoId` to the new type `ObjectId`
  - Modification of variable names to match collections (`var mongoId` to `var objectId`)

----

Thanks


sergeyklay:
@tacxticx88 Could you please rebase

phalcon/Mvc/Collection.zep
```diff
@@ -1689,4 +1644,46 @@ abstract class Collection implements EntityInterface, CollectionInterface, Injec
 
         return collection;
     }
+
+    /**
+	 * {@inheritdoc}
```
sergeyklay: Could you correct indents

phalcon/Mvc/Collection.zep
```diff
@@ -1689,4 +1644,46 @@ abstract class Collection implements EntityInterface, CollectionInterface, Injec
...
+		let this->modelsManager = manager;
+
+		/**
+		 * Update the objects attributes
+		 */
+		for key, value in data {
+			let this->{key} = value;
+		}
+
+		if (method_exists(this, "afterFetch")) {
```
sergeyklay:
```suggestion
		if method_exists(this, "afterFetch") {
```

But in general I wold like work with interfaces

phalcon/Mvc/Collection/ManagerInterface.zep
```diff
@@ -41,56 +40,56 @@ interface ManagerInterface
...
      */
-    public function addBehavior(<CollectionInterface> model, <BehaviorInterface> behavior);
+    public function addBehavior(<CollectionInterface> collection, <BehaviorInterface> behavior);
 
     /**
-     * Returns the connection related to a model
+     * Returns the connection related to a collection
      */
-    public function getConnection(<CollectionInterface> model) -> <AdapterInterface>;
+    public function getConnection(<CollectionInterface> collection);
```
sergeyklay: what about return type?

phalcon/Mvc/Collection.zep
```diff
@@ -1015,6 +1024,8 @@ abstract class Collection implements EntityInterface, CollectionInterface, Injec
 
     /**
      * Allows to perform a summatory group for a column in the collection
+     *
+     * @deprecated
```
sergeyklay: Please don't introduce deprecations in the brand new version

phalcon/Mvc/Collection.zep
```diff
@@ -1194,24 +1205,18 @@ abstract class Collection implements EntityInterface, CollectionInterface, Injec
...
-        if typeof status == "array" {
-            if fetch ok, status["ok"] {
-                if ok {
-                    let success = true;
-                }
-            }
-        } else {
-            let success = false;
+        if status->isAcknowledged() {
+            let success = true;
```
sergeyklay: Otherwise NULL?

phalcon/Mvc/Collection.zep
```diff
@@ -558,13 +560,13 @@ abstract class Collection implements EntityInterface, CollectionInterface, Injec
      * }
      * ```
      */
-    public static function findById(var id) -> <CollectionInterface> | null
+    public static function findById(var id) -> <CollectionInterface> | bool
```
sergeyklay: Could you please return this and all other similar changes (null -> bool) back.

sergeyklay (changes requested):

phalcon/Mvc/Collection.zep
tacxticx88: Okay, should I delete the methods?

phalcon/Mvc/Collection/ManagerInterface.zep
tacxticx88: getConnection returns a "MongoDB Database", do I have to correct that? In phalcon:3.4.4.x, getConnection returned "\MongoDB"

phalcon/Mvc/Collection.zep
tacxticx88: the "findFirst" and "findById" must return a null? Because in the original code, the return type was "Collection" or "false" ?

phalcon/Mvc/Collection.zep
tacxticx88: What do you mean?

phalcon/Mvc/Collection.zep
tacxticx88: should I delete the methods in this case?

sergeyklay: Yes, I think so. And don't forget to amend change log for _any notable change_, e.g. changed method signature,  method deletion, etc.

phalcon/Mvc/Collection.zep
sergeyklay: I mean that `success` is NULL before this line

phalcon/Mvc/Collection.zep
sergeyklay: Please make it consistent with https://github.com/phalcon/cphalcon/issues/14044

phalcon/Mvc/Collection/ManagerInterface.zep
sergeyklay: as a far as I see it should return an interface/abstract class instance, isn't?
```zep
        // ...
        let connection = container->getShared(service);
        // ...

        return connection;
```

tacxticx88: I didn't change this part of the code, it returns an instance of \MongoDB\Database;  in 3.4.x it returned an instance of \Mongo

phalcon/Mvc/Collection.zep
tacxticx88: It works, I'll do it. For the "createIfNotExists" method, is it useful knowing that it is not documented and that the "uniqness" validator replaces this behavior ?

phalcon/Mvc/Collection.zep
tacxticx88: Do the methods have to lift an exception?

phalcon/Mvc/Collection/ManagerInterface.zep
sergeyklay: But now it isn't 3.4. Let's change method signature

phalcon/Mvc/Collection.zep
niden: You can either add that in the changelog or add it to the documents or even send me a message so that we can add it in the documentation. :)

phalcon/Mvc/Collection.zep
niden: The `findFirst` etc. have been changed to return `null` or the record itself. In this case it will be a Collection object. Have a look at the implementation that the model has and let us replicate that. 

These are the changes made to make the methods return `null`

https://github.com/phalcon/cphalcon/pull/14089/files

niden:
@tacxticx88 when you have some time can you look at this PR:

https://github.com/phalcon/cphalcon/pull/13779/files

to make sure that the supplied fix is included with your PR?

Thanks

phalcon/Mvc/Collection.zep
tacxticx88: Normally these changes are made now

phalcon/Mvc/Collection.zep
```diff
@@ -32,7 +31,7 @@ use Serializable;
  * This component implements a high level abstraction for NoSQL databases which
  * works with documents
  */
-abstract class Collection implements EntityInterface, CollectionInterface, InjectionAwareInterface, Serializable
+abstract class Collection implements EntityInterface, CollectionInterface, InjectionAwareInterface, Unserializable, \Serializable
```
sergeyklay: Why do we need implement Unserializable? Can we avoid this?

tacxticx88: It is this interface and the bsonUnserialize method that allows you to fill in the result sets

sergeyklay: Not all users have mongodb installed. Declaring interface here will cause fatal error for those users who don not have mongodb installed even they not use Collection class at all. Just at PHP startup time

tacxticx88: If I delete this interface, MongoDB raises an Exception

`[MongoDB\Driver\Exception\InvalidArgumentException] Class Phalcon\Test\Fixtures\Mvc\Collections\Robots does not implement MongoDB\BSON\Unserializable`

https://www.php.net/manual/en/mongodb-driver-cursor.toarray.php

phalcon/Mvc/Collection.zep
tacxticx88: I made some modifications, is that good for you?

phalcon/Mvc/Collection.zep
sergeyklay: Can you declare Robots implemets Unserializable and leave Collection as is?

niden: @

> If I delete this interface, MongoDB raises an Exception
> 
> `[MongoDB\Driver\Exception\InvalidArgumentException] Class Phalcon\Test\Fixtures\Mvc\Collections\Robots does not implement MongoDB\BSON\Unserializable`
> 
> https://www.php.net/manual/en/mongodb-driver-cursor.toarray.php

We cannot use that interface there because that introduces a limitation for the framework. Here is why:

We have introduced PDO (a few versions back). That came with some interfaces that we wanted to use to wrap Phalcon classes. As a result, because these are interfaces, we added the requirement to load PDO first before Phalcon

With v4 we introduced PSR-7 and its interfaces. New requirement: PSR has to be loaded first.

If we add this interface there, it will require all users, irrespective of whether they want to use Mongo or not, to load the mongo extension for Phalcon to work.

You might ask why does `apcu` work then if I do not have it installed? It works because we use functions for it and not interfaces. Another example is one of the `Redis` adapters. If you check the code, we create the adapter inside Phalcon and use it from there, but never expose any interfaces that Redis or any other extension uses.

I think the best way forward is to remove the interface from there and require models to implement the `Unserializable` interface for the mongo collection to work.

In short, if we introduce one interface from a different extension, Zend Engine requires it in every request, even though you might not want to use that interface ever.

tacxticx88: The solution is what then, to create an interface if it does not exist, or not to use the "typeMap" functionality ?

tacxticx88: https://www.php.net/manual/fr/mongodb.persistence.deserialization.php
This method still offers many possibilities. It allows you to transform a field into a class, for example

CameronHall:
What is this waiting on?

phalcon/Mvc/Collection.zep
sergeyklay: Require models to implement the Unserializable interface for the MongoDB collection will do all we need here. Just  implement all interface requirements in this PR but do not declare "implement ..." interface. Indicate in the comment that the user must declare such a inheritance for each of its model of his own.

tacxticx88: Okay, I'll fix it and take care of it.

ruudboon:
@tacxticx88 Can you fix the 2 line enters to pass the last test?
https://travis-ci.org/phalcon/cphalcon/jobs/567264410

phalcon/Mvc/Collection.zep
```diff
@@ -631,15 +632,15 @@ abstract class Collection implements EntityInterface, CollectionInterface, Injec
...
+     *             "_id" => new \MongoDB\BSON\ObjectId("45cbc4a0e4123f6920000002"),
      *         ]
      *     ]
      * );
      *
      * echo "The robot id is ", $robot->_id, "\n";
      * ```
      */
-    public static function findFirst(array parameters = null) -> array
+    public static function findFirst(array parameters = null) -> <CollectionInterface> | null
```
sergeyklay: This change should be reflected in the change log.

phalcon/Mvc/Collection.zep
```diff
@@ -752,7 +753,7 @@ abstract class Collection implements EntityInterface, CollectionInterface, Injec
     /**
      * Returns the value of the _id property
      *
-     * @return \MongoId
```
sergeyklay: This change should be reflected in the change log.

phalcon/Mvc/Collection.zep
```diff
@@ -1013,55 +1021,6 @@ abstract class Collection implements EntityInterface, CollectionInterface, Injec
         let this->skipped = skip;
     }
 
-    /**
-     * Allows to perform a summatory group for a column in the collection
-     */
-    public static function summatory(string! field, conditions = null, finalize = null) -> array
```
sergeyklay: Why you decided to remove this method? Anyway, should we reflect this removal in the change log?

phalcon/Mvc/Collection/Document.zep
```diff
@@ -20,8 +20,30 @@ use Phalcon\Mvc\Collection\Exception;
  * This component allows Phalcon\Mvc\Collection to return rows without an associated entity.
  * This objects implements the ArrayAccess interface to allow access the object as object->x or array[x].
  */
-class Document implements EntityInterface, ArrayAccess
+class Document implements EntityInterface, \ArrayAccess, \Countable
```
sergeyklay: This change should be reflected in the change log.

phalcon/Mvc/Collection/Manager.zep
```diff
@@ -104,21 +105,21 @@ class Manager implements InjectionAwareInterface, EventsAwareInterface
     }
 
     /**
-     * Returns the connection related to a model
+     * Returns the connection related to a collection
      *
-     * @return \Mongo
+     * @return \MongoDB\Database
```
sergeyklay:
```suggestion
     * @return \MongoDB\Database|AdapterInterface
```

And this change should be reflected in the change log.

phalcon/Mvc/Collection/Manager.zep
```diff
@@ -278,30 +279,30 @@ class Manager implements InjectionAwareInterface, EventsAwareInterface
          */
         let eventsManager = this->eventsManager;
         if typeof eventsManager == "object" {
-            return eventsManager->fire("model:" . eventName, model, data);
+            return eventsManager->fire("collection:" . eventName, collection, data);
```
sergeyklay: This change should be reflected in the change log.

phalcon/Mvc/Collection/ManagerInterface.zep
```diff
@@ -41,56 +41,56 @@ interface ManagerInterface
...
      */
-    public function getConnection(<CollectionInterface> model) -> <AdapterInterface>;
+    public function getConnection(<CollectionInterface> collection) -> <AdapterInterface>;
 
     /**
-     * Returns a custom events manager related to a model
+     * Returns a custom events manager related to a collection
      */
-    public function getCustomEventsManager(<CollectionInterface> model) -> <EventsManagerInterface>;
+    public function getCustomEventsManager(<CollectionInterface> collection) -> var | null;
```
sergeyklay: I propose change this to 
```suggestion
    public function getCustomEventsManager(<CollectionInterface> collection) -> <EventsManagerInterface> | null;
```

as well as https://github.com/phalcon/cphalcon/blob/3addad41e14ee83d6bb76138fb4706cadff50182/phalcon/Mvc/Collection/Manager.zep#L92

phalcon/Mvc/CollectionInterface.zep
```diff
@@ -52,7 +52,7 @@ interface CollectionInterface
     /**
      * Allows to query the first record that match the specified conditions
      */
-    public static function findFirst(array parameters = null) -> array;
+    public static function findFirst(array parameters = null) -> <CollectionInterface> | null;
```
sergeyklay: This change should be reflected in the change log.

phalcon/Mvc/CollectionInterface.zep
```diff
@@ -124,9 +124,4 @@ interface CollectionInterface
      * @param mixed id
      */
     public function setId(id);
-
-    /**
-     * Check whether validation process has generated any messages
-     */
-    public function validationHasFailed() -> bool;
```
sergeyklay: Why you decided to remove this method? I would like to see a descriptive reason in the change log

sergeyklay (commented):
There are some minor questions I would like understand to. Also we'll need update change log (see my comments).  But at the moment it seems good to me.

phalcon/Mvc/CollectionInterface.zep
tacxticx88: The method did not exist in the collection, I just synchronized

ruudboon:
@tacxticx88 If you have time can you rebase your branch to see where we stand?

tacxticx88:
@ruudboon I'll do that. Yes, my work was moved to a private storage facility before the switch to Phalcon v4.

ruudboon:
@tacxticx88 We're currently upgrading our testsuite as well. If you need some help rewriting the test let us know.

ruudboon:
@tacxticx88 Any news on this? If you have time to work on this can you switch to the 4.1.x branch? thnx

niden:
Mentioning this one more time just in case. 

For this PR to work and to be part of core, we need to ensure that it does not reference any classes or interfaces that a different extension has (in this case the php mongo extension). Doing so will require that extension to be loaded first before Phalcon even starts. Example is the PSR extension which is now a requirement for Phalcon to work.

tacxticx88:
Hello, I'm currently looking for a solution to this problem because the Collection class should implement the interface of MongoDb to be able to automatically caste the model and its child classes..

tacxticx88:
I suggest to finally write the ODM in PHP in order to take advantage of MongoDb's Composer repository.  In my case the work is already well advanced. What do you think of that?  MongoDb having its driver written mostly in PHP.

niden:
> I suggest to finally write the ODM in PHP in order to take advantage of MongoDb's Composer repository. In my case the work is already well advanced. What do you think of that? MongoDb having its driver written mostly in PHP.

@tacxticx88 I think that might be the best solution at least for the time being until we regroup and figure out a way to solve the technical issues that will appear by including the mongo interfaces in the core.

We can set up an incubator repository and all the work and support can go there. People will then be able to add the component using composer.

Maybe later on we can include this functionality with FFI in the core or something similar - not sure now just thinking out loud.

I agree, we need to get this done in the incubator.

ruudboon:
@tacxticx88 Do you agree on closing this one?

tacxticx88:
> @tacxticx88 Do you agree on closing this one?

Yes. See the new phalcon/incobator-mongodb repository

ruudboon:
Closing this in favour of the new https://github.com/phalcon/incubator-mongodb repo

Issue #12396, opened 2016-11-07
Segfaults on user level (syntax) errors


stamster:
### Expected and Actual Behavior

PHP fatal/syntax error.

> **Describe what you are trying to achieve and what goes wrong.**

From any controller, trying to call simple facade method defined in ControllerBase or wherever, making a mistake which should cause only syntax error.
Instead, nginx throws 502 Bad Gateway HTTP response, while PHP-FPM throws:
`[480094.083490] php-fpm7.0[30488]: segfault at 1c ip 00005563990e8d28 sp 00007ffdf56be410 error 4 in php-fpm7.0[556398e93000+3a9000]`

**Backtrace**:

```
Core was generated by `php-fpm: pool middleware-php7                                                '.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00005563990e8d28 in _object_and_properties_init ()
(gdb) bt
#0  0x00005563990e8d28 in _object_and_properties_init ()
#1  0x00007f53984f47f7 in zim_Phalcon_Mvc_Micro_LazyLoader___call () from /usr/lib/php/20151012/phalcon.so
#2  0x00005563990d62fa in dtrace_execute_internal ()
#3  0x000055639916a2b3 in ?? ()
#4  0x00005563991265eb in execute_ex ()
#5  0x00005563990d6191 in dtrace_execute_ex ()
#6  0x00005563990d7df9 in zend_call_function ()
#7  0x00007f5398263b39 in zephir_call_user_func_array_noex () from /usr/lib/php/20151012/phalcon.so
#8  0x00007f53984f3199 in zim_Phalcon_Mvc_Micro_handle () from /usr/lib/php/20151012/phalcon.so
#9  0x00005563990d62fa in dtrace_execute_internal ()
#10 0x000055639916afa0 in ?? ()
#11 0x00005563991265eb in execute_ex ()
#12 0x00005563990d6191 in dtrace_execute_ex ()
#13 0x000055639917a3b7 in zend_execute ()
#14 0x00005563990e6393 in zend_execute_scripts ()
#15 0x0000556399086c00 in php_execute_script ()
#16 0x0000556398f6d626 in main ()
```

> Provide minimal script to reproduce the issue

```php
var_dump(
            $this->auth('   contact@my(site).com', '        mytestpass123'),
        );
```
(if I remove syntax error - comma sign at the end of the statement, stack works fine)

This can happen anytime developer will omit closing `}` or comma sign.
[...]

PR #14238, opened 2019-07-09
Fix Phalcon\Mvc\Micro\LazyLoader segfault

davidofferman:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/12396

Small description of change:
Fixes a segfault when using Micro with Collections. The segfault happens when you lazy load a handler that has a syntax error.

Sample Application:
This sample application with `index.php` and `controllers/BadController.php`. The controller is missing a semi-colon on purpose when assigning the array to a variable.

```
<?php

try {
	$collection = new Phalcon\Mvc\Micro\Collection();
	$collection->setPrefix("");
	$collection->setHandler("Controller\\BadController");
	$collection->setLazy(true);
	$collection->head("/", "bugAction");
	$collection->get("/", "bugAction");

	$loader = new Phalcon\Loader();
	$loader->registerNamespaces([
		"Controller" => __DIR__ . "/controllers/"
	]);

	$loader->register();

	$app->mount($collection);
	$app->handle();
} catch(Throwable $t) {
	print_r($t);
}
```
```
<?php

namespace Controller;

class BadController {

	public function bugAction($slug) {
		$a = []
	}

}
```

Thank you.

### DIFF
--- a/phalcon/mvc/micro/lazyloader.zep
+++ b/phalcon/mvc/micro/lazyloader.zep
@@ -60,7 +60,11 @@ class LazyLoader
 		let definition = this->_definition;
 
 		if typeof handler != "object" {
-			let handler = new {definition}();
+			if(!class_exists(definition)) {
+				throw new Exception("Handler '" . definition . "' doesn't exist");
+			}
+
+			let handler = create_instance(definition);
 			let this->_handler = handler;
 		}
 
--- a/phalcon/mvc/micro/lazyloader.zep
+++ b/phalcon/mvc/micro/lazyloader.zep
@@ -60,7 +60,7 @@ class LazyLoader
 		let definition = this->_definition;
 
 		if typeof handler != "object" {
-			if(!class_exists(definition)) {
+			if !class_exists(definition) {
 				throw new Exception("Handler '" . definition . "' doesn't exist");
 			}
 
--- a/CHANGELOG-3.4.md
+++ b/CHANGELOG-3.4.md
@@ -1,9 +1,11 @@
+# [3.4.5](https://github.com/phalcon/cphalcon/releases/tag/v3.4.5) (2019-XX-XX)
+- Fixed segfault in `Phalcon\Mvc\Micro\LazyLoader::__call()` when the handler has a syntax error. [#12396](https://github.com/phalcon/cphalcon/issues/12396)
+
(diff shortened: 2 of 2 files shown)


niden:
@davidofferman Thank you for this. Could you please target the `4.x` branch and update the changelog with your change?

Thanks!

phalcon/mvc/micro/lazyloader.zep
```diff
@@ -60,7 +60,11 @@ class LazyLoader
 		let definition = this->_definition;
 
 		if typeof handler != "object" {
-			let handler = new {definition}();
+			if(!class_exists(definition)) {
+				throw new Exception("Handler '" . definition . "' doesn't exist");
+			}
+
+			let handler = create_instance(definition);
```
sergeyklay: As a far as I know `new {definition}();` faster than `create_instance(definition);`. So I'm not sure we should change this after class_exists check. Could you elaborate on that

phalcon/mvc/micro/lazyloader.zep
```diff
@@ -60,7 +60,11 @@ class LazyLoader
 		let definition = this->_definition;
 
 		if typeof handler != "object" {
-			let handler = new {definition}();
+			if(!class_exists(definition)) {
```
sergeyklay:
```diff
-			if(!class_exists(definition)) {
+			if !class_exists(definition) {
```

sergeyklay (changes requested):

phalcon/mvc/micro/lazyloader.zep
davidofferman: I'm not sure why `new {definition}()` causes a segment fault, but it does even if the class exists when there is a syntax error in the file inside of `_object_and_properties_init`.

I ended up using `create_instance` since it is the same function that `Phalcon\Di` uses to create classes and that doesn't have a segment fault when loading classes with syntax errors.

sergeyklay (dismissed):

sergeyklay:
Thank you

sergeyklay:
@niden 4.0.x: https://github.com/phalcon/cphalcon/pull/14243

stamster:
Thumbs up for @davidofferman

PR #14259, opened 2019-07-24
Annotations Router to use patterns as prefix

zsilbi:
Hello!

* Type: performance tuning

To use `Mvc\Router\Annotations`, you have to add the resources that you want it to parse. 
You can assign a prefix for every resource for faster handling.

```php
$di->setShared(
     "router",
     function() {
         // Use the annotations router
         $router = new Annotations(false);

         // This will do the same as above but only if the handled uri starts with /robots
         $router->addResource("Robots", "/robots");

        return $router;
     }
);
```

Adding controllers with distinct prefixes will work great, even in a huge a project.

If you don't use prefixes, it will basicly need to check and parse every resource that you added. It will be a bit slow but will work good enough for small projects.

However, if you have a lot of controllers with prefixes that have parameters or placeholders in the beginning, you won't be able to use them. 

For example if you have 75 controllers with prefixes like this: 
 - /clients/{clientId:[0-9]+}/
 - /clients/{clientId:[0-9]+}/robots
 - /clients/{clientId:[0-9]+}/parts
 - ...

In this case, only the `/clients` prefix can be used for all the 75 controllers.

In my test runs, I used a total of 150 resources, 75 with the same prefix and the average route handling time was over 50ms. Without using any prefixes, it was around 90ms.

**My idea would be to compile the prefixes as standard route patterns, so you could pass any type of pattern as a prefix along with your resource to the `Router`.**

I repeated the same tests using this code, and my average route handling time was around 3ms which I find somewhat acceptable. 

Thanks,
zsilbi

### DIFF
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -12,6 +12,8 @@
 - Changed `Phalcon\Flash\Direct` to allow `escaper` in the constructor [#14349](https://github.com/phalcon/cphalcon/issues/14349)
 - Changed `Phalcon\Flash\Session` to allow `escaper` in the constructor [#14349](https://github.com/phalcon/cphalcon/issues/14349)
 - Changed `Phalcon\Di\AbstractDIAware` to `Phalcon\Di\AbstractInjectionAware` [#14359](https://github.com/phalcon/cphalcon/issues/14359)
+- Changed `Phalcon\Mvc\Router\Annotations` to be able to handle patterns az prefixes [#14259](https://github.com/phalcon/cphalcon/pull/14259)
+- Changed `Phalcon\Mvc\Router\Group::routes` to an array as default [#14259](https://github.com/phalcon/cphalcon/pull/14259)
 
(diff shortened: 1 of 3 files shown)


ruudboon:
If you rebase probably your mysql will build again.

composer.json
```diff
@@ -45,7 +45,7 @@
     "ext-json": "*",
     "ext-libxml": "*",
     "ext-mbstring": "*",
-    "ext-psr": "^0.6",
+    "ext-psr": "^0.7",
```
zsilbi: I am going to revert this, as soon as #14297 is merged.

sergeyklay: @zsilbi Fixed in the 4.0.x branch. Could you please rebase

zsilbi: Rebased.

sergeyklay (approved):

niden (approved):

niden:
Thank you @zsilbi

Issue #14260, opened 2019-07-24
Travis build error during mysql-server install


zsilbi:
It fails to install  `mysql-server`, so all the new CI builds fail.
See: https://github.com/phalcon/cphalcon/pull/14259

```
The following packages have unmet dependencies:
 mysql-server : Depends: mysql-server-5.5 but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
Looking for 'mysql' as: mysql
Looking for 'mysqlcheck' as: mysqlcheck
This installation of MySQL is already upgraded to 5.6.33, use --force if you still need to run mysql_upgrade
The command ".ci/install-mysql-5.7.sh" failed and exited with 1 during .
```
https://travis-ci.org/phalcon/cphalcon/builds/563215912

sergeyklay:
It is weird that file called install-mysql-5.7.sh tries to install mysql 5.5. @SidRoberts Any idea?

SidRoberts:
I have no idea but I had a similar problem a while back though. The official builds would run fine on Travis but mine would fail - interestingly because of this MySQL script. The problem was because Travis was using different Ubuntu versions. I fixed this in https://github.com/phalcon/cphalcon/pull/14189 by specifying trusty.

It's interesting to note that on this commit, the Phalcon build passed but my build failed:
- https://travis-ci.org/phalcon/cphalcon/jobs/561642481
- https://travis-ci.org/SidRoberts/cphalcon/jobs/561694697

I'll take a look and see what I can remember.

SidRoberts:
Just to provide an update to this, I'm currently working on changing the Ubuntu version used in Travis to 18.04 (bionic). We're currently using 14.04 (trusty) which is woefully outdated and I suspect might be causing some of these problems. It's not complete yet but should be finished today.

sergeyklay:
Fixed in the `4.0.x` branch. Feel free to open a new issue if the problem appears again. Thank you for contributing.

PR #14268, opened 2019-07-30
Travis now uses Ubuntu 18.04 for builds

SidRoberts:
Hello!

* Type: bug fix
* Link to issue: #14260 

Two important things of note:

* [From OpenSSL 1.1.1, ciphers are lowercased](https://www.php.net/manual/en/function.openssl-get-cipher-methods.php#123319)
* [Memcached::getAllKeys() appears to be broken from 1.4.23](https://www.php.net/manual/en/memcached.getallkeys.php#123793)

I've tried to alter the code to work around these problems but I'd like someone with a fresh set of eyes to double check as I've been looking at this too long. :stuck_out_tongue_winking_eye:

### DIFF
--- a/tests/unit/Cache/Adapter/Libmemcached/GetKeysCest.php
+++ b/tests/unit/Cache/Adapter/Libmemcached/GetKeysCest.php
@@ -39,6 +39,26 @@ class GetKeysCest
             getOptionsLibmemcached()
         );
 
+        $memcachedServerVersions = $adapter->getAdapter()->getVersion();
+        $memcachedExtensionVersion = phpversion('memcached');
+
+        foreach ($memcachedServerVersions as $server => $memcachedServerVersion) {
+            // https://www.php.net/manual/en/memcached.getallkeys.php#123793
+            // https://bugs.launchpad.net/libmemcached/+bug/1534062
+            if (version_compare($memcachedServerVersion, '1.4.23', '>=') && version_compare($memcachedExtensionVersion, '3.0.1', '<')) {
+                $I->skipTest(
+                    'getAllKeys() does not work in certain Memcached versions'
+                );
+            }
+
+            // https://github.com/php-memcached-dev/php-memcached/issues/367
+            if (version_compare($memcachedServerVersion, '1.5.0', '>=')) {
+                $I->skipTest(
+                    'getAllKeys() does not work in certain Memcached versions'
+                );
+            }
+        }
+
         $adapter->clear();
 
         $adapter->set('key-1', 'test');
--- a/tests/unit/Storage/Adapter/Libmemcached/GetKeysCest.php
+++ b/tests/unit/Storage/Adapter/Libmemcached/GetKeysCest.php
@@ -39,6 +39,26 @@ class GetKeysCest
             getOptionsLibmemcached()
         );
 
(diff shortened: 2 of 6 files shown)


tests/unit/Cache/Adapter/Libmemcached/GetKeysCest.php
```diff
@@ -39,6 +39,17 @@ public function cacheAdapterLibmemcachedGetKeys(UnitTester $I)
             getOptionsLibmemcached()
         );
 
+        $versions = $adapter->getAdapter()->getVersion();
+
+        foreach ($versions as $server => $version) {
+            // https://www.php.net/manual/en/memcached.getallkeys.php#123793
+            if (version_compare($version, '1.4.23', '>=')) {
```
sergeyklay: As I can see it is fixed https://bugs.launchpad.net/libmemcached/+bug/1534062
Could we compare more specific version range  here?

sergeyklay: Also add please launchpad link too,

tests/unit/Storage/Adapter/Libmemcached/GetKeysCest.php
```diff
@@ -39,6 +39,17 @@ public function storageAdapterLibmemcachedGetKeys(UnitTester $I)
             getOptionsLibmemcached()
         );
 
+        $versions = $adapter->getAdapter()->getVersion();
+
+        foreach ($versions as $server => $version) {
+            // https://www.php.net/manual/en/memcached.getallkeys.php#123793
+            if (version_compare($version, '1.4.23', '>=')) {
```
sergeyklay: See my previous comment

sergeyklay (commented):
In general looks good to me

sergeyklay (dismissed):

sergeyklay:
It's a little bit confusing with the bugs in the getAllKeys() method

- One says it doesn't work with memcached 1.5+: https://github.com/php-memcached-dev/php-memcached/issues/367
- One says it doesn't work with php-memcached before 3.0.1: https://github.com/php-memcached-dev/php-memcached/issues/203#issuecomment-282041918

sergeyklay:
@SidRoberts Could you please change memcached version check to reflect this 
See: https://github.com/php-memcached-dev/php-memcached/pull/315

Make sure you use php-memcached extension >= 3.0.1 and memcached server >= 1.4.24

SidRoberts:
I agree that this is very confusing. Currently Travis is using Memcached server 1.5.6 and Memcached extension 3.1.3. In theory, it should work [but it doesn't](https://travis-ci.org/phalcon/cphalcon/jobs/565634267#L797).

I don't like the idea of a test doing version checking. I think the version checks should be in the `getKeys()` method and we should maybe implement code that actually gets all the keys instead of dealing with this mess.

How about keep the version check at >=1.2.24 for now, merge it and then deal with it later?

SidRoberts:
Or, as a last ditch effort, what about https://github.com/phalcon/cphalcon/pull/14268/commits/bc0712eeed23fd27f9c07e82640d5e52e24e46b2?

CameronHall (approved):
Looking good 👍

sergeyklay (approved):

sergeyklay:
Thank you

Issue #14270, opened 2019-07-31
Related models not saved when querying them first


gwijnja:
* Phalcon version: 3.4.4
* PHP Version: 7.2.19
* Operating System: CentOS 7
* Installation type: installing via package manager
* Server: Apache, but also on the command line
* Other related info (Database, table schema): MySQL 5.7.27

I'm opening this issue/bug report on request in this topic: https://forum.phalconphp.com/discussion/19903/robot-parts-not-updated-after-querying-parts-first

Please have a look at the code below. It's the default Robots/Parts/RobotsParts setup. Adding parts to the robot works fine. But when querying the parts first using `$robot->parts` and then adding new parts, nothing is saved to the database.

I commented the code below to illustrate what goes wrong. The models are stored in MySQL tables with cascading foreign keys.

```php
<?php

class Robot extends \Phalcon\Mvc\Model
{
    public $id;
    public $name;

    public function initialize()
    {
        $this->setSource('robots');
        $this->hasManyToMany(
            'id', RobotsParts::class, 'robot_id',
            'part_id', Part::class, 'id',
            ['alias' => 'parts']
        );
    }
}
```

```php
<?php

class RobotsParts extends \Phalcon\Mvc\Model
{
    public $id;
    public $robot_id;
    public $part_id;
}
```

```php
<?php

class Part extends \Phalcon\Mvc\Model
{
    public $id;
    public $name;

    public function initialize()
    {
        $this->setSource('parts');
    }
}
```

```php
<?php

use Phalcon\Cli\Task;

class MainTask extends Task
{
    /**
     * Test sequence
     */
    public function testAction()
    {
        $this->initAction();
        $this->addAction();
        $this->bugAction();

        $robot = Robot::findFirstByName("bob");
        foreach ($robot->parts as $part) {
            printf("%s (id %d)\n", $part->name, $part->id);
        }
    }

    /**
     * Delete everything in the database,
     * then add 1 robot with 2 parts.
     */
    public function initAction()
    {
        Robot::find()->delete();
        Part::find()->delete();
        // The link table (RobotsParts) has cascading foreign keys,
        // so they will be deleted by MySQL.

        $part1 = new Part;
        $part1->name = "arm";

        $part2 = new Part;
        $part2->name = "leg";

        $robot = new Robot;
        $robot->name = "bob";
        $robot->parts = [$part1, $part2];
        $robot->save(); // Works as expected
    }

    /**
     * Add 2 more parts
     */
    public function addAction()
[...]

PR #14277, opened 2019-08-03
Functional test for issue 14270

ruudboon:
Created a test for #14270. Can't reproduce this behaviour on 4.0.x but would like to keep this test as standard test.

### DIFF
--- a/tests/integration/Mvc/Model/SaveCest.php
+++ b/tests/integration/Mvc/Model/SaveCest.php
@@ -380,10 +380,10 @@ class SaveCest
     {
         $I->wantToTest('Mvc\Model::save() after feteching related');
 
-        $part1 = new Part;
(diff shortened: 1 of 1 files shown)


sergeyklay (approved):

sergeyklay:
@ruudboon Do you want to backport this test to the 3.4.x branch?

sergeyklay:
Thank you for the test!

ruudboon:
> @ruudboon Do you want to backport this test to the 3.4.x branch?

I tried, but the test is based on all the new Cest stuff so it will involve writing a different test.
I've contacted @gwijnja to investigate this issue a bit more. I've checked the changes in v4 for the model and I'm not sure if this is fixed. I think I'm missing some condition. I'm almost certain I had this problem a few weeks ago in V4.

Issue #14269, opened 2019-07-31
"$this->eventsManager" return null


zuozhehao:
phalcon 4

I found that "protected event manager" would affect the magic method _get to get "eventsManager" when I did not use "setEventsManager"

https://github.com/phalcon/cphalcon/blob/4173bca9b85b4b7e3493146036ed91b977dff390/phalcon/Di/Injectable.zep#L60-L82

CameronHall:
In which case we will need to add the `__isset` magic method to catch that. @niden

niden:
Sorry I do not understand the issue. The code is identical to v3 meaning `eventsManager` is protected 

What are you trying to do?

niden:
Never mind @CameronHall explained it

CameronHall:
It's worth considering removing the `eventsManager` property from the `DI` altogether alongside the `setEventsManager` method altogether. Considering no one has noticed (until now) that it hasn't been an issue, there should be minimal impact (if any).

CameronHall:
@zuozhehao can you please provide a code sample?

zuozhehao:
```php
class BaseController extends \Phalcon\Mvc\Controller
{

    public function afterExecuteRoute($dispatcher)
    {

        $this->eventsManager->fire('someEvent' ,$this);

    }

}
```

@CameronHall

CameronHall:
@zuozhehao how do you populate your Di? With `FactoryDefault`?

zuozhehao:
@CameronHall yes, `Phalcon\Di\FactoryDefault`

CameronHall:
Thanks @zuozhehao I'll take a look.

CameronHall:
@zuozhehao I've tried reproducing this but haven't be able too. After rereading through all the code, there's no reason why this shouldn't work, the only reason it wouldn't is if the the eventsManager isn't in your DI.

zuozhehao:
@CameronHall  
When initializing the controller, the `setEventsManager` was not executed,
When I used `this->eventsManager`, it did not go into the magic method`__get`
```php
<?php

class BaseController {


    /**
     * Events Manager
     *
     * @var \Phalcon\Events\ManagerInterface
     */
    protected $eventsManager;


    /**
     * Magic method __get
     */
    public function __get($propertyName) {

        var_dump($propertyName);

    }

    /**
     * Sets the event manager
     */
    public function setEventsManager($eventsManager)
    {

        $this->eventsManager = $eventsManager;

    }

}

[...]

PR #14280, opened 2019-08-04
Unittest for issue 14269

ruudboon:
Added unit test to test behaviour of issue #14269


sergeyklay (changes requested):
Could you please submit just the commits that belong to the issue, e.g. exclude session related changes?

CameronHall:
As discussed on Discord @ruudboon you can't use `FactoryDefault` to populate the DI that will not reproduce the issue in #14269.

Issue #14265, opened 2019-07-29
Phalcon\Session\Adapter\Stream   wrong open save path


ZhangRuiMingZRM:
* Phalcon version: `4.0.0-beta.1`
* PHP Version: `7.3.6`
* Operating System: ubuntu
* Installation type: Compiling from source
* Server: php web server
   
When running the following code and get this:
```
 Warning: Phalcon\Session\Adapter\Stream::write(/90192p0avuirodu6r2914kvn9o): failed to open stream: Permission denied in Unknown on line 0
```
```php
    $session = new Manager();
    $files = new Stream(
        [
            'savePath' => '/tmp',
        ]
    );
    $session->setHandler($files);
    $session->start();
```
This may be due to the `public function open(var savePath, var sessionName)` in the `phalcon/phalcon/Session/Adapter/Stream.zep` 
https://github.com/phalcon/cphalcon/blob/4173bca9b85b4b7e3493146036ed91b977dff390/phalcon/Session/Adapter/Stream.zep#L97
```
    public function open(var savePath, var sessionName) -> bool
    {
        var path;

        if true !== ends_with(savePath, "/") {
            let path = savePath . "/";
        }

        let this->path = path;

        return true;
    }
```

Previous attributes `path` are  reassigned

beareable:
what time this bug will be fixed?

ruudboon:
Want to fix it before the next release. Bug is fixed in attached pull above. But I'm struggling with the tests. Need to fix that before merging.

sergeyklay:
Fixed in the `4.0.x` branch. Feel free to open a new issue if the problem appears again. Thank you for the bug report.

bnlambert:
Hi

I am currently having this same issue on 

CentOS Linux 7.8.2003
PHP 7.4.10 
Phalcon 7.4

It just suddenly happened no idea on how to fix it

PR #14289, opened 2019-08-07
Fix for issue 14265

ruudboon:
Hello!

* Type: bug fix 
* Link to issue: #14265 

Small description of change:
- Set session_save_path so open() will be called with the correct path.
- Fixed tests that kept session open


ruudboon:
I think the issue is solved but need to find out why the other tests are failing. Looks like the session stays alive or the handler is still set to the stream

ruudboon:
If somebody got experience with testing sessions in Codeception. I could use some help.

CameronHall:
@sergeyklay is a Codeception wiz. Serghei do you think you could help @ruudboon?

sergeyklay:
@ruudboon As I can see the method `open` still overwrites the path by NULL. Can you please use your first approach (just return TRUE and ignore passed params).

ruudboon:
@sergeyklay Ok will do that. Btw. I believe the bug is fixed but looks like phpunit/codeception won't allow us to reset the session handler. Individual tests are working.

Issue #14331, opened 2019-08-26
Phalcon\Url\Get and double slashes


scrnjakovic:
The implementation of this method in `v3.4` may return links like:
`https://www.example.com//images/booya.jpg`

I see that this has been reported and fixed for `v4` in #13495 

However, why don't we utilize our own helper and replace these 8 lines

https://github.com/phalcon/cphalcon/blob/e78ab440fec479369bda68e69b9e880e58f5aa34/phalcon/Url.zep#L166-L174

with 

```zephir
let uri = \Phalcon\Helper\Str::reduceSlashes(baseUri . strUri);
```
furthermore, we can reduce the overhead of invoking an extra function and directly use

```zephir
let uri = preg_replace("#(?<!:)//+#", "/", baseUri . strUri);
```
The only scenario in which this wouldn't work is if we passed an url with relative protocols like `//netdna.bootstrapcdn.com/twitter-bootstrap/2.3.4/css/bootstrap-combined.min.css` but in that case, that wouldn't be a local URL, therefore it does not pose a problem.

cc @niden @sergeyklay @Jurigag

PR #14332, opened 2019-08-26
Update Phalcon\Url::get

scrnjakovic:
Hey guys.

* Type: bug fix && new feature
* Link to the issue: #14331 

Why do we complicate things? Let's utilize `Phalcon\Helper\Str::reduceSlashes` method instead of
```zephir
if substr(baseUri, -1) == "/" && strlen(strUri) > 2 && strUri[0] == '/' && strUri[1] != '/' {
    let uri = baseUri . substr(strUri, 1);
} else {
    if baseUri == "/" && strlen(strUri) == 1 && strUri[0] == '/' {
        let uri = baseUri;
    } else {
        let uri = baseUri . strUri;
    }
}
```

`Phalcon\Helper\Str::concat('/', baseUri, strUri)` would've worked as well.

### DIFF
--- a/phalcon/Url.zep
+++ b/phalcon/Url.zep
@@ -16,6 +16,7 @@ use Phalcon\Url\Exception;
 use Phalcon\Mvc\RouterInterface;
 use Phalcon\Mvc\Router\RouteInterface;
 use Phalcon\Di\InjectionAwareInterface;
+use Phalcon\Helper\Str;
 
 /**
  * This components helps in the generation of: URIs, URLs and Paths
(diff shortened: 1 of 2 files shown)


niden (dismissed):

phalcon/Url.zep
```diff
@@ -16,6 +16,7 @@ use Phalcon\Url\Exception;
 use Phalcon\Mvc\RouterInterface;
 use Phalcon\Mvc\Router\RouteInterface;
 use Phalcon\Di\InjectionAwareInterface;
+use Phalcon\Helper\Str;
```
ruudboon: Don't think we need this one if we use preg_replace

ruudboon (changes requested):

phalcon/Url.zep
scrnjakovic: You are right.

ruudboon (approved):

PR #14333, opened 2019-08-27
Fixes #13919, make response->setFileToSend support non-ASCII filename.

### LINKED ISSUE #13919: Downloads are broken because of semicolon appended to filename
Changes from #13496  breaks our all downloads. Downloaded file has semicolon appended  to the extension which disable file type recognition by system (which is important). Tested both linux and windows, chrome and firefox. Even [RFC 6266  Content-Disposition in HTTP ](https://tools.ietf.org/html/rfc6266) omits semicolon at the end of header. Browser does not strip this. Tried to wrap filename in double quotes but with no luck. Only solution is to overwrite Content-Disposition header to remove semicolon from the end.

```php
$this->response->setFileToSend($fileDir, $filename);

//fix
$this->response->setHeader("Content-Disposition", "attachment; filename=$filename");
```

### Details

* Phalcon version:  3.4.2
* PHP Version: 7.2.15
* Operating System: Debian
* Server:  Apache

ian4hu:
- see https://bugs.php.net/bug.php?id=37738
 - see https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21File%21FileSystem.php/function/FileSystem%3A%3Abasename/8.2.x
 - see https://tools.ietf.org/html/rfc2231#section-5

Hello!

* Type: bug fix
* Link to issue: #13919 

Small description of change:
Add non-ASCII support `basename()` by `Phalcon\Helper\Fs` and fixed the #13919 

Thanks


ruudboon:
Thank you for the pull. We're not execepting pulls on the master. Can you sent into the 4.0.x branch and would it be possible to add some unit tests?

Jurigag:
We don't accept PRs to master, can you send it to 4.0.x?

ian4hu:
> Thank you for the pull. We're not accepting pulls on the master. Can you sent into the 4.0.x branch and would it be possible to add some unit tests?

I am not sure about how to make a right unit test, because the it only affect the response header.
The `Phalcon\Helper\Fs->basename` could be unit tested, and I will try to make the unit tests.

phalcon/Helper/Fs.zep
```diff
@@ -0,0 +1,42 @@
...
+     * @param string $uri
+     * @param string  $suffix
+     *
+     * @return string
+     */
+    final public static function basename(string! uri, string suffix = null) -> string
+    {
+    	var filename;
+    	var separators;
+    	var matches;
```
sergeyklay:
```diff
-    	var filename;
-    	var separators;
-    	var matches;
+    	var filename, separators, matches;
```

phalcon/Helper/Fs.zep
```diff
@@ -0,0 +1,42 @@
...
+    	let separators = DIRECTORY_SEPARATOR;
```
sergeyklay: Why do you need the dedicated variable to store constant value?

phalcon/Helper/Fs.zep
```diff
@@ -0,0 +1,42 @@
...
+        let uri = rtrim(uri, separators);
```
sergeyklay:
```suggestion
        let uri = rtrim(uri, DIRECTORY_SEPARATOR);
```

phalcon/Http/Response.zep
```diff
@@ -546,18 +547,24 @@ class Response implements ResponseInterface, InjectionAwareInterface, EventsAwar
...
         if typeof attachmentName != "string" {
-            let basePath = basename(filePath);
+            let basePath = Fs::basename(filePath);
         } else {
             let basePath = attachmentName;
         }
-
         if attachment {
+            let basePathEncoded = rawurlencode(basePath);
+            let basePathEncoding = mb_detect_encoding(basePath, mb_detect_order(), true);
```
sergeyklay:
```suggestion
            // mbstring is a non-default extension
            if function_exists("mb_detect_encoding") {
                let basePathEncoding = mb_detect_encoding(
                    basePath,
                    mb_detect_order(),
                    true
                );
            }                
```

phalcon/Http/Response.zep
```diff
@@ -546,18 +547,24 @@ class Response implements ResponseInterface, InjectionAwareInterface, EventsAwar
     public function setFileToSend(string filePath, attachmentName = null, attachment = true) -> <ResponseInterface>
     {
         var basePath;
+        var basePathEncoding;
```
sergeyklay:
```suggestion
        var basePathEncoding = "ASCII";
```

sergeyklay (changes requested):

phalcon/Helper/Fs.zep
ian4hu: because it is shorter than constant

tests/unit/Helper/Fs/FsBasenameCest.php
```diff
@@ -0,0 +1,71 @@
...
+     * Tests Phalcon\Helper\Fs :: basename()
+     * with non-ASCII $uri support
+     * 
+     * @author Ian Hu <hu2008yinxiang@163.com>
+     * @since  2019-08-27
+     */
+    public function helperFsBasenameNonASCII(UnitTester $I) {
+    	$I->wantToTest('Helper\Fs - basename() with non-ASCII uri');
+    	$filePathAndExpects = [
+    		'/file/热爱中文.txt' => '热爱中文.txt',
```
ruudboon: We do love chinese!

ruudboon (approved):

ruudboon (changes requested):
Can you run phpcs to fix the code style? See failing test
https://travis-ci.org/phalcon/cphalcon/jobs/577361625

tests/unit/Helper/Fs/FsBasenameCest.php
ian4hu: Ditto

phalcon/Http/Response.zep
```diff
@@ -546,18 +547,31 @@ class Response implements ResponseInterface, InjectionAwareInterface, EventsAwar
...
+                );
+            }                
             this->setRawHeader("Content-Description: File Transfer");
             this->setRawHeader("Content-Type: application/octet-stream");
-            this->setRawHeader("Content-Disposition: attachment; filename=" . basePath . ";");
+            this->setRawHeader("Content-Disposition: attachment; filename=" . basePathEncoded . ";");
             this->setRawHeader("Content-Transfer-Encoding: binary");
+            if basePathEncoding != "ASCII" {
+                this->setRawHeader("Content-Disposition: attachment; filename=" . basePathEncoded . "; filename*=". strtolower(basePathEncoding) . "''" . basePathEncoded . ";");
+            }
```
sergeyklay: I would like to avoid duplicate `Content-Disposition` headers. It may cause  an error on Chrome, and an empty page in Firefox. This can be achieved with `Headers::set("Content-Disposition", value)` (instead of using `setRaw`) or:
```zep
if basePathEncoding != "ASCII" {
    // ...
} else {
    this->setRawHeader("Content-Disposition: attachment; filename=" . basePathEncoded . ";");
}
```

sergeyklay (changes requested):

StudioMaX:
Take note that the semicolon at the end of the header is still not allowed according to http://test.greenbytes.de/tech/tc2231/#attwithasciifilenamenqs.
> The trailing semicolon makes the header field value syntactically incorrect, as no other parameter follows. Thus the header field should be ignored.

phalcon/Http/Response.zep
ruudboon: Based on the comment of @StudioMaX I think this should look like this:
```
if basePathEncoding != "ASCII" {
    // ...
} else {
    this->setRawHeader("Content-Disposition: attachment; filename='" . basePathEncoded . "'");
}
```
Removing trailing ';' and quotes around filename (in case of space)

StudioMaX:
@ian4hu could you test these changes with the all common browsers with ASCII/non-ASCII filenames?

I may be wrong, but the file name should still be in double quotes, as the file name may contain spaces (see http://test.greenbytes.de/tech/tc2231/#attwithasciifilenamenqws).

Also, filename in `filename=` shouldn't be urlencoded, but `filename*=UTF-8''` should (see http://test.greenbytes.de/tech/tc2231/#attwithfnrawpctenca).

Maybe something like this:
```php
// basePathEncoded is not rawurlencoded
if basePathEncoding != "ASCII" {
    this->setRawHeader("Content-Disposition: attachment; filename=\"" . basePathEncoded . "\"; filename*=". strtolower(basePathEncoding) . "''" . rawurlencode(basePathEncoded));
} else {
    this->setRawHeader("Content-Disposition: attachment; filename=\"" . basePathEncoded . "\"");
}
```

sergeyklay (approved):

sergeyklay:
Thank you

ian4hu:
> @ian4hu could you test these changes with the all common browsers with ASCII/non-ASCII filenames?
> 
> I may be wrong, but the file name should still be in double quotes, as the file name may contain spaces (see http://test.greenbytes.de/tech/tc2231/#attwithasciifilenamenqws).
> 
> Also, filename in `filename=` shouldn't be urlencoded, but `filename*=UTF-8''` should (see http://test.greenbytes.de/tech/tc2231/#attwithfnrawpctenca).
> 
> Maybe something like this:
> 
> ```
> if basePathEncoding != "ASCII" {
>     this->setRawHeader("Content-Disposition: attachment; filename=\"" . basePath . "\"; filename*=". strtolower(basePathEncoding) . "''" . basePathEncoded);
> } else {
>     this->setRawHeader("Content-Disposition: attachment; filename=\"" . basePath . "\"");
> }
> ```

I could not make a complete tests in http://test.greenbytes.de/tech/tc2231/#attwithasciifilenamenqws,  I only have the main stream browser sets to test, which including latest Chrome, Firefox, Safari, IE or Edge, and because I am busy at work recently, I don't have time to test so I just make the header to follow [RFC2231#Section 5](https://tools.ietf.org/html/rfc2231#section-5).

IMO,  the RFC is the standard specific between browser and server. if I make the framework with browser specific behavior, then if the browser changed its behavior your framework may not work with it again.

ian4hu:
> @ian4hu could you test these changes with the all common browsers with ASCII/non-ASCII filenames?
> 
> I may be wrong, but the file name should still be in double quotes, as the file name may contain spaces (see http://test.greenbytes.de/tech/tc2231/#attwithasciifilenamenqws).
> 
> Also, filename in `filename=` shouldn't be urlencoded, but `filename*=UTF-8''` should (see http://test.greenbytes.de/tech/tc2231/#attwithfnrawpctenca).
> 
> Maybe something like this:
> 
> ```
> if basePathEncoding != "ASCII" {
>     this->setRawHeader("Content-Disposition: attachment; filename=\"" . basePath . "\"; filename*=". strtolower(basePathEncoding) . "''" . basePathEncoded);
> } else {
>     this->setRawHeader("Content-Disposition: attachment; filename=\"" . basePath . "\"");
> }
> ```

Your suggesstion about quoting is right. as the RFC mentioned:
image

I will re-read the RFC and make some changes later. thanks.

PR #14339, opened 2019-08-28
Adding Environment Handler And PhpDotEnv Adapter & Testing

Fenikkusu:
Hello!

* Type: new feature

Small description of change:

I will openly admit that I do not like Laravel for many different reasons. However, I must admit that there are certain 'features' of it that are typically handy. This PR is to add one such feature to Phalcon. Laravel has a general function called `env` that is used to retrieve env vars via getenv,  with defaults, and format them appropriately. Phalcon doesn't have an equivalent. 

This PR adds two new classes. The first is a base Environment class. The second is an Adapter for the ever popular PhpDotEnv from vlucas. It also updates the Factories to include the base environment by default.

Thanks

### DIFF
--- a/phalcon/Environment.zep
+++ b/phalcon/Environment.zep
@@ -1,0 +1,141 @@
+namespace Phalcon;
+
+use Exception;
+use Phalcon\Di\Injectable;
+use RuntimeException;
+
+/**
+ * Environment Class To Retrieve Environment Variables (IE: getenv).
+ *
+ */
+class Environment extends Injectable
+{
+    /** @var array */
+    private envVars = [];
+
+    /** @var bool */
+    private _isLoaded = false;
+
+    /**
+    * Has environment been loaded?
+    */
+    protected function isLoaded() -> bool
+    {
+        return this->_isLoaded;
+    }
+
+    /**
+     * Disables Debug Info
+     */
+    public function __debugInfo()
+    {
+        return [];
+    }
+
+    /**
+     * Set Environment Variable Override
+     *
+     * @param string envName Environment Variable To Set
+     * @param string envValue Environment Variable Value
+     */
+    public function setValue(string envName, var envValue) -> <Environment>
+    {
+        let this->envVars[envName] = envValue;
+
+        return this;
+    }
+
+    /**
+     * Retrive Environemnt Variable
+     *
+     * @param string envName ENV To Retrieve
+     * @param mixed defaultValue Default Value If ENV Does Not Exist
+     */
+    public function getValue(string envName, var defaultValue = null) -> var
+    {
+        //Lazy Load Environment
+        if (!this->isLoaded()) {
+            this->loadEnvironment();
+        }
+
+        var value = this->retrieveValue(envName);
+        if unlikely value === null {
+            return defaultValue;
+        }
+
+        return this->parseValue(value);
+    }
+
+    /**
+     * Retrieve Value.
+     *
+     * @param string envName ENV To Retrieve
+     * @private
+     * @return mixed
+     */
+    protected function retrieveValue(string envName) -> var
+    {
+        var value = getenv(envName);
+        if value !== false {
+            return value;
+        } elseif isset this->envVars[envName] {
+            return this->envVars[envName];
+        }
+
+        let value = null;
+
+        return value;
+    }
+
+    /**
+     * Parses ENV To Scalar Type If Possible
+     *
+     * @param string Value To Parse
+     * @return mixed
+     */
+    private function parseValue(var value)
+    {
+        //Strip Off Wrapping Quotes
+        if unlikely strlen(value) > 1 && substr(value, 0, 1) === "\"" && substr(value, -1) === "\"" {
+            let value = substr(value, 1, -1);
+        }
+
+        //Return Literal Numbers
+        if is_numeric(value) {
+            //Cannot Do Strict Comparison
+            if value != (int) value || strpos(value, '.') !== false {
+                return (float) value;
+            }
+
+            return (int) value;
+        }
+
+        //Convert Specific Types
+        switch strtolower(value) {
+            case "true":
+            case "(true)":
+                return true;
+            case "false":
+            case "(false)":
+                return false;
+            case "empty":
+            case "(empty)":
+                return "";
+            case "null":
+            case "(null)":
+                return;
+        }
+
+        return value;
+    }
+
+    /**
+     * Loads Project Overrides
+     */
+    protected function loadEnvironment() -> <Environment>
+    {
+        let this->_isLoaded = true;
+
+        return this;
+    }
+}
--- a/phalcon/Environment.zep
+++ b/phalcon/Environment.zep
@@ -43,15 +43,7 @@ class Environment extends Injectable
     private envVars = [];
 
     /** @var bool */
-    private _isLoaded = false;
-
-    /**
-    * Has environment been loaded?
-    */
-    protected function isLoaded() -> bool
-    {
-        return this->_isLoaded;
-    }
+    private wasLoaded = false;
 
     /**
      * Disables Debug Info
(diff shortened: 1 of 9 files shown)


phalcon/Environment.zep
```diff
@@ -0,0 +1,170 @@
+namespace Phalcon;
```
sergeyklay: Can we move this to `Phalcon\Di` ?

phalcon/Environment.zep
```diff
@@ -0,0 +1,170 @@
...
+ * A=2.0 => float (2.0)
+ * A=empty => string ("")
+ * A=(empty) => string ("")
+ * A=null => null
+ * A=(null) => null
+ * A="something" => string ("something")
+ * A=else => string ("else")
+ *
+ *
+ *```php
```
sergeyklay: As I know we don use markdown syntax inside phpdoc. 

/cc @niden

phalcon/Environment.zep
```diff
@@ -0,0 +1,170 @@
...
+ *
+ *```
+ */
+class Environment extends Injectable
+{
+    /** @var array */
+    private envVars = [];
+
+    /** @var bool */
+    private _isLoaded = false;
```
sergeyklay: [PSR-2: Coding Style Guide](https://www.php-fig.org/psr/psr-2/#42-properties):
> Property names SHOULD NOT be prefixed with a single underscore to indicate protected or private visibility.

sergeyklay (changes requested):

niden (changes requested):
This is really good work. Thank you.

One question. Where is the `env` function?

phalcon/Environment/Adapter/PhpDotEnv.zep
```diff
@@ -0,0 +1,60 @@
+namespace Phalcon\Environment\Adapter;
+
+use DotEnv\DotEnv;
```
ruudboon: I'm wondering if this dependency could cause some issues. If you want to use this adapter you need to have composer installed, vlucas/phpdotenv installed and setup the autoloader to use the vendor dir.
I can imagine use cases (for example cronjobs) where you would like to use this service but prefer not to have composer etc configured. 
Would it be possible to replace DotEnv with our own implementation or is this too much work?

ruudboon:
Great idea in general to add this!

sergeyklay:
@niden
> Yeah that is what I thought. The issue is we cannot set functions in Phalcon the way Laravel does it or any other PHP based framework or code. Maybe we can set it up as a namespaced static function in the `Helpers` namespace.

In general I don't think we should introduce functions in _global_ namespace like other framework does. The same for tons of static methods. Also note this code should work perfectly:

```zep
namespace Phalcon;

function env () { // Phalcon\env()
    // ...
}

class Env { // Phalcon\Env
    // ...
}
```

Fenikkusu:
I've made the requested changes. The only one I haven't done yet is the change of the namespace. I wasn't sure the final thoughts on that. I'm fine with moving it to Phalcon\Config. I'd rather leave it in the base than put it in Di.

phalcon/Environment/Adapter/PhpDotEnv.zep
ruudboon: I think the task is to big too rewrite this in Phalcon now but would love to see this in the feature so we don't depend on the package and only extend on php extensions.
Should we add this as recommended to the composer.json?

sergeyklay: @ruudboon Phalcon uses composer.json only for testing purposes. Usually users installs `phalcon.so` and don't use Phalcon's `composer.json`.

cnyyk:
Should make phalcon framework clean.

niden:
@Fenikkusu We have tagged this for 4.1. I honestly have very little time right now to fully review this, although it seems pretty straight forward.

I need to also wrap my head around the need for this i.e. _if we are going to use `DotEnv` why do we need this component_.  It seems that the Environment class does what we need in a rudimentary manner but not sure 100%, I just have to pull the code and experiment a bit.

Stay tuned on this, we will address it right after we release v4.

Thank you for the hard work.

CameronHall:
I just wanted to throw in my 2c; I'm share very little interest in having runtime packages as dependencies for Phalcon and requiring `vlucas/dotenv` as an external dependency doesn't fit well with the rest of Phalcon. 
Considering the only thing it seems to offer (because you're re-implemented everything else) is parsing of `.env` files, it seems most appropriate that we rewrite the .env parser in Zephir to align with the rest of the components in the framework. I've spent a considerable amount of time on the `Phalcon\Config` component, so I'm more than happy to assist you with that if need be.

Fenikkusu:
> _if we are going to use `DotEnv` why do we need this component_. 

This kind of goes back to what I was saying in one of my earlier comments. If you look at DotEnv's repo as of roughly back in April I believe, you will see that DotEnv even said 'not intended for production use'. The current best practices state is that you should configure your environment through your server configuration. It is my belief that DotEnv goes against this best practice. Without going too in-depth into this argument (one I've had pretty heated battle about with a co-worker unfortunately), it stands to reason we still have two primary methods of configuring the environment: DotEnv for local development, and the base env for everything else.

> Phalcon and requiring vlucas/dotenv as an external dependency doesn't fit well with the rest of Phalcon

Not so sure I agree. I say this for two reasons. First, consider the Db Adapters. If what you said stood true, why is it Phalcon is able to support MySQL, Oracle, NoSql, Etc, right out of the gate. MySQL, the most common, does require the extension to be installed. I would assume the main reason is that these are the most common ones. In the same light, and the second reason, is that `vlucas/dotenv` is pretty much the most common Env Loader in existence right now, being used by everything. 

--

With that all said, it really doesn't matter to me. I just like the idea of relying on Phalcon to do Phalcon things. Without this, I have to add extra logic to the composer.json to init the project in development mode. At the least, I could see a php Equivalent In the Incubator, but I've never really liked using the Incubator. It bothers me sometimes that many of the features there haven't been incorporated into the framework. I feel like the incubator should be converted to Zephir if I'm honest.

For the most part, I had two goals. The first was to incorporate the base Environment piece. The second was that I wanted to incorporate at least one extended class. `DotEnv` was the easiest to implement. Alternative ones would be a `Encrypted` one, where the values get decrypted on the fly. I could see others, like one that loads from the DB, etc. This component has the potential to expand quite a bit beyond the base `Environment` class.

Fenikkusu:
> @Fenikkusu We have tagged this for 4.1. I honestly have very little time right now to fully review this, although it seems pretty straight forward.

You're fine, I understand.

CameronHall:
> DotEnv even said 'not intended for production use'.

I 100% agree with this statement. The counter side of the argument is that if your .env file is compromised; then it's likely the same goes for the rest of your file system and you have bigger problems than just your secrets being leaked. I personally prefer the Symfony component instead because of the way it handles multiple .env files - I think you can do this with `vlucas/dotenv` but I found it less plug and play to achieve what I wanted.

> Not so sure I agree. I say this for two reasons. First, consider the Db Adapters. If what you said stood true, why is it Phalcon is able to support MySQL, Oracle, NoSql, Etc, right out of the gate. MySQL, the most common, does require the extension to be installed. I would assume the main reason is that these are the most common ones. In the same light, and the second reason, is that `vlucas/dotenv` is pretty much the most common Env Loader in existence right now, being used by everything.

It's not so much about what's common and what's not. It's more that, Phalcon comes with everything "out of the box" and just leverages what's already provided by PHP natively. You won't find any other part of Phalcon that will require you to load a runtime package, I'm also not a fan of forcing the use of a particular package when there's other alternatives such as the Symfony one I mentioned earlier.

> I've never really liked using the Incubator. It bothers me sometimes that many of the features there haven't been incorporated into the framework. I feel like the incubator should be converted to Zephir if I'm honest.

I can understand that from your perspective, and for some part I wholeheartedly agree. The justifications behind it though I'd probably argue are a little old-fashioned now if you will - bearing in mind most of the components are there because they don't fall in line or contribute to the performance mantra of the Phalcon framework. On the other hand, the benefits of it being in PHP are that we can optionally add support or interfaces for things that do extend runtime packages without any ramifications (and release more frequently). But you're right, there are components of the Incubator that should probably be moved to the framework.

---

Besides the above; I genuinely appreciate this pull-request as it is a really good idea and something that I believe Phalcon should have had a long time ago. As you mentioned, the base class is pretty powerful in it's offering and I'm looking forward to seeing how we can take advantage of that in the future.

Fenikkusu:
Rebased on 4.1.x, Edited PR To Point to 4.1.x

zsilbi:
Thank you @Fenikkusu but I have to close this for now as we won't be able to include this in 4.1

Issue #14367, opened 2019-09-08
Static methods in Generate API


niden:
The `generate-api-docs.php in `tests/_ci/` does not take into account `static` methods.

We need to adjust the query to print those out also.

Example `Phalcon\Helper` class

niden:
cc @Jeckerson

PR #14369, opened 2019-09-08
#14367 - Fix Static methods in Generate API

Jeckerson:
Hello!

* Type: bug fix | documentation
* Link to issue: #14367


Jeckerson:
```php
public function delete( mixed $key ): bool;
```
Delete an item from the cache by its unique key.


@return bool True if the item was successfully removed. False if there was an error.

@throws Phalcon\Cache\Exception\InvalidArgumentException
  MUST be thrown if the $key string is not a legal value.

niden (dismissed):

niden (approved):

Issue #14376, opened 2019-09-10
Model::hasChanged() and getChangedFields() return false values


zsilbi:
When snapshot data is saved to a `Model`, all attributes are being saved as `string`.
There can be issues regardless `orm.cast_on_hydrate` is turned on or off. 

Here is a quick example:
```php
class Robots extends \Phalcon\Mvc\Model
{
    public $id; // unsigned int(10) in MySQL
    public $year; // unsigned int(10) in MySQL

    public function initialize()
    {
        $this->keepSnapshots(true);
    }
}

/****************************************/
ini_set('phalcon.orm.cast_on_hydrate', 0);

$robot = Robots::findFirst();

var_dump($robot->toArray()); // array(2) { ["id"] => string(1) "1" ["year"] => string(4) "1900" }

var_dump($robot->year); // string(4) "1900"

var_dump($robot->getSnapshotData()['year']); // string(4) "1900"

var_dump($robot->hasChanged('year')); // bool(false)

$robot->year = 1900;
var_dump($robot->hasChanged('year')); // bool(true)

$robot->year = "1900";
var_dump($robot->hasChanged('year')); // bool(false)

/****************************************/
ini_set('phalcon.orm.cast_on_hydrate', 1);

$robot = Robots::findFirst();

var_dump($robot->toArray()); // array(2) { ["id"] => int(1) ["year"] => int(1900) }

var_dump($robot->year); // int(1900)

var_dump($robot->getSnapshotData()['year']); // string(4) "1900"

var_dump($robot->hasChanged('year')); // bool(true)

$robot->year = 1900;
var_dump($robot->hasChanged('year')); // bool(true)

$robot->year = "1900";
var_dump($robot->hasChanged('year')); // bool(false)
```

When `orm.cast_on_hydrate` is on, I think `Model::cloneResultMap()` should save the casted values to the snapshot data.
If it's off, then `Model::getChangedFields()` should not use `===` to check if the field has changed.

Any thoughts?

### Details
* Phalcon version: 4.0.x branch

zsilbi:
Resolved in #14377

PR #14377, opened 2019-09-10
Mvc\Model::snapshot data cast

zsilbi:
Hello!

* Type: bug fix
* Link to issue: #14376

Small description of change:

When `orm.cast_on_hydrate` is on, I think `Model::cloneResultMap()` should save the casted values to the snapshot data.
If it's off, then `Model::getChangedFields()` should not use `===` to check if the field has changed.

Thanks,
zsilbi

### DIFF
--- a/phalcon/Mvc/Model.zep
+++ b/phalcon/Mvc/Model.zep
@@ -1511,6 +1511,7 @@ abstract class Model extends AbstractInjectionAware implements EntityInterface, 
     public function getChangedFields() -> array
     {
         var metaData, name, snapshot, columnMap, allAttributes, value;
+        bool hasChanged;
         array changed;
 
         let snapshot = this->snapshot;
@@ -1564,6 +1565,12 @@ abstract class Model extends AbstractInjectionAware implements EntityInterface, 
                 let changed[] = name;
 
                 continue;
+            }
+
+            if globals_get("orm.cast_on_hydrate") {
+                let hasChanged = value != snapshot[name];
+            } else {
+                let hasChanged = value !== snapshot[name];
             }
 
             /**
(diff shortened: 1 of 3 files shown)


phalcon/Mvc/Model.zep
```diff
@@ -1570,9 +1570,9 @@ abstract class Model extends AbstractInjectionAware implements EntityInterface,
             }
 
             if globals_get("orm.cast_on_hydrate") {
-                let hasChanged = value != snapshot[name];
-            } else {
                 let hasChanged = value !== snapshot[name];
+            } else {
+                let hasChanged = value != snapshot[name];
```
zsilbi: This fix conflicts with previous issues: 

https://github.com/phalcon/cphalcon/pull/12655
https://forum.phalcon.io/discussion/15529/bug-regarding-getchangedfields-and-haschanged#C45969

sergeyklay: Could you please correct the test?

zsilbi: The purpose of this test is to check if the `Model` detects the difference between `null` and `0`.
We could keep this because the snapshot data stores `null` correctly with both `cast_on_hydrate` settings.

If either the actual or the snapshot value is `null`, we could use the `!==`

Jeckerson: https://travis-ci.org/phalcon/cphalcon/jobs/583222050#L601

zsilbi: > https://travis-ci.org/phalcon/cphalcon/jobs/583222050#L601

@Jeckerson, yes, I was talking about this test:

```php
$I->assertNotEmpty([]) at tests/integration/Mvc/Model/Refactor-SnapshotCest.php:370
```

This failed because it couldn't tell the difference between `deleted = 0` and `deleted = null` (`orm.cast_on_hydrate` was off)

ruudboon: @zsilbi I've discussed this with the team. Can you keep the compare as it is? If you cast your model you should set your properties as the required type otherwise we expect a hasChanged to be true.

zsilbi: Thanks @ruudboon. I removed the comparison and the test for it, and rebased.

ruudboon (approved):

PR #14444, opened 2019-10-03
Fixed $model->countRelated() failing when the relation is reusable

zsilbi:
Hello!

* Type: bug fix

Small description of change:

The `uniqueKey` used to store reusable related records in `ModelsManager` is currently generated only from the conditions used to lookup the records.

For example:
It uses the same key for `$robot->getParts(["some" = "thing"])` and `$robot->countParts(["some" = "thing"])`. 
If the related records are accessed again using a different method, it will return the result of the first.
In this case, it will be a `Resultset` instead of an `integer`.

I added the `retrieveMethod` variable to the `uniqueKey` generation.
Added some tests for `Model::__call()`

Thanks,
zsilbi

### DIFF
--- a/phalcon/Mvc/Model.zep
+++ b/phalcon/Mvc/Model.zep
@@ -199,7 +199,7 @@ abstract class Model extends AbstractInjectionAware implements EntityInterface, 
          */
         let records = this->_getRelatedRecords(modelName, method, arguments);
 
-        if records !== null {
+        if records !== false {
             return records;
         }
 
@@ -4080,7 +4080,7 @@ abstract class Model extends AbstractInjectionAware implements EntityInterface, 
              * Return if the relation was not found because getRelated() throws an exception if the relation is unknown
              */
             if typeof relation != "object" {
-                return null;
+                return false;
             }
 
             return this->getRelated(alias, extraArgs);
@@ -4101,7 +4101,7 @@ abstract class Model extends AbstractInjectionAware implements EntityInterface, 
              * If the relation was found perform the query via the models manager
              */
             if typeof relation != "object" {
-                return null;
+                return false;
             }
 
             return manager->getRelationRecords(
@@ -4112,7 +4112,7 @@ abstract class Model extends AbstractInjectionAware implements EntityInterface, 
             );
         }
 
-        return null;
+        return false;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


phalcon/Mvc/Model.zep
```diff
@@ -188,7 +188,7 @@ abstract class Model extends AbstractInjectionAware implements EntityInterface,
 
         let records = self::_invokeFinder(method, arguments);
 
-        if records !== null {
```
sergeyklay: IMO Is more suitable (and safe for future `_invokeFinder` changes):

```zep
if !empty records
```

isn't?

phalcon/Mvc/Model.zep
```diff
@@ -199,7 +199,7 @@ abstract class Model extends AbstractInjectionAware implements EntityInterface,
          */
         let records = this->_getRelatedRecords(modelName, method, arguments);
 
-        if records !== null {
```
sergeyklay: The same here

phalcon/Mvc/Model.zep
zsilbi: Unfortunately no.

This is from a test which I wrote to see if non existent related records are handled correctly using `Model::__call()`.
```php
$nonExistentPart = $robotPart->getPart(
            [
                'id < 0',
                'order' => 'id DESC',
            ]
        );
```
If you use `!== null` or `empty` and no records are found, it will throw an exception that the method doesn't exist.

`[Phalcon\Mvc\Model\Exception] The method 'getPart' doesn't exist on model 'Phalcon\Test\Models\RobotsParts' `

phalcon/Mvc/Model.zep
zsilbi: This is from a test which I wrote to see if non existent related records are handled correctly using `Model::__call()`.
```php
$nonExistentPart = $robotPart->getPart(
            [
                'id < 0',
                'order' => 'id DESC',
            ]
        );
```
If you use `!== null` or `empty` and no records are found, it will throw an exception that the method doesn't exist.

`[Phalcon\Mvc\Model\Exception] The method 'getPart' doesn't exist on model 'Phalcon\Test\Models\RobotsParts' `

zsilbi:
@sergeyklay I'd also change `Model::__callStatic()` to throw an exception if the method is unknown, like `Modell::__call()` does.

What do you think?

sergeyklay:
@zsilbi Let's use a separated PR to change Model::__callStatic()

sergeyklay (approved):

ruudboon (approved):

niden:
Thank you @zsilbi

PR #14453, opened 2019-10-06
Test improvements

ruudboon:
Hello!

* Type: code quality
* Link to issue:

Improved getServerAddr test.
Implemented Cli Task tests

### DIFF
--- a/tests/cli/Cli/Task/ConstructCest.php
+++ b/tests/cli/Cli/Task/ConstructCest.php
@@ -15,6 +15,8 @@ namespace Phalcon\Test\Cli\Cli\Task;
 use CliTester;
 use OnConstructTask;
 use Phalcon\Test\Fixtures\Traits\DiTrait;
+use Phalcon\Cli\Task;
+use MainTask;
 
 class ConstructCest
 {
@@ -34,8 +36,9 @@ class ConstructCest
     public function cliTaskConstruct(CliTester $I)
     {
         $I->wantToTest('Cli\Task - __construct()');
-
-        $I->skipTest('Need implementation');
+        require_once dataDir('fixtures/tasks/MainTask.php');
+        $task = new MainTask();
+        $I->assertInstanceOf(Task::class, $task);
     }
 
     public function testOnConstruct(CliTester $I)
--- a/tests/cli/Cli/Task/GetEventsManagerCest.php
+++ b/tests/cli/Cli/Task/GetEventsManagerCest.php
@@ -13,6 +13,8 @@ declare(strict_types=1);
 namespace Phalcon\Test\Cli\Cli\Task;
 
 use CliTester;
+use Phalcon\Events\Manager;
+use Phalcon\Cli\Task;
 
 class GetEventsManagerCest
 {
@@ -25,7 +27,9 @@ class GetEventsManagerCest
     public function cliTaskGetEventsManager(CliTester $I)
     {
         $I->wantToTest('Cli\Task - getEventsManager()');
-
-        $I->skipTest('Need implementation');
+        $eventsManager = new Manager();
+        $task = new Task();
+        $task->setEventsManager($eventsManager);
+        $I->assertEquals($eventsManager, $task->getEventsManager());
     }
 }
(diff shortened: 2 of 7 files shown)


tests/cli/Cli/Task/ConstructCest.php
```diff
@@ -34,8 +36,9 @@ public function _before(CliTester $I)
     public function cliTaskConstruct(CliTester $I)
     {
         $I->wantToTest('Cli\Task - __construct()');
-
-        $I->skipTest('Need implementation');
+        require_once dataDir('fixtures/tasks/MainTask.php');
+        $task = new MainTask();
+        $I->assertInstanceOf(Task::class, $task);
```
sergeyklay: I have no idea what exactly we're test here. Just a PHP's ability to "require" a file or execute a function called __constructor? And why "require_once"? Shouldn't we use composer autoload policy? IMO would be better to test this
https://github.com/phalcon/cphalcon/blob/762184777a6dc425787ffee531a7b6c8a4b2532d/phalcon/Cli/Task.zep#L46-L48

sergeyklay (changes requested):

tests/cli/Cli/Task/GetEventsManagerCest.php
```diff
@@ -25,7 +27,9 @@ class GetEventsManagerCest
     public function cliTaskGetEventsManager(CliTester $I)
     {
         $I->wantToTest('Cli\Task - getEventsManager()');
-
-        $I->skipTest('Need implementation');
+        $eventsManager = new Manager();
+        $task = new Task();
+        $task->setEventsManager($eventsManager);
```
sergeyklay:
```
[Error] Call to undefined method Phalcon\Cli\Task::setEventsManager()
```
See: https://github.com/phalcon/cphalcon/pull/14445

tests/cli/Cli/Task/SetEventsManagerCest.php
```diff
@@ -26,6 +28,9 @@ public function cliTaskSetEventsManager(CliTester $I)
     {
         $I->wantToTest('Cli\Task - setEventsManager()');
 
-        $I->skipTest('Need implementation');
+        $eventsManager = new Manager();
+        $task = new Task();
+        $task->setEventsManager($eventsManager);
```
sergeyklay: The same as above

tests/cli/Cli/Task/UnderscoreGetCest.php
```diff
@@ -26,6 +29,15 @@ public function cliTaskUnderscoreGet(CliTester $I)
...
         $I->wantToTest('Cli\Task - __get()');
 
-        $I->skipTest('Need implementation');
+
+        $task = new Task();
+        $di = new CliDi();
+        $task->setDi($di);
+
+        $eventsManager = new Manager();
+        $task->setEventsManager($eventsManager);
```
sergeyklay: the same as above

niden:
I believe the above error relates to this change https://github.com/phalcon/cphalcon/pull/14445

tests/cli/Cli/Task/ConstructCest.php
ruudboon: I only moved this from the original test to proper place. onConstruct is tested below

tests/cli/Cli/Task/GetEventsManagerCest.php
ruudboon: Will rebuild first. Looks like I'm missing this locally

ruudboon: Starting to wonder if we even should have this methods here. Maybe we can just remove the test.

ruudboon:
> I believe the above error relates to this change #14445

Think so to. Wondering if the local events manager should be even part of this class.

tests/cli/Cli/Task/GetEventsManagerCest.php
CameronHall: It makes sense to keep it because we need to have tests for CLI events.

tests/cli/Cli/Task/SetEventsManagerCest.php
CameronHall: I think you can remove this one in favour of the other one.

ruudboon:
Let me close this for now and let me rework the full cli tests with proper namespaces etc

PR #14457, opened 2019-10-11
Fix/cache key 

ekmst:
Hello!

* Type: bug fix
* Link to issue:

Small description of change: added missed `.` to key pattern
[see](https://www.php-fig.org/psr/psr-16/#12-definitions)

Thanks


niden:
@ekmst Thank you for this. If you don't mind update the changelog also.

CameronHall (approved):

niden (approved):

niden:
Thank you @ekmst !

PutinVladimir:
Do not follow blindly PSR.

There told must at least support this chars, but not should be limited this chars and "MAY support additional characters and encodings".
So you limited to cache content of page where page name or page url in national characters as cache key.

Also in PSR told "The following characters are reserved for future extensions and MUST NOT be supported by implementing libraries: {}()/\@:" it is agains" again most of this symbols corectly worked in apcu, memcached, redis and at least ":", "/" needed for whole page caching by url.

PR #14458, opened 2019-10-11
WIP: Update params for deep assoc completion plugin

Jurigag:
Hello!

* Type: new feature | documentation
* Link to issue:https://github.com/phalcon/cphalcon/issues/13991

- [ ] I wrote some tests for this PR - not needed really, needs to check generated stubs
- [x] I have updated the relevant CHANGELOG - not required since it doesn't add any new feature
- [x] I have created a PR for the [documentation](https://github.com/phalcon/docs) about this change - no documentation change required

Small description of change: it adds `@param array config` etc on all methods having things like options, config etc with all keys which are accepted in those arrays, using plugin https://github.com/klesun/deep-assoc-completion for phpstorm it will autocomplete it in all usages of those methods.

Thanks

### DIFF
--- a/phalcon/Annotations/Adapter/Apcu.zep
+++ b/phalcon/Annotations/Adapter/Apcu.zep
@@ -35,6 +35,11 @@ class Apcu extends AbstractAdapter
     protected ttl = 172800;
 
     /**
+     * @param array options = [
+     *     'prefix' => 'phalcon'
+     *     'lifetime' => 3600
+     * ]
+     *
      * Phalcon\Annotations\Adapter\Apcu constructor
      */
     public function __construct(array options = [])
--- a/phalcon/Annotations/Adapter/Stream.zep
+++ b/phalcon/Annotations/Adapter/Stream.zep
@@ -35,6 +35,10 @@ class Stream extends AbstractAdapter
     protected annotationsDir = "./";
 
     /**
+     * @param array options = [
+     *     'annotationsDir' => 'phalconDir'
+     * ]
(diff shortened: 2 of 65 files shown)


niden:
We need to make sure that these comments end up in the stubs with `zephir stubs`

sergeyklay:
@niden The `zephir stubs` command should leave they as is. If there are bugs, we can follow up in separated Zephir issue.

Jurigag:
Yea they are leaving them because regex don't detect it as `@param` currently, but then it still add new `@param mixed config` etc because of that. Need to test if everything will work with this.

Jurigag:
Currently in generated php file it looks like this:

```
/**
@param array config = [
    'test' => 'key'
]
@param mixed $config
*/
```

So we need changes to zephir stubs generation first.

sergeyklay:
@Jurigag As a far as I know @AlexNDRmac working on this by improving `zephir stubs` command. 

Notice that annotations in the following form:
```
@param array config = [
    'test' => 'key'
]
```
are in non standard form and working with this should considered as a feature request. But actually there are more urgent (and honestly critical) things we should do firstly. So working with non-standard annotation isn't a top priority

Jurigag:
Will check it tomorrow.

ruudboon:
@Jurigag Any progress on this one?

ruudboon (approved):

niden:
Thank you @Jurigag 

We need to make sure that these propagate to the stubs.

Issue #14464, opened 2019-10-12
[NFR]: Add casting to Arr::get and Collection::get


niden:
`Arr::get` and `Collection->get` need to have a `cast` parameter to ensure that returned data is casted to whatever the user needs.

niden:
Implemented

PR #14465, opened 2019-10-12
T14464 cast on arr collection

niden:
Hello!

* Type: enhancement
* Link to issue: #14464 

Added `cast` parameter to `Phalcon\Collection::get` and `Phalcon\Helper\Arr::get` allowing developers to cast the result returned to the type they want to.

Thanks

### DIFF
--- a/phalcon/Helper/Arr.zep
+++ b/phalcon/Helper/Arr.zep
@@ -112,29 +112,15 @@ class Arr
         array! collection,
         var index,
         var defaultValue = null,
-        string cast = null
+        string! cast = null
     ) -> var {
         var value;
-        array casts;
 
         if unlikely !fetch value, collection[index] {
             return defaultValue;
         }
 
-        let casts = [
-            "array"   : 1,
-            "bool"    : 1,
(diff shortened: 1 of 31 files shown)


niden:
cc @ruudboon @sergeyklay

phalcon/Helper/Arr.zep
```diff
@@ -108,14 +108,36 @@ class Arr
...
+            "double"  : 1,
+            "float"   : 1,
+            "int"     : 1,
+            "integer" : 1,
+            "null"    : 1,
+            "object"  : 1,
+            "string"  : 1
+        ];
+
+        if unlikely isset casts[cast] {
```
ruudboon: Shouldn't we throw an Exception when the cast value doesn't exist in the array?

ruudboon (changes requested):

sergeyklay (dismissed):

niden:
@ruudboon `settype` will return a notice if the wrong type is passed so we are OK on that.

PR #14504, opened 2019-11-01
Model paginator optimization

ekmst:
Hello!

* Type: new feature | optimization | performance
* Link to issue:

Small description of change:

Rewrote `Paginator\Adapter\Model` using offset.
Since the previous implementation received all the records and paged it at the application level. Below is an example of queries:

```php
use Phalcon\Paginator\Adapter\Model;

$paginator = new Model (
    [
        'data' => Test :: find (
            [
                'columns' => 'id, column, column1'
            ]
        ),
        'limit' => 3,
        'page' => 4,
    ]
);

```

```sql
SELECT `test`.`id`,` test`.`column`, `test`.`column1` FROM` test`
```

And here is an example request using offset:

```php
use Phalcon\Paginator\Adapter\Model;

$paginator = new Model (
    [
        'model' => Test :: class,
        'parameters' => [
            'columns' => 'id, column, column1',
            'limit'  => 12,  // will be ignored
        ],
        'limit' => 3,
        'page' => 4,
    ]
);

```

```sql
SELECT `test`.`id`,` test`.`column`, `test`.`column1` FROM` test` LIMIT: APL0 OFFSET: APL1

-- [APL0] => 3
-- [APL1] => 9
```

But I had to change the `Model` adapter settings. Which breaks backward compatibility with the previous version

What do you think of it?

Thanks


niden:
@ekmst if you are going to break backwards compatibility now is the time to do it before v4 is out.

Go ahead with this. We will also need to update the docs and note the BC change in the upgrade document

phalcon/Paginator/Adapter/Model.zep
```diff
@@ -41,64 +81,39 @@ class Model extends AbstractAdapter
...
+        if rowcount > 0 {
+            let parameters["limit"]  = limit,
+                parameters["offset"] = limit * (pageNumber - 1);
 
-                items->next();
-            }
+            let items = <ResultsetInterface> call_user_func(
+                [modelClass, "find"],
+                parameters
+            );
```
niden: @sergeyklay isnt' 

```
{modelClass}::find(parameters);
```

faster than `call_user_func` ?

sergeyklay: No, in that case call_user_func will be slover

niden (approved):

niden:
Thank you @ekmst

PR #14511, opened 2019-11-04
Model has-one-through relations

zsilbi:
Hello!

* Type: new feature

Small description of change:

This is very similar to  `hasManyToMany`, but it can only use one intermediate row.
The related record is either `ModelInterface`, or `null`.

This is kind of an antipattern, but it can be useful when you have to work with a poorly designed database structure that can't be altered.

Thanks,
zsilbi

### DIFF
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -8,6 +8,14 @@
   - Added `Phalcon\Html\Link\Serializer\Header`
   - Added `Phalcon\Html\Link\Serializer\SerializerInterface`
 - Added `Phalcon\Collection:getKeys` and `Phalcon\Collection\getValues` for getting data from the collection [#14507](https://github.com/phalcon/cphalcon/issues/14507)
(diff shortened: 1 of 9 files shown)


Jeckerson (approved):

sergeyklay (approved):

niden:
Thank you @zsilbi

scrnjakovic:
@zsilbi Could you explain why it's anti-pattern? I genuinely want to know. I thought it's perfectly normal to design database that way if you want to have an optional relationship and normalized database, i.e.

**document**
document_id (PK)
...

**document_approved_by**
document_id (PK)
user_id (FK)

**user**
user_id (PK)

instead of

**document**
document_id (PK)
approved_by (FK) nullable

Thanks and also thanks for PR :)

zsilbi:
> @zsilbi Could you explain why it's anti-pattern?

What I think I really meant there, that performance wise there can be better alternatives to this. 
But I also started to use it in some distinct cases, and I also see the beauty in your example. 😄

Issue #14562, opened 2019-11-30
[BUG]: Stream backend cache 


duongkhoangiam:
Using stream adapter in Model cache
Config setting
```php
'cache' => [
        'adapter' => 'stream',
        'options' => [
            'defaultSerializer' => 'Json',
            'storageDir' => BASE_PATH . '/storage/caches/',
            'lifetime' => 172800,
            'prefix' => 'data-'
        ]
    ],
```

In Provider 
```php
public function register(DiInterface $di): void
    {
        $di->setShared('modelsCache', function () {
            $config = $this->getShared('config');
            $cache = $config->cache->toArray();

            $serializerFactory = new SerializerFactory();
            $adapterFactory    = new AdapterFactory($serializerFactory);

            $adapter = $adapterFactory->newInstance($cache['adapter'], $cache['options']);
            return new Cache($adapter);
        });
    }
```

In Controller
```php
class Post extends BaseController{
    public function index(){
        $results = Model\Post\Post::find([
            'cache' => [
                'key' => 'my-cache',
            ],
        ]);

        var_dump($this->cache->get('my-cache'));exit;
        
    }
}
```

When first run Phalcon created folder 
like 
 

>  storage
>     --caches
>        --data
>           --my
>              ---c
>                   --ac
>                      --my-cache

Second run get error
Cache didn't return a valid resultset

Phalcon 4.0 rc3
Macos Catalina
Mamp Pro
PHP 7.3.7

I am ready test for redis adapter working nice!

ekmst:
I also ran into this error.
I think the problem is here https://github.com/phalcon/cphalcon/blob/858d0e57ed42f0dcd8723cca64798dded84790ea/phalcon/Storage/Serializer/Json.zep#L29-L33

Before encoding in JSON, it is checked for `isSerializable`, but in the `Storage\Serializer\Json::unserialize` method, the data is returned without checking whether it is encoded or not.

https://github.com/phalcon/cphalcon/blob/858d0e57ed42f0dcd8723cca64798dded84790ea/phalcon/Storage/Serializer/Json.zep#L41

niden:
I cannot reproduce this guys. I saw it once in my test and local environment but it was caused by a corrupted cache. Once I cleared the cache (deleted the file) it worked just fine.

This is the full build: https://travis-ci.com/niden/cphalcon/builds/138859860

[...]

PR #14570, opened 2019-12-01
T14562 cache unserialize

niden:
Hello!

* Type: bug fix - new feature
* Link to issue: #14562 

Small description of change:

Thanks

### DIFF
--- a/phalcon/Helper/Json.zep
+++ b/phalcon/Helper/Json.zep
@@ -1,0 +1,109 @@
+
+/**
+ * This file is part of the Phalcon.
+ *
+ * (c) Phalcon Team <team@phalcon.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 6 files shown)


phalcon/Helper/Json.zep
```diff
@@ -0,0 +1,109 @@
...
+use InvalidArgumentException;
+
+/**
+ * This class offers a wrapper for JSON methods to serialize and unserialize
+ */
+class Json
+{
+    /**
+     * Decodes a string using `json_decode` and throws an exception if the
+     * JSON data cannot be decoded
```
sergeyklay: It is default behavior for PHP 7.3:
```php
json_decode($foo, JSON_THROW_ON_ERROR);
```

niden: The options defaults to `0` so you have to specify that flag. Also we support (for now) 7.2 and above so for now this will suffice until we deprecate 7.2

sergeyklay (approved):

PR #14581, opened 2019-12-04
Fix debug var always show TRUE value

mvanvu:
** PR for issue #

The issue: when I use debugVar to dump some variables and get the results like below image.
Selection_007


ruudboon:
@mvanvu Thank you for submitting. We don't accept Pull's on the master. Can you sent it to the 4.0.x branch?

ruudboon:
@mvanvu Can you check recheck your rebase? Looks like a readme is also pushed.

niden (approved):

niden:
Thank you @mvanvu

Issue #13655, opened 2018-12-11
Refactor Unit Tests - Add more tests


niden:
Refactoring unit tests and adding new ones
Tests: 3488, Assertions: 8134, Skipped: 444

#### Refactor
- [ ] Acl\Adapter\MemoryCest.php

#### Acl\Adapter\Memory
- [ ] Acl\Adapter\Memory\DropComponentAccessCest.php

#### Refactor
- [x] Annotations\Adapter\ApcuCest.php
- [ ] Annotations\Adapter\FilesCest.php
- [x] Annotations\Adapter\MemoryCest.php
- [x] Annotations\ReaderCest.php
- [x] Annotations\ReflectionCest.php

#### Annotations\Adapter\Apcu
- [ ] Annotations\Adapter\Apcu\ConstructCest.php
- [ ] Annotations\Adapter\Apcu\GetCest.php
- [ ] Annotations\Adapter\Apcu\GetMethodCest.php
- [ ] Annotations\Adapter\Apcu\GetMethodsCest.php
- [ ] Annotations\Adapter\Apcu\GetPropertiesCest.php
- [ ] Annotations\Adapter\Apcu\GetPropertyCest.php
- [ ] Annotations\Adapter\Apcu\GetReaderCest.php
- [ ] Annotations\Adapter\Apcu\ReadCest.php
- [ ] Annotations\Adapter\Apcu\SetReaderCest.php
- [ ] Annotations\Adapter\Apcu\WriteCest.php

#### Annotations\Adapter\Files
- [ ] Annotations\Adapter\Files\ConstructCest.php
- [ ] Annotations\Adapter\Files\GetCest.php
- [ ] Annotations\Adapter\Files\GetMethodCest.php
- [ ] Annotations\Adapter\Files\GetMethodsCest.php
- [ ] Annotations\Adapter\Files\GetPropertiesCest.php
- [ ] Annotations\Adapter\Files\GetPropertyCest.php
- [ ] Annotations\Adapter\Files\GetReaderCest.php
- [ ] Annotations\Adapter\Files\ReadCest.php
- [ ] Annotations\Adapter\Files\SetReaderCest.php
- [ ] Annotations\Adapter\Files\WriteCest.php

#### Annotations\Adapter\Memory
- [ ] Annotations\Adapter\Memory\GetCest.php
- [ ] Annotations\Adapter\Memory\GetMethodCest.php
- [ ] Annotations\Adapter\Memory\GetMethodsCest.php
[...]

PR #14587, opened 2019-12-05
Add two tests to check addInherit

jenovateurs:
Hello!
I hope the way I wrote these tests is Ok.
The first one is just to check if addInherit doesn't return any error.
The second one is deeper. It veriifies if inherit works well by checking different access.

Type: code quality
Link to issue: #13655

Thanks


niden (dismissed):

niden:
@jenovateurs Looking good. You forgot the `use` statement at the top of the test for the `Memory` adapter.

jenovateurs:
@niden Yes indeed, thanks for your help :-)

niden (approved):

niden:
Thank you @jenovateurs

Issue #11770, opened 2016-05-09
Cookie is coupled with the Session


djfd:
Hi,

synopsis: Cookie class is coupled with session class, it is wrong approach

in full:
let consider next micro-app

``` php
<?php

use Phalcon\Mvc\Router;
use Phalcon\Mvc\Micro;
use Phalcon\DiInterface;
use Phalcon\Di\FactoryDefault;
use Phalcon\Session\Adapter\Files;

class MyApp extends Micro
{
    public function __construct(DiInterface $di = null)
    {
        $di = $di ?: new FactoryDefault();

        // bot detector service mockup
        $di->setShared('iambot', function() {
            return null;
        });

        // only need session if not a bot
        $di->setShared('session', function() {
            $bot = $this->getShared('iambot');
            return $bot ? null : new Files();
        });

        parent::__construct($di);

        $this->router->setUriSource(Router::URI_SOURCE_SERVER_REQUEST_URI);

        $this->get('/{anything}', function($uri) use ($di) {
            $c = $di->getShared('cookies')->get('test');
            echo "got cookie<br>";

            $v = $c->getValue();
            echo "got cookie value `$v`<br>";
            echo "URI is `$uri`";
        });
    }
}

(new MyApp())->handle();
```

The idea is to do not create session for a bots, actually, why to do it.

This app wont work producing error:

```
Uncaught exception 'RuntimeException' with message
'Trying to call method isstarted on a non-object'
Stack trace:
  #0 [internal function]: Phalcon\\Http\\Cookie->restore()
```

see
https://github.com/phalcon/cphalcon/blob/master/phalcon/http/cookie.zep#L139
https://github.com/phalcon/cphalcon/blob/master/phalcon/http/cookie.zep#L300

Also we get infinite loop of session service creation if we have something like this:

``` php
$di->setShared('session', function() {
    // we do use custom cookie for a session bootstrap
    $bootstrap_cookie = $this->getShared('cookies')->get('session-bootstrap-cookie'); // this works
    $session_id = $bootstrap_cookie->getValue(); // this wont work, it is calling Cookie::restore() which tries to take session, and so on
    return new CustomSession($session_id);
});
```

thanks

djfd:
it looks that storing cookie copy in session makes no sense at all. Could you please explain what is rationale behind of doing that?

niden:
[...]

PR #14649, opened 2019-12-23
Removed cookie binding to session

ekmst:
Hello!

* Type: improvement
* Link to issue: https://github.com/phalcon/cphalcon/issues/11770

Small description of change:
Removed cookie binding to session
`Phalcon\Http\Cookie` no longer depends on the session service and data will not be duplicated in the session. This made it difficult to use cookies in stateless applications (SPA).

Thanks


niden:
@ekmst can you rebase and send this to the `4.1.x` branch please?

sergeyklay (changes requested):
I'm not opposed to this change, but I would like to see a more detailed description of these changes. Not just "fixed issue XXX".  What suggested in return? Does it breaks backward compatibility? If so, what are workarounds? Why can't we go further with this? 

Regarding change log: why you put these changes in the "Changed" but not in the "Removed" section? Also move please this to the 4.1.x version in the change log.

niden:
This will definitely break BC not by means of interfaces but by functionality itself. For instance right now I set my cookies and I know they will go in the session. With this change they will not so I will need to write that code myself. 

There are plans to completely rewrite the HTTP layer and offer cookie management with different storage engines i.e. files, session etc. but we are not there yet.

@ruudboon @sergeyklay Thoughts on this? If we are to use this we will need to introduce also some sort of a storage i.e. take the session storage from where it was (see this PR) and put it somewhere else so that people can use it at least for now.

ekmst:
I am currently developing a SPA application where sessions are not used at all. But I need to work with cookies. But since the session is not used, I get errors that the session service was not found. And now I have to use a stub to avoid this.

I do not quite understand why cookies are also stored in sessions?

niden:
@ekmst storing cookies in files or the session are the most common storage engines. 

This code is really _ancient_ and has never seen much love - we just let it be. Definitely needs to be reworked so what you did I would say is step 1 and we need step 2 to at least have the option to store the cookies in the session (current functionality)

I will see what I can come up with in the next few days and will ping you so that we can align efforts.

ekmst:
@niden  
Can we save cookies in the session only if the `session` service is defined in the DI container? If the `session` service is not defined, then we work with cookies and we will not throw an exception.

Or we can enter a flag to explicitly indicate that you want to save cookies in the session.

niden:
> @niden
> Can we save cookies in the session only if the `session` service is defined in the DI container? If the `session` service is not defined, then we work with cookies and we will not throw an exception.
> 
> Or we can enter a flag to explicitly indicate that you want to save cookies in the session.

That is actually a great idea. If we mention this in the documentation it should be fine and most people would not be affected if they already have the `session` service registered.

@ekmst can you make the adjustments? 

The above will definitely help and give us time until we redesign the http layer.

ekmst:
@niden 
Yes, I will try to do it soon.

niden:
@ekmst looking good so far. Once you are done make sure the changelog is updated and ping me so that I can merge this

ekmst:
@niden 

Done

ruudboon:
With these changes I wouldn't mind to push this into 4.0.x. Don't think we're breaking anything

niden:
> With these changes I wouldn't mind to push this into 4.0.x. Don't think we're breaking anything
> Thnx ekmst

We will have to cherry pick them because this is a different branch. We can talk about this in TG.

niden:
Thank you @ekmst

PR #14698, opened 2020-01-11
T3135 best language

niden:
Hello!

* Type: new feature
* Link to issue: #3135 

Added `Phalcon\Http\Request::getBestLanguageIso()` to return the base language if this is a specific one (`en` vs `en-US`) 

Thanks


phalcon/Http/Request.zep
```diff
@@ -146,6 +146,21 @@ class Request extends AbstractInjectionAware implements RequestInterface
         return this->getBestQuality(this->getLanguages(), "language");
     }
 
+    /**
+     * Gets best language accepted by the browser/client from
+     * _SERVER["HTTP_ACCEPT_LANGUAGE"] and returns the base part of it i.e.
+     * `en` instead of `en-US`
+     */
+    public function getBestLanguageIso() -> string
```
ruudboon: I think the method name is a bit confusing. `en-US` is also an iso standard.

tests/unit/Http/Request/GetBestLanguageCest.php
```diff
@@ -27,6 +30,56 @@ public function httpRequestGetBestLanguage(UnitTester $I)
...
         $I->wantToTest('Http\Request - getBestLanguage()');
 
-        $I->skipTest('Need implementation');
+        $store = $_SERVER ?? [];
+
+        $this->setNewFactoryDefault();
+        $request = $this->container->get('request');
+
+        $_SERVER = [
+            'HTTP_ACCEPT_LANGUAGE' => 'es,es-ar;q=0.8,en;q=0.5,en-us;q=0.3,de-de; q=0.9'
```
ruudboon: Can we add some test cases with corrupt values? HTTP_ACCEPT_LANGUAGE can be manipulated by the client.

tests/unit/Http/Request/GetClientAddressCest.php
```diff
@@ -27,6 +30,35 @@ public function httpRequestGetClientAddress(UnitTester $I)
...
         $I->wantToTest('Http\Request - getClientAddress()');
 
-        $I->skipTest('Need implementation');
+        $store = $_SERVER ?? [];
+
+        $this->setNewFactoryDefault();
+        $request = $this->container->get('request');
+
+        $_SERVER = [
+            'REMOTE_ADDR'          => '192.168.0.1',
```
ruudboon: Can we add IPv6 as well?

phalcon/Http/Request.zep
sergeyklay:
```suggestion
    public function getPreferredLocaleVariant() -> string
```

sergeyklay: We shouldn't use Zephir variables in PHP doc. Please consider using `$_SERVER` instead of `_SERVER`. 

```php
/**
 * Gets the preferred ISO locale variant.
 *
 * Gets the preferred locale accepted by the client from the
 * "Accept-Language" request HTTP header and returns the
 * base part of it i.e. `en` instead of `en-US`.
 *
 * Note: This method rely on `$_SERVER["HTTP_ACCEPT_LANGUAGE"]` header.
 *
 * @link https://www.iso.org/standard/50707.html
 */
```

None: I didn't check @ruudboon comment about `en-US`

tests/unit/Http/Request/GetClientAddressCest.php
niden: To be honest I have no idea about IPv6 and how they work. I just copied the tests to the correct file.

ruudboon (approved):

tests/unit/Http/Request/GetClientAddressCest.php
ruudboon: Ok. Will add a task for 4.1 to do a full IPv6 check. It's getting more and more important these days.

Issue #14723, opened 2020-01-17
[NFR]: Make mbstring required


alexbusu:
**Describe the bug**
A `php -r 'echo \Phalcon\Version::get();'` returns without errors. 
Trying to run the application, fails instead. Core is dumped by segmentation fault.

**To Reproduce**
I'm not sure what component causes the core dump, but I'll try to figure out.
Here is the coredump

> (gdb) bt
> #0  0x00005555557dd1a5 in zval_get_string_func ()
> #1  0x00005555557ec1e8 in zend_get_callable_name_ex ()
> #2  0x00005555557d735c in zend_call_function ()
> #3  0x00007fffe71d4a93 in ?? () from /usr/lib/php/20190902/phalcon.so
> #4  0x00007fffe71eba4f in ?? () from /usr/lib/php/20190902/phalcon.so
> #5  0x00007fffe72012fa in ?? () from /usr/lib/php/20190902/phalcon.so
> #6  0x00005555557d6f49 in zend_call_function ()
> #7  0x00007fffe71d4a93 in ?? () from /usr/lib/php/20190902/phalcon.so
> #8  0x00007fffe71e579a in ?? () from /usr/lib/php/20190902/phalcon.so
> #9  0x00007fffe731039e in ?? () from /usr/lib/php/20190902/phalcon.so
> #10 0x00005555557d6f49 in zend_call_function ()
> #11 0x00007fffe71d4a93 in ?? () from /usr/lib/php/20190902/phalcon.so
> #12 0x00007fffe71e579a in ?? () from /usr/lib/php/20190902/phalcon.so
> #13 0x00007fffe7388ef8 in ?? () from /usr/lib/php/20190902/phalcon.so
> #14 0x00005555557d6f49 in zend_call_function ()
> #15 0x00007fffe71d4a93 in ?? () from /usr/lib/php/20190902/phalcon.so
> #16 0x00007fffe71e579a in ?? () from /usr/lib/php/20190902/phalcon.so
> #17 0x00007fffe73debb0 in ?? () from /usr/lib/php/20190902/phalcon.so
> #18 0x00005555558721c7 in execute_ex ()
> #19 0x00005555557d709c in zend_call_function ()
> #20 0x00007fffe71d3a5f in ?? () from /usr/lib/php/20190902/phalcon.so
> #21 0x00007fffe75ab2bd in ?? () from /usr/lib/php/20190902/phalcon.so
> #22 0x00005555557d6f49 in zend_call_function ()
[...]

PR #14732, opened 2020-01-17
T14723 mbstring

niden:
Hello!

* Type: bug fix
* Link to issue: #14723 

Changed methods internally to use `mbstring` equivalent ones (`strtolower`, etc.) making `mbstring` a prerequisite

Thanks


CameronHall (approved):

sergeyklay:
Please add mbstring to the list of the required extension https://github.com/phalcon/cphalcon/blob/ef31ca8fdfdcbaeef24bd93a32107a43b6c1ba6b/config.json#L17-L26

phalcon/Config/ConfigFactory.zep
```diff
@@ -97,11 +97,11 @@ class ConfigFactory extends AbstractFactory
             );
         }
 
-        let adapter = strtolower(config["adapter"]),
+        let adapter = mb_strtolower(config["adapter"]),
```
sergeyklay: Is there any reason to use multi byte function for the adapter name? Do we plan provide support for PHP non in ASCII? AFAIK, mbstring functions are slower. So, it is curious to me, what we offer for such the price? Adapter names in national encodings?

sergeyklay (changes requested):

phalcon/Config/ConfigFactory.zep
sergeyklay: I don't mind paying where it is necessary. But I don't quite understand why I should always pay for what I don't use.

niden:
Closing this. We need to run more tests since this impacts performance especially in the HTTP layer

PR #14762, opened 2020-01-23
#13655 - All tests for Annotations\Annotation

jenovateurs:
Hello team!
I'm happy to tell you that I have tested Annotation class.
But I have some problems about constants. 
I can't get the value of PHANNOT_T_ARRAY, PHANNOT_T_STRING, ... 
These constants come from https://github.com/phalcon/cphalcon/blob/a48b31967a1ce27c10cbed61bee42e0c90addc86/ext/phalcon/annotations/scanner.h
I have made a private property to handle this but I'm not proud of it.
Any ideas ?
Thanks

* Type: code quality
* Link to issue: #13655

Thanks

### DIFF
--- a/tests/unit/Annotations/Annotation/GetArgumentCest.php
+++ b/tests/unit/Annotations/Annotation/GetArgumentCest.php
@@ -13,20 +13,45 @@ declare(strict_types=1);
 
 namespace Phalcon\Test\Unit\Annotations\Annotation;
 
+use Phalcon\Annotations\Annotation;
 use UnitTester;
 
 class GetArgumentCest
 {
+    private $PHANNOT_T_STRING = 303;
     /**
      * Tests Phalcon\Annotations\Annotation :: getArgument()
      *
-     * @author Phalcon Team <team@phalcon.io>
-     * @since  2018-11-13
+     * @author Jeremy PASTOURET <https://github.com/jenovateurs>
+     * @since  2020-01-22
      */
     public function annotationsAnnotationGetArgument(UnitTester $I)
     {
         $I->wantToTest('Annotations\Annotation - getArgument()');
 
-        $I->skipTest('Need implementation');
+        $sValue = 'test';
+        $sValue1 = 'test1';
+
+        $oAnnotation = new Annotation([
+            'name'       => 'NovAnnotation',
+            'arguments'  => [
+                [
+                    'expr' => [
+                        'type'  => $this->PHANNOT_T_STRING,
+                        'value' => $sValue
+                    ]
+                ],
+                [
+                    'expr' => [
+                        'type'  => $this->PHANNOT_T_STRING,
+                        'value' => $sValue1
+                    ]
+                ]
+            ]
+        ]);
+        
+        $I->assertEquals($oAnnotation->getArgument(0), $sValue);
+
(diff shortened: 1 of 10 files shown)


Jeckerson (dismissed):

tests/unit/Annotations/Annotation/GetArgumentCest.php
```diff
@@ -13,20 +13,45 @@
...
+     * @author Jeremy PASTOURET <https://github.com/jenovateurs>
+     * @since  2020-01-22
      */
     public function annotationsAnnotationGetArgument(UnitTester $I)
     {
         $I->wantToTest('Annotations\Annotation - getArgument()');
 
-        $I->skipTest('Need implementation');
+        $sValue = 'test';
+        $sValue1 = 'test1';
```
ruudboon: @niden do we have something in our code standards regarding variable naming? Didn't see this Hungarian notation before in our Phalcon projects. 
@jenovateurs not saying this is wrong, just wanna check if there's a strict policy on this.

tests/unit/Annotations/Annotation/GetExpressionCest.php
```diff
@@ -13,20 +13,51 @@
...
     public function annotationsAnnotationGetExpression(UnitTester $I)
     {
         $I->wantToTest('Annotations\Annotation - getExpression()');
 
-        $I->skipTest('Need implementation');
+        $sValue = 'test';
+        $sValue1 = 'test1';
+
+        $aOneExpr = [
+            'type'  => $this->PHANNOT_T_STRING,
```
ruudboon: Something in me says we should change `$this->PHANNOT_T_STRING` to `self::PHANNOT_T_STRING` but remember why. I think it was something performance wise but can't remember. @niden ideas?

tests/unit/Annotations/Annotation/GetArgumentCest.php
niden: We have the coding standard for Zephir and although we do not explicitly mention it, it kind of seeps into our tests i.e. camelCase, PSR-2 etc.

https://docs.phalcon.io/4.0/en/coding-standard

@jenovateurs lets change this to camelCase please just to be in sync with the rest of the tests. I will need to add something in the docs for this.

jenovateurs: @ruudboon and @niden, Okay, no problem, I will change it :-).

tests/unit/Annotations/Annotation/GetExpressionCest.php
jenovateurs: I have two problems about constants:
- Currently, I have copied/pasted the same constants in a few tests of Annotation
- When I try to use constants in the class, PHP_CodeSniffer triggers an error because for PHP > 7.1, we need to specify if it is a private/public/protected constant.

ruudboon (approved):

ruudboon:
Great work @jenovateurs I think we need to rewrite some constants later but let's establish some standards for that first

niden:
Thank you @jenovateurs

jenovateurs:
Thanks @ruudboon and @niden :-)

Issue #14779, opened 2020-01-26
[NFR]: Integration Suite Rework


niden:
- Check all tables created by the integration suite
- Remove any data from the dump files
    - adjust tests accordingly
- Enable auto reloading of the schema from Codeception
- Utilize `_envs` folder and configs to run tests on each RDBMS
- Ensure that each test can run independent of any other

Nice to have - integrate it with GitHub actions

ruudboon:
Preferable we merge this after https://github.com/phalcon/cphalcon/issues/14775 or merge it to that branch

niden:
Resolved in https://github.com/phalcon/cphalcon/pull/14803

PR #14780, opened 2020-01-26
4.0.x ci and tests

ruudboon:
Fixed codecoverage
dd dll's to Artifacts
Fixed a few windows tests


tests/unit/Image/Adapter/Gd/GetRealpathCest.php
```diff
@@ -32,6 +32,7 @@ public function imageAdapterGdGetRealpath(UnitTester $I)
         $I->wantToTest('Image\Adapter\Gd - getRealpath()');
 
         foreach ($this->getImages() as $image) {
+            $file    = str_replace("/", DIRECTORY_SEPARATOR, $image);
```
Jeckerson: `$file` variable is unused.

ruudboon: Thnx. Fixed

niden:
Once this passes I think we should merge it and move forward. OSX/Windows tests can fail (for now) and we can start working on fixing them or adjusting the run. 

Also this would help us with the #14779 to enable integration tests on a per RDBMs

Issue #14775, opened 2020-01-25
[NFR]: Move CI to github Actions


ruudboon:
Work is bundled in 4.0.x-ci-and-tests branch. We will merge this is when all all tasks are completed.

* [x] Build Pecl package
* [x] Build on Linux
* [x] Build on MacOS
* [x] Build on Windows (Only possible on Windows 2016, see https://github.com/actions/virtual-environments/issues/68)
* [x] Upgrade Codeception and use Phalcon 4 module
* [x] Database import handeling by Codeception
* [x] Run full test suite on Linux
* [x] Run unit tests and cli on MacOS
* [x] Run unit tests and cli on Windows
* [x] Fix failing tests MacOS
* [x] Fix failing tests Windows
* [x] Implement code coverage
* [x] Remove AppVeyor
* [x] Remove Travis
* [x] Update Readme how to run test suite
* [x] Remove or fix Nanobox
* [x] Add dll's to Artifacts 
* [x] Gcc backtrace on segfault
* [x] Fix large ddl issue
* [x] Run test build on zephir latest
* [x] Run syntax tests

Issue #14763, opened 2020-01-23
[BUG]: Character set issues under Catalina 


ruudboon:
Our Github runners run on the latest MacOS and show some issues with character conversion.
I was able to reproduce this on a local Mac as well. 

https://github.com/phalcon/cphalcon/runs/404587405#step:12:76

Expecting that we need to look at ext/kernel/filter.c - zephir_escape_multi()
Can you have a look at this @sergeyklay ?

Phalcon 4.0.2 - PHP 7.2 / PHP 7.3 / PHP 7.4

ruudboon:
Steps to reproduce this. Run on MacOS Catalina with Phalcon 4.0.2
```
<?php

$source = 'émotion';
$expected = '\e9 motion';

$escaper = new \Phalcon\Escaper();
$result = $escaper->escapeCss($source);

echo "Source: " . $source . PHP_EOL;
echo "Expected: " . $expected . PHP_EOL;
echo "Result: " . $result . PHP_EOL;
if ($result != $expected) {
    echo "fail" . PHP_EOL;
} else {
    echo "ok" . PHP_EOL;
}
```
I think escapeCss is calling: https://github.com/phalcon/cphalcon/blob/master/ext/kernel/filter.c#L294

And then:
https://github.com/phalcon/cphalcon/blob/master/ext/kernel/filter.c#L149

ruudboon:
This could maybe happening on windows as well. https://github.com/ruudboon/cphalcon/runs/409006716?check_suite_focus=true#step:28:380

ruudboon:
@sergeyklay Let's leave this for now. I think other issues are currently more important.

ruudboon:
This issue will be resolved with the next Zephir release. (0.12.17)

sergeyklay:
Zephir 0.12.17 released

ruudboon:
Fixed in 4.0.x

PR #14782, opened 2020-01-26
4.0.x ci and tests

ruudboon:
Fixed codecoverage
dd dll's to Artifacts
Fixed a few windows tests
Continue on failure for osx and windows
Setup tests on travis

See: #14775


niden (approved):

niden:
Merging this so that we can move forward with Actions. 

As mentioned before next tasks:
- Fix tests for osx/windows #14763 
- Create stack trace on failure
- Reduce the size of DLLs
- Restructure the integration suite to run on a per driver #14779 

Postgres tests fail on Travis but that is fine for now.

Great work buddy!

Issue #14733, opened 2020-01-18
[NFR]: Data Mapper: Connection


niden:
Build a new PDO Wrapper/Connection object

Parent: https://github.com/phalcon/cphalcon/issues/14126

niden:
Resolved in https://github.com/phalcon/cphalcon/pull/14824

PR #14824, opened 2020-02-08
T14733 Data Mapper PDO Connection

niden:
Hello!

* Type: new feature
* Link to issue: #14733 

Added new PDO wrapper for the Data Mapper implementation, with decorated instance, locator and profiler
  - `DM\Connection`
  - `DM\Connection\Decorated`
  - `DM\Profiler\Profiler`
  - `DM\Profiler\MemoryLogger`
  - `DM\ConnectionLocator`
This component will be used in the Data Mapper implementation but can be used as a stand alone component for PDO connections.

Thanks


phalcon/DM/Pdo/Connection.zep
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * Connects to the database.
+     */
+    public function connect() -> void
+    {
+        var dsn, options, password, query, queries, username;
+
+        if !this->pdo {
+            // connect
+            this->profiler->start(__FUNCTION__);
```
ruudboon: Can we make the profiler optional for perfomance?

phalcon/DM/Pdo/Connection.zep
```diff
@@ -0,0 +1,154 @@
...
+ * For the full copyright and license information, please view the LICENSE.txt
+ * file that was distributed with this source code.
+ *
+ * Implementation of this file has been influenced by AtlasPHP
+ *
+ * @link    https://github.com/atlasphp/Atlas.Pdo
+ * @license https://github.com/atlasphp/Atlas.Pdo/blob/1.x/LICENSE.md
+ */
+
+namespace Phalcon\DM\Pdo;
```
ruudboon: I think the the DM can be a bit confusing. This could also be Device management, Device mapper, Distributed Mama's ;) etc. Can we just name it DataMapper? It isn't super long, and I don't see the advantages of abbreviation here.

ruudboon (changes requested):
Great work buddy!

phalcon/DM/Pdo/Connection.zep
niden: The `if` statement is in the Profiler class (`active`). It was a tradeoff between putting if statements in every `fetch` method and not having one method call vs. having the method call and the `if` is in the profiler.

ruudboon (approved):

PR #14716, opened 2020-01-16
Fix ManyToMany select query across multiple schemas

alrieckert:
Hello!

Same bug fix PR as 
https://github.com/phalcon/cphalcon/pull/14715
Just for v4.0.x

Thanks

PR #14854, opened 2020-02-15
Test ManyToMany with multiple schemas

jenovateurs:
Hello team!
I created a test for the PR #14716 .
I added new table in the schema.sql.
I added new models with different schema.
I fix an error in .env.exemple of nanobox.

* Type: code quality
* Link to issue: #14716

Thanks


ruudboon:
@jenovateurs Thank you! I cancelled Appveyor to give priority to the release. I will restart it afterwards

jenovateurs:
@ruudboon No problem, thanks for the info ;-).

Jeckerson:
@jenovateurs Use fixtures migration classes instead of raw SQL.

Example here: `tests/_data/fixtures/Migrations/InvoicesMigration.php`

jenovateurs:
@Jeckerson No problem about fixtures migration. Should I delete my raw SQL in the schemas files ?

ruudboon:
@jenovateurs yes, you can delete the sql files you've added. We will move everything to the migrations soon. After that we can fully clean up.

jenovateurs:
What can I do about that: 

```
In Db.php line 278:
                                                                               
  Db: SQLSTATE[42000]: Syntax error or access violation: 1044 Access denied f  
  or user 'phalcon'@'%' to database 'hardware'                                 
  SQL query being executed:                                                    
  CREATE SCHEMA IF NOT EXISTS hardware  
```

Jeckerson:
@jenovateurs Did you rebased with 4.0.x?

jenovateurs:
@Jeckerson I've just rebased now but I still have the same issue.

ruudboon:
@jenovateurs I can have a look at it tomorrow. Must be something related to permissions on the github actions.

jenovateurs:
@ruudboon I think the problem is here : 
https://github.com/phalcon/cphalcon/blob/4.0.x/tests/_envs/pgsql.yml
In the database, I haven't got the permission to create a new schema.
I can create the new schema in the file environnement or add more privileges. 
We will see tomorrow, no problem ;-).

niden:
@jenovateurs If you have a better understanding and knowledge about Postgres make any necessary changes that are needed. 

Other than the area that you pointed regarding the schema creation, I think the rest of the database tests could run on postgres without major hitches (provided that the migrations are created)

ruudboon:
@jenovateurs Probably your latest commit will fail because you don't have permissions. 
To fix this we have a few options.
- Run tests as root
- Initialise permissions on starting of mysql/postgres containers.

Besides that I would like to use schema's that will be used for the full suite. This could be `hardware` but maybe we could go for something more general like `app` or `private` (we already have the `public` I think)

ruudboon:
Let’s run the suite as root for now. Looks like we need to extend the mysql and postgres containers otherwise. 
Can you change the example env and actions env to use the root user?

.github/workflows/codecoverage.yml
```diff
@@ -166,6 +166,7 @@ jobs:
           DATA_POSTGRES_PORT: ${{ job.services.postgres.ports['5432'] }}
           DATA_REDIS_PORT: ${{ job.services.redis.ports['6379'] }}
           DATA_MEMCACHED_PORT: ${{ job.services.memcached.ports['11211'] }}
+	  DATA_MYSQL_USER: root
```
ruudboon: Can indent this?

tests/_data/assets/db/schemas/mysql_schema.sql
```diff
@@ -12792,4 +12792,4 @@ CREATE TABLE `tiny_int_test` (
   `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
   `test` tinyint(1) unsigned DEFAULT NULL,
   PRIMARY KEY (`id`)
-) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```
ruudboon: Please end files with a new line

tests/_data/assets/db/schemas/postgresql_schema_nanobox.sql
```diff
@@ -7045,4 +7045,4 @@ GRANT ALL ON SCHEMA public TO PUBLIC;
 
 --
 -- PostgreSQL database dump complete
---
+--
```
ruudboon: same

ruudboon (changes requested):

.github/workflows/codecoverage.yml
```diff
@@ -166,7 +166,7 @@ jobs:
           DATA_POSTGRES_PORT: ${{ job.services.postgres.ports['5432'] }}
           DATA_REDIS_PORT: ${{ job.services.redis.ports['6379'] }}
           DATA_MEMCACHED_PORT: ${{ job.services.memcached.ports['11211'] }}
-	  DATA_MYSQL_USER: root
+	        DATA_MYSQL_USER: root
```
ruudboon: I think there is a tab here. Please use spaces only.

ruudboon (dismissed):

jenovateurs:
It's the sqlite test that trigger the error. Because I haven't the SQL script. I don't know if it's possible to create a schema with sqlite.

jenovateurs:
I skiped sqlite for this special test so if you have another idea... I'm all ears (or eyes) :-).

ruudboon (approved):

niden:
I wish I had seen this earlier. Nothing you did Jeremy but the goal is to move away from the Robots and related tables and use a more realistic schema such as invoices/customers/products etc. Also by doing this we ensure that we use only models we need and not throw one model/table here just for one test.

Great work nonetheless. Thank you for taking the initiative on this.

Also for the future if you do not want a test to run for a particular RDBMs you can use the `getDriver()` method from the database helper.

PR #14856, opened 2020-02-17
Clean db test (Robots/RobotsPart/RobotsPartToPart) -> Products

jenovateurs:
Hello team!

I remove some "robot" classes that I added yesterday. To replace the robots, I added Product and OrderProducts. 
I hope @niden that was you have in mind. If not we can discuss about that 😉.
I also delete the other public schema and just keep the private schema.

* Type: code quality
* Link to issue: #14854 

Thanks


Jeckerson (dismissed):

niden:
@jenovateurs That is great work. Thank you. I had something slightly different in mind but this gets us very close to that. 

Based on this: https://docs.phalcon.io/4.0/en/use-case

We already have `co_invoices`. Also there is a `inv_cst_id` field that links to `co_customers`.

Now we can have `co_orders` (prefix `ord`) and `co_products` (prefix `prd`). The cross table between them would be `co_orders_x_products` (`x` dictating that this is a cross table) with 

`oxp_ord_id` and `oxp_prd_id`

The methodology I had in mind is that every table has a 3 letter unique prefix - even cross tables so as to be able to identify the field where it comes from. The `co_` in tables is `core`. We can utilize that prefix for special cases that we might want to cover.

We don't need to have the full schema as one would expect for products and orders. For now the link ids and name or title would be fine, perhaps quantity in products also. As we move through the fields we can add more fields as need by.

Send me a message on discord if you need to discuss more or have additional input 

Great work so far!

jenovateurs:
Thanks  @niden for ypur review.
You've thought through the structure of the database and the methodology of creating tables and names.
Thus we will have a great schema to run efficient tests. 
I hope this new update matches you have in mind. I've put the quantity of product order in the cross table.

If you need new updates, tell me.

Thanks for your help and your ideas.

niden (approved):

ruudboon (approved):

ruudboon:
@jenovateurs great work. Thnx

PR #14876, opened 2020-02-25
#14862 - Swap condition if column allows null values in PostgreSQL

### LINKED ISSUE #14862: [BUG]: Model - Update - Postgres NULL value is not allowed on Phalcon 4.0.4
When I try to save a NULL value on postgresql, it does not save;

$model->getMessages() "return field1,field2,field3.... is required

I've get model from database and try to save it but got required errors on NULL values.

It's similar to issue https://github.com/phalcon/cphalcon/issues/14722

Jeckerson:
Hello!

* Type: bug fix
* Link to issue: #14862

Small description of change:

Thanks

### DIFF
--- a/tests/_data/fixtures/Migrations/AbstractMigration.php
+++ b/tests/_data/fixtures/Migrations/AbstractMigration.php
@@ -78,16 +78,9 @@ abstract class AbstractMigration
                     'delete from ' . $this->table . ';'
                 );
             } else {
-                $schema = getenv('DATA_POSTGRES_SCHEMA');
-                $exists = $this
-                    ->connection
-                    ->query("SELECT to_regclass('$schema.$this->table') AS exists;")
-                    ->fetchColumn();
-                if ($exists) {
-                    $this->connection->exec(
-                        'truncate table ' . $this->table . ' cascade;'
-                    );
-                }
+                $this->connection->exec(
+                    'truncate table ' . $this->table . ' cascade;'
+                );
             }
         }
     }
--- a/tests/_data/fixtures/Migrations/AbstractMigration.php
+++ b/tests/_data/fixtures/Migrations/AbstractMigration.php
@@ -69,14 +69,25 @@ abstract class AbstractMigration
                 ->connection
                 ->getAttribute(PDO::ATTR_DRIVER_NAME)
             ;
-            if ('sqlite' !== $driver) {
+            if ($driver === 'mysql') {
                 $this->connection->exec(
                     'truncate table ' . $this->table . ';'
                 );
-            } else {
+            } elseif ($driver === 'sqlite') {
                 $this->connection->exec(
                     'delete from ' . $this->table . ';'
                 );
+            } else {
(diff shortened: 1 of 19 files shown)


tests/_data/fixtures/Migrations/AbstractMigration.php
```diff
@@ -69,14 +69,25 @@ public function clear()
...
                 $this->connection->exec(
                     'delete from ' . $this->table . ';'
                 );
+            } else {
+                $schema = getenv('DATA_POSTGRES_SCHEMA');
+                $exists = $this
+                    ->connection
+                    ->query("SELECT to_regclass('$schema.$this->table') AS exists;")
+                    ->fetchColumn();
+                if ($exists) {
```
Jeckerson: Tomorrow will re-check if this is really needed for pgsql.

sergeyklay (dismissed):
LGTM

tests/_data/fixtures/Migrations/OrdersProductsMigration.php
```diff
@@ -71,18 +71,18 @@ protected function getSqlPgsql(): array
     {
         return [
             "
-drop table if exists private.co_orders_x_products;
+drop table if exists public.co_orders_x_products;
```
jenovateurs: @Jeckerson This table needs to remain in the "private" schema. Because I ran a test in the JoinCest file which checks the function ManyToMany with different schemas #14854. Before this test and bug fix, Phalcon didn't detect both schemas.
So can you put back the private schema only for this table. 
Thank you

jenovateurs (changes requested):

tests/_data/fixtures/Migrations/OrdersProductsMigration.php
ruudboon: @Jeckerson We can also merge this and let @jenovateurs change this in combination with the other tests.

ruudboon (dismissed):
Looks good.

tests/_data/fixtures/Migrations/OrdersProductsMigration.php
Jeckerson: @jenovateurs The problem is that in PostgreSQL we do not have private schema. I'll see if possible to adapt and work in both: MySQL and PGSQL.

jenovateurs: @Jeckerson You need to create it, I have already done this in my local branch. Like I said @ruudboon I can change this by doing the other issue.

Jeckerson: @jenovateurs alright, then make sense to you create an PR that will be merged, and then I'll merge it with current PR.

jenovateurs: @Jeckerson If we merge the two PR, it should be work. Do you want me to update the GitHub Action to run the PostgreSQL test in my PR?

jenovateurs (dismissed):

sergeyklay (approved):

PR #14888, opened 2020-03-02
Fix mysql column numeric default

pfz:
Hello!

* Type: bug fix

Small description of change:

- In Mysql, column manipulation is generating some lonely NULL at the end of a definition (compare to valid NOT NULL. This could mess up with some parsers.
- Remove quotes around numerical default Mysql column value

### DIFF
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -5,6 +5,7 @@
 
 ## Fixed
 - Fixed `Phalcon\Mvc\Model\Query\Builder::getPhql` to add single quote between string value on a simple condition. [#14874](https://github.com/phalcon/cphalcon/issues/14874) [@jenovateurs](https://github.com/jenovateurs)
+- Fixed schema manipulation in `Phalcon\Db\Dialect\Mysql`, remove unnecessary `NULL` in column definition, unquote numerical defaults. [#14888](https://github.com/phalcon/cphalcon/pull/14888) [@pfz](https://github.com/pfz)
 
 # [4.0.5](https://github.com/phalcon/cphalcon/releases/tag/v4.0.5) (2020-03-07)
 ## Added
--- a/phalcon/Db/Dialect/Mysql.zep
+++ b/phalcon/Db/Dialect/Mysql.zep
@@ -50,8 +50,6 @@ class Mysql extends Dialect
 
         if column->isNotNull() {
             let sql .= " NOT NULL";
-        } else {
-            let sql .= " NULL";
         }
 
         if column->isAutoIncrement() {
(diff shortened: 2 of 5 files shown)


pfz:
Sorry 2 PR to fix 2 different problems. Either we could merge this one with both solved, or do I remove the first commit from this one ?

niden:
@pfz Do you mind rebasing this branch with 4.0.x? I don't see why the test should fail from the current v4 code.

And yes we can close the other PR if this one fixes everything.

niden:
Thank you for the contribution!

pfz:
I've rebased with the new branch... The tests are erratics, sometimes they pass, sometimes not. They pass on branch, and here, on the PR, it fails on weird states, for db tests not concerns by my fix.
Now on windows.

niden:
> I've rebased with the new branch... The tests are erratics, sometimes they pass, sometimes not. They pass on branch, and here, on the PR, it fails on weird states, for db tests not concerns by my fix.
> Now on windows.

Thanks @pfz Yeah we know about the Windows tests. We will fix them in the next few days. Not sure what happened there, they used to work fine. The db issue you saw was related to a namespace, nothing to do with your branch.

niden (dismissed):

niden:
Ack forgot. @pfz Do you mind adding an entry in the CHANGELOG-4.0 please with this fix? Dont forget to add your name to get credit for this.

Other than that this is ready to be merged. Solid work! Thanks

niden (approved):

niden:
Solid work. Thank you @pfz

Jeckerson:
Unfortunately I skipped sawing this, and it broke again migrations...

This was introduced:
```
else {	
    let sql .= " NULL";	
}
```

Because it is not possible to create nullable columns in Phalcon. Will create new PR with more tests

pfz:
@Jeckerson Sorry I think I don't quite understand your statement. Do you mean that this was introduced (actually re-introduced) on purpose ?

I made this PR because this broke my own migration path. I use a schema diff from `SHOW CREATE TABLE` to what Phalcon is providing. With at least Mysql 5.7+/ mariadb 10.1+ (prior to that I cannot tell), fields are NULL by default if type allows it. You must add `NOT NULL` to enforce that a value is always set. Also with a default value, adding `NULL` is irrelevant, as you always can set to this column to NULL as long as `NOT NULL` is not specified. And with a NULL default, I ended up with a strange `DEFAULT NULL NULL`.
Also phalcon can handle nullable columns, depends on how you manage your schema, I inherit my schema migration from a script I made since Phalcon 1.
You may have a better insight on this matter, and I could be all wrong. I'm neither a DBA nor Phalcon expert.

PR #14971, opened 2020-05-01
Remove underscore prefix in methods

Jeckerson:
Hello!

* Link to issue:

Removed underscores in protected and extendable (not final) methods to be PSR-12 compliant.

Thanks


sergeyklay (approved):

niden:
We cannot do this now. It breaks backwards compatibility. If it was only private methods that would be fine but since it touches protected methods it will break applications that extend these classes and use those methods.

We can put a @todo on this for v5

sergeyklay:
Is there any place where we can start collect things for v5?

niden (changes requested):
Lets make a small change here so that this can be merged safely without breaking BC. 

Leave the underscore named methods where they are. Copy each of those methods to new ones without the underscore. From the one with the underscore, call the other one without. The underscore named methods become proxy ones. Add a @todo or @deprecated in the docblock of the underscore one so that we can remove those in v5.

This way we do what we need but also keep BC

ruudboon:
@Jeckerson If you have time can you rebase and have look at the requested changes?

ruudboon (changes requested):
Please add the BC methods

Jeckerson:
Will try to add it during today.

ruudboon (changes requested):
I would like to suggest to implement to logic in the methods that are going to be in v5.
Now we are getting extra unnecessary look ups `groupResult()` calls `_groupResult()` and it will be easier when cleaning up the deprecated methods.

phalcon/Mvc/Model.zep
```diff
@@ -3769,6 +3784,19 @@ abstract class Model extends AbstractInjectionAware implements EntityInterface,
      *
      * @param string|array table
      */
+     protected function doLowUpdate(<MetaDataInterface> metaData, <AdapterInterface> connection, var table) -> bool
+     {
+         return this->_doLowUpdate(metaData, connection, table);
```
sergeyklay, line 3789: Here and all other places: As a far as doLowUpdate used by default could you please invert this logic wrap doLowUpdate inside _doLowUpdate.

sergeyklay (changes requested):

ruudboon (approved):

sergeyklay (approved):

niden (approved):

zsilbi:
I think this is going to break BC this way, because the logic is not quite right.

For example, if someone had custom a `_postSaveRelatedRecords()` implemented in their `Model` before, it's not going to be ever called by the `Model` itself, after a save because it will only call the new function (`postSaveRelatedRecords()`) instead. 

```php
namespace My;

class Model extends \Phalcon\Mvc\Model {
   /**
    * My custom _postSaveRelatedRecords implementation
    */
   protected function _postSaveRelatedRecords(AdapterInterface $connection, $related): bool {
      // This won't be ever called by Model

      return parent::_postSaveRelatedRecords($connection, $related);
   }
}
```

Following this example, this solution only applies to those, who originally had `Model::_postSaveRelatedRecords()` calls in their `Model` for some reason.


```php
namespace My;

class Model extends \Phalcon\Mvc\Model {
   /**
     * My custom function
     * 
     * @throws Exception
     * @throws Model\Exception
     */
    public function doSomething(): void
    {
        // This will work, thanks to the proxy function
        $this->_postSaveRelatedRecords(
            $this->getWriteConnection(),
            $this->dirtyRelated
        );
    }
}
```

PR #15041, opened 2020-05-13
T15040 clone 2

### LINKED ISSUE #15040: [BUG]: Issue with Phalcon\Http\Message\AbstractCommon::cloneInstance
Currently `Phalcon\Http\Message\AbstractCommon::cloneInstance` clones the passed object and directly modifies object's property:
https://github.com/phalcon/cphalcon/blob/d59613f52e4cfc1c76f939f2e9ade01d04e4bbb3/phalcon/Http/Message/AbstractCommon.zep#L32-L40

This is not always correct because in some cases object's property is private. For example:
https://github.com/phalcon/cphalcon/blob/d59613f52e4cfc1c76f939f2e9ade01d04e4bbb3/phalcon/Http/Message/Uri.zep#L52-L59

There is a PoC on PHP just to show the problem:
```php
<?php

abstract class AbstractCommon
{
    final protected function cloneInstance($element, string $property)
    {
        $newInstance             = clone $this;
        $newInstance->{$property} = $element;

        return $newInstance;
    }
}

final class Uri extends AbstractCommon
{
    private $path = "";

    public function withPath(string $path)
    {
        return $this->cloneInstance($path, "path");
    }
}

$test = new Uri();
$test->withPath('action/reaction');
```

**Output:**
```
PHP Fatal error:  Uncaught Error: Cannot access private property Uri::$path in /mnt/work/phalcon/cphalcon/manual.php:8
Stack trace:
#0 /mnt/work/phalcon/cphalcon/manual.php(20): AbstractCommon->cloneInstance('action/reaction', 'path')
#1 /mnt/work/phalcon/cphalcon/manual.php(25): Uri->withPath('action/reaction')
#2 {main}
  thrown in /mnt/work/phalcon/cphalcon/manual.php on line 8

Fatal error: Uncaught Error: Cannot access private property Uri::$path in /mnt/work/phalcon/cphalcon/manual.php:8
Stack trace:
#0 /mnt/work/phalcon/cphalcon/manual.php(20): AbstractCommon->cloneInstance('action/reaction', 'path')
#1 /mnt/work/phalcon/cphalcon/manual.php(25): Uri->withPath('action/reaction')
#2 {main}
  thrown in /mnt/work/phalcon/cphalcon/manual.php on line 8
```

niden:
Hello!

* Type: bug fix 
* Fixes #15040 

Changed the visibility of methods and properties in `Phalcon\Http\Message\Response`, `Phalcon\Http\Message\ServerRequest` and `Phalcon\Http\Message\Uri` to work with `clone`.

Thanks

### DIFF
--- a/CHANGELOG-4.0.md
+++ b/CHANGELOG-4.0.md
@@ -13,6 +13,7 @@
 - Changed `Phalcon\Mvc\Model\MetaData::getDI` so that now it will throw a `Phalcon\Mvc\Model\Exception` if there is no `DiInterface` instance  [#15011](https://github.com/phalcon/cphalcon/issues/15011)
 - Changed `Phalcon\Http\Request::getJsonRawBody` to use `json_decode` instead of `Phalcon\Json::decode` [#14936](https://github.com/phalcon/cphalcon/issues/14936)
 - Changed `Phalcon\Factory\AbstractFactory` to expose `getService` which will throw an exception if it does not exist. Removed `checkService` and adjusted all references in `newInstance()` calls. [#15038](https://github.com/phalcon/cphalcon/issues/15038)
+- Changed the visibility of methods and properties in `Phalcon\Http\Message\Response`, `Phalcon\Http\Message\ServerRequest` and `Phalcon\Http\Message\Uri` to work with `clone`. [#15040](https://github.com/phalcon/cphalcon/issues/15040)
 
 ## Fixed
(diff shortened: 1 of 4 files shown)


ruudboon (approved):
Changes look good. Not sure if we can do this in 4.0 due to semver.
On the other hand I don't think a lot of people are affected by this so ok by me.

niden:
> Changes look good. Not sure if we can do this in 4.0 due to semver.
> On the other hand I don't think a lot of people are affected by this so ok by me.

It really does not affect people since we went from private to protected in visibility.

Issue #15074, opened 2020-06-05
[BUG]: afterFetch method in the model does not work when using model paginator


maulanasatyaadi:
I made a model that using the afterFetch method inside, it's work when directly fetch data from the model. Here is my code

    <?php
    declare(strict_types=1);

    use Phalcon\Mvc\Model;

    class Artist extends Model
    {
        public $id;
        public $slug;
        public $title;
        public $content;
        public $artwork;
        public $wallpaper;
        public $tags;
        public $website;
        public $genre;
        public $hits;
        public $verified;
        public $view;

        public function initialize()
        {
            $this->setSource('artist');
            $this->belongsTo('tags', Tags::class, 'id');
            $this->hasMany('id', Chord::class, 'artist');
            $this->hasMany('id', News::class, 'artist');
            $this->hasMany('id', UserRequest::class, 'artist');
            $this->hasMany('id', Vote::class, 'content', [
                'params' => [
                    'conditions' => 'division = "artists"'
                ]
            ]);
        }

        public function beforeUpdate()
        {
            $this->genre = implode('|', $this->genre);
        }

        public function afterFetch()
        {
            $this->genre = explode('|', $this->genre);
        }
    }

And this my pagination script

    public function genreAction()
    {
        $genre = $this->dispatcher->getParam('genre');
        $this->tag->prependTitle('Genre ' . $genre);

        $paginator = new ModelPaginator([
            'model' => Artist::class,
            'parameters' => [
                'genre LIKE :genre:',
                'bind' => [
                    'genre' => '%' . $genre . '%'
                ]
            ],
            'limit' => 29,
            'page' => $this->request->get('page') ?? 1
        ]);

        // Need to be cached
        $paginated = $paginator->paginate();

        $this->view->genre = $genre;
        $this->view->artists = $paginated;
    }

The result should be like

    object(Artist)[3]
        public 'id' => string '4761' (length=4)
        public 'title' => string 'Via Vallen' (length=10)
        public 'genre' =>
            array (size=2)
                0 => string 'Pop' (length=3)
                1 => string 'Rock' (length=4)
        ...
        // next object

But the result is

    array (size=3)
        'id' => string '4761' (length=4)
        'title' => string 'Via Vallen' (length=10)
        'genre' => string 'Pop|Rock' (length=8)
        ...
        // next array items

I can use another method like an array_map to generate a result items, but I think it will better if the result in pagination is a result object.

dwinmj:
[...]

PR #15075, opened 2020-06-05
Change pagination result items

maulanasatyaadi:
Let the result items output as an object, because sometimes we still need ORM feature to fetch another related data.

Type: bug fix
Link to issue: [https://github.com/phalcon/cphalcon/issues/15074](15074)

### DIFF
--- a/phalcon/Paginator/Adapter/Model.zep
+++ b/phalcon/Paginator/Adapter/Model.zep
@@ -122,7 +122,7 @@ class Model extends AbstractAdapter
                 [modelClass, "find"],
                 parameters
             );
-            let pageItems = items->toArray();
+            let pageItems = items;
         }
 
         //Fix next


ruudboon:
We're not accepting any pull's on the master. Our working branch is 4.1.x now. 
This change can cause refactoring for our users that are expecting an array. I do see the the need for having an option to return objects but maybe we can implement it in a backwards compatible way. 
Besides that we need to adjust/add tests, and changelog as well. 

@phalcon/core-team What do you guys think?

sergeyklay:
In general I don't see the real issue and can't get why we need this by design.

ruudboon:
Gonna close this one. We need some more info why we need this and need it in a BC compatible if we agree upon this. Let's start with a NFR on the repo and let's work from there.

PR #15076, opened 2020-06-07
Refactor db tests

ruudboon:
Hello!

* Type: code quality
* Link to issue: #13655 

Small description of change:

Refactored UniquenessCest


ruudboon:
@niden If you have time can you ping me to check how we can solve the sql db-lock? https://github.com/phalcon/cphalcon/runs/747581338?check_suite_focus=true#step:17:180

niden (approved):

niden:
Sqlite does not like group statements, where etc. Anything "fast" i.e. tests running seems to lock the sqlite database.

ruudboon:
Ok, let's leave it open then. I will try to fix it or remove sqlite from these tests.

ruudboon:
Looks like connection to db is still open (maybe cause by a transaction?) 
I'm wondering if `PRAGMA journal_mode = wal;` could solve our sqlite issues.

ruudboon:
@jenovateurs Do you have an idea what is causing this error?https://github.com/phalcon/cphalcon/pull/15076/checks?check_run_id=752726095#step:19:27

jenovateurs:
Hi @ruudboon, I think it's a type trouble. PostgreSQL is very strict about that.
In your case, `obj_name` it's a varchar and `obj_type` is a smallint, so two differents types but in the SQL request there is no differences.
So I think you discovered a new Phalcon bug. I don't understand why Phalcon check `obj_name` and `obj_type` with the same parameters, it makes no sense to me.
I think it's work with MySQL because there is no strong check about the type and it doesn't affect the result with the keyword OR.
The problem might be here : 
https://github.com/phalcon/cphalcon/blob/6f9ce5c4af4486f6b40413571182408732dc617c/phalcon/Validation/Validator/Uniqueness.zep#L373
These days, I have too much work to do. I haven't the time to debug line by line this code.

PR #15080, opened 2020-06-09
#15077 - Fix findFirst() return type

### LINKED ISSUE #15077: findFirst method return null, if record is not found
Hi,
i found that in phalcon4, when record is not found with findfirst method, this method retrun null, not false, like in phalcon3.
In offcial phalcon4 documentation its not mentioned. But if someoone looks in to phalcon .zep files, its ease to find, that findFirst can return mixed data type (because of Pahlcon/Mvc/Model/Query->execute() can https://github.com/phalcon/cphalcon/blob/fe17e3ae897b8d3f41d65ed328fcb734d1955b51/phalcon/Mvc/Model/Query.zep#L3643).

https://github.com/phalcon/ide-stubs/blob/d2adae3fba488ccba913a3f7fc310f23aa6de4a3/src/Phalcon/Mvc/Model.php#L791

Jeckerson:
Hello!

* Type: bug fix  | code quality
* Link to issue: #15077

### DIFF
--- a/phalcon/Mvc/Model.zep
+++ b/phalcon/Mvc/Model.zep
@@ -1534,7 +1534,7 @@ abstract class Model extends AbstractInjectionAware implements EntityInterface, 
      *     'hydration' => null
      * ]
      */
-    public static function findFirst(var parameters = null) -> <ModelInterface> | bool
+    public static function findFirst(var parameters = null) -> <ModelInterface> | null
     {
         var params, query;
 
--- a/phalcon/Mvc/ModelInterface.zep
+++ b/phalcon/Mvc/ModelInterface.zep
@@ -100,7 +100,7 @@ interface ModelInterface
      *
      * @param array parameters
      */
-    public static function findFirst(parameters = null) -> <ModelInterface> | bool;
+    public static function findFirst(parameters = null) -> <ModelInterface> | null;
 
     /**
      * Fires an event, implicitly calls behaviors and listeners in the events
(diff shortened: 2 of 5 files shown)


phalcon/Mvc/Model.zep
```diff
@@ -1534,7 +1534,7 @@ abstract class Model extends AbstractInjectionAware implements EntityInterface,
      *     'hydration' => null
      * ]
      */
-    public static function findFirst(var parameters = null) -> <ModelInterface> | bool
+    public static function findFirst(var parameters = null) -> <ModelInterface> | null
```
sergeyklay, line 1537: Please correct interface

sergeyklay (changes requested):

sergeyklay (approved):
From design point of view I'm OK with this change. But I'm not sure about the target branch. Are we going to break backward compatibility in minor branch?

/сс @phalcon/core-team

Jeckerson:
The problem, is that this needed to be changed back in `4.0.0`.

ruudboon:
Why can't we keep it as null and update the docs?

Jeckerson:
The docs are already updated.

ruudboon:
Hmm now I see. We do break BC so if we approach it purely like that we should do it in v5. On the other hand I think it was intended to have it like this in v4 so you could see this as a bug that we need we need to fix and address it in v4.1

ruudboon (approved):
Technically it looks good to me. Let's wait for @niden before merging this one into 4.1.x

niden (approved):

PR #15091, opened 2020-06-20
Let result item still in an object

maulanasatyaadi:
Hello!

* Type: bug fix 
* Link to issue: [#15074 ](https://github.com/phalcon/cphalcon/issues/15074)

Small description of change:
Item object result can give us ORM compatibility after make paging results.

Thanks

### DIFF
--- a/phalcon/Paginator/Adapter/Model.zep
+++ b/phalcon/Paginator/Adapter/Model.zep
@@ -122,7 +122,6 @@ class Model extends AbstractAdapter
                 [modelClass, "find"],
                 parameters
             );
-            let pageItems = items->toArray();
         }
 
         //Fix next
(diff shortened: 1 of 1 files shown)


ruudboon:
This will break backwards compatibility and we can't do that in 4.1. We could implement this but it should be in a backwards compatible way.  Rest of @phalcon/core-team what do you guys think? 

Besides that we need some tests for this.

sergeyklay:
I agree with @ruudboon

maulanasatyaadi:
But in Phalcon 3.4.x model adapter for paginator returns a resultset object, and that helps us to call-related model.

Jurigag:
Yea but for phalcon 4 it seems it was changed. We would need to add some Model setup option i guess to add option to change this behavior, but keep old as default to avoid breaking backwards compability and add this way without returning both array and object to avoid memory usage.

zsilbi:
Thank you @maulanasatyaadi but this breaks BC so we have to revisit this issue later.

Issue #13378, opened 2018-05-15
[NFR] add autowire to DI container


nejtr0n:
Modern DI containers supprot autowire feature.
(http://php-di.org/doc/autowiring.html)
But phalcons not. And it could not be replaced by them.
It would be grate to make container autowireable.

JABirchall:
+1, this is something laravel has done really well.
If a controller expects the id of a user, you can type him that $id as `User $id` and the framework automatically contructs a user model and fetches the user from the database for that ID and passes it as the paramiter.

nejtr0n:
@JABirchall This is not what i mean. Feature with user id is model binding. Sometimes it becomes antipattern imho.
I mean just resolve constructor argument from di, when it is not passed.
For example.
```
class Foo
{
public function __construct(Bar $bar)
....
}
```
And both of them are in di
```
$di->set("bar", Bar::class);
$di->set("foo", Foo::class);
```
So when we try to resolve foo
```
$di->get("foo");
```
it will automatically inject bar from di to foo.
We do not need build structures like
```
$di->set(
    'someComponent',
    [
        'className' => 'SomeApp\SomeComponent',
        'arguments' => [
            [
                'type' => 'service',
                'name' => 'response',
            ],
            [
                'type'  => 'parameter',
                'value' => true,
            ],
        ]
    ]
);
```

Ultimater:
+1, this would make it easier to implement the repository and decorator patterns.

While this NFR is asking for this to be implemented on the [DI service container](https://github.com/phalcon/cphalcon/blob/master/phalcon/di/service.zep).
I'd also like to see this to be implemented on controllers, like the Slim example in nejtr0n's link.

```
class UserController
{
    private $userRepository;

    public function __construct(UserRepository $userRepository)
    {
        $this->userRepository = $userRepository;
    }

    public function delete($request, $response)
    {
        $this->userRepository->remove($request->getAttribute('id'));

        $response->getBody()->write('User deleted');
        return $response;
    }
}

$app->delete('/user/{id}', ['UserController', 'delete']);
```
This makes it easier to implement the repository pattern if the repository can be injected into the constructor.

I see this way too often, and it ties down the dependency:
```
    public function __construct()
    {
        $this->userRepository = new UserRepository();
    }
```
Ideally we also want to support the decorator pattern so it can work with the dependency being sent in.
Without auto-wiring, we wind up writing this monstrosity: 
```
    public function __construct(UserRepository $userRepository = null)
    {
[...]

PR #15115, opened 2020-07-22
WIP: [4.1.x] Autowire component for phalcon

Jurigag:
Hello!

* Type: new feature
* Link to issue:https://github.com/phalcon/cphalcon/issues/13378

Small description of change: This is basic concept of autowire component for Phalcon's Di. How it works:

When creating `Phalcon\Di` there is optional argument implementing interface `Phalcon\Di\AutowireInterface`:

```php
$di = new \Phalcon\Di(new \Phalcon\Di\Autowire());
```

Which is used by `Phalcon\Di->get()` when called with string(class name) like this:

```php
$di->get(\Test\TestService::class); // or:
$di->set('testService', \Test\TestService::class);
$di->get('testService');
```

It analyzes constructor arguments and if they are classes tries to automatically get them from container and then creates new instance.

It also adds way to bind classes or interfaces, it will allow to either bind implementations to interfaces or add autowiring using currently written services to new projects. It also supports multiple implementations for one interfaces, to use then different implementation for different service developer needs to tell which exactly implementation he wants to use.

**Case with interface:**

Classes:

```php
namespace Test;

interface TestInterface
{
    public function test();
}

class TestService
{
    protected $testClass;

    public function __construct(TestInterface $test, $xyz)
    {
        $this->testClass = $test;
    }
}

class TestClass implements TestInterface
{
    protected $test;

    public function __construct()
    {
        $this->test = rand();
    }

    public function test()
    {
        // TODO: Implement test() method.
    }
}
```

Code:

```php
$di->bind(\Test\TestInterface::class, \Test\TestClass::class, true); // 3rd argument is to tell that this should be resolved as shared, so it will always return the same instance
var_dump($di->get(\Test\TestService::class));
var_dump($di->get(\Test\TestService::class));
```

Output:

```
object(Test\TestService)#10 (1) {
  ["testClass":protected]=>
  object(Test\TestClass)#8 (1) {
    ["test":protected]=>
    int(601129658)
  }
}
object(Test\TestService)#10 (1) {
  ["testClass":protected]=>
  object(Test\TestClass)#8 (1) {
    ["test":protected]=>
    int(601129658)
  }
}
```

When class we try to resolve has some other parameters like:

```
class TestService
{
    protected $testClass;

    public function __construct(TestInterface $test, $xyz)
    {
        $this->testClass = $test;
    }
}
```

This service then can be resolved like this:

```php
$di->get(\Test\TestService::class, ['xyz' => '123'])
```

Default parameters will be used if they exists, can be overridden the same way via second argument of get.

**Case with existing service**(you can also bind existing service to class name, and use then class name in __construct, it will automatically resolve to your existing service):

Code:

```php
$di->set('xyz', function () {
    return new \Test\TestClass();
}, true);
$di->bind(\Test\TestInterface::class, 'xyz', true);
var_dump($di->get(\Test\TestService::class));
var_dump($di->get(\Test\TestService::class));
```

Output:

```
 object(Test\TestService)#10 (1) {
  ["testClass":protected]=>
  object(Test\TestClass)#8 (1) {
    ["test":protected]=>
    int(601129658)
  }
}
object(Test\TestService)#10 (1) {
  ["testClass":protected]=>
  object(Test\TestClass)#8 (1) {
    ["test":protected]=>
    int(601129658)
  }
}
```

**Case with multiple implementations:**

Additional implementation:

```php
class TestClassOther implements TestInterface
{
    public function test()
    {
        // TODO: Implement test() method.
    }
}
```

Code:

```php
$di->bind(\Test\TestInterface::class, \Test\TestClass::class, true);
$di->bind(\Test\TestInterface::class, \Test\TestClassOther::class, true);
$di->set('testService', \Test\TestService::class)
    ->setAutowireTypes([\Test\TestInterface::class => \Test\TestClass::class]);
$di->set('testServiceOther', \Test\TestService::class)
    ->setAutowireTypes([\Test\TestInterface::class => \Test\TestClassOther::class]);
var_dump($di->get('testService'));
var_dump($di->get('testServiceOther'));
```

Output:

```
object(Test\TestService)#15 (1) {
  ["testClass":protected]=>
  object(Test\TestClass)#13 (1) {
    ["test":protected]=>
    int(1528930295)
  }
}
object(Test\TestService)#12 (1) {
  ["testClass":protected]=>
  object(Test\TestClassOther)#15 (0) {
  }
}
```

Things to consider(let me know what you think about these):

- automatically resolve multiple implementations of interface based on parameter name and/or resolved class name, not sure if it would be anyhow usefull and also it would add some bottleneck, need to figure out how it should work
- add interface `Phalcon\Di\AutowireAwareInterface` so only classes implementing it will be auto wired(though empty interface doing nothing feels kind of weird)
- add way to set non-class parameters for nested services - currently when resolving you can set non-class parameters only for resolving class, for nested classes it would need more work, either set them using bind, or add method to ServiceInterface to set constructor parameters for example
- component itself adds option to set Autowire to resolve not defined classes via set/bind as shared - not sure if this should be kept or not
- ~caching - needs tests how exactly it affects performance~ - not worth it, performance hit is really low, and any caching is not required, just localCache storing in property already resolved parameters for methods
- `Phalcon\Di::set` method is also changed, when definition is a class it will automatically bind it so it will be ready for autowire - it's just way to make it work if someone already uses calls like `$di->set('xyz', Class::class)`, it's to avoid unnecessary bind calls by developer
- add some kind of method like updateBinds to container - which will fetch all defined services, check if their definition is class/object not being closure and will add them to bind so autowire will resolve it with ease(we might also store it in some cache)
- since bind calls are pretty much pair of strings and bool add way to also implement cache for it to avoid method calls - what im thinking is it could be good for example add new method to `ServiceProviderInterface` in 5.0, for now create interface like maybe `ServiceProviderProvisionInterface` or `ServiceProviderCacheInterface` - which will be called only if container cache was empty
- ~resolve onConstruct methods called in a framework~ - **DONE**
- ~when using bind method with class as first argument(not interface) fetch also interfaces which class implements and add them as well(so either we will use interface or class in `__construct` it will get resolved)~ - implemented as option to enable, disabled by default
- is BindDefinition class necessary - we could just operate only on arrays and reduce memory usage
- add method resolver for autowire and implement it in dispatcher - so when user typehint class in action it will get resolved(separated feature request if this will get implemented and added?) - method resolver is done pretty much

Thanks


sergeyklay:
@Jurigag Could you please add msgpack to fix CI issue:
```
  Problem 1
    - The requested PHP extension ext-msgpack * is missing from your system. Install or enable PHP's msgpack extension.
```

phalcon/Di/Autowire.zep
```diff
@@ -0,0 +1,160 @@
...
+         */
+        if className->length() > this->maxKeyLength {
+            let className = className->sha1();
+        }
+
+        if isset this->localCache[className] {
+            return this->localCache[className];
+        }
+
+        if cache != null && cache->has("PHAW!" . className) {
```
Jeckerson, line 119: In another components there is prefix that can be specified.

Jurigag: > In another components there is prefix that can be specified.

Yea i guess this is not necessary. If user wants will set whatever prefix they want in cache adapter.

Jurigag: I removed it.

Jurigag:
I checked caching performance on PHP 7.2 and caching in redis/memcached/apcu seems useless, it's few times slower actually. Reflection Api is pretty fast, at least in PHP 7.2, will need to recheck it in PHP 7/7.1, if results will be similar i will remove it.

Storing parsed reflections in property gives around 2 times boost - so i will keep this.

Tested it on around 11 arguments constructor.

Jurigag:
I added way to resolve method calls. So `onConstruct` is autowired as well.

phalcon/Cache.zep
```diff
@@ -205,9 +205,9 @@ class Cache implements CacheInterface
     {
         let key = (string) key;
 
-        if preg_match("/[^A-Za-z0-9-_.]/", key) {
+        if preg_match("/[^A-Za-z0-9-_:\\\.]/", key) {
```
Jeckerson, line 208: Need to check if all adapters support that type of keys.

Jurigag: Yea, i was thinking in first place why exactly this was added. Any idea how i should check it? Is there any source why this was key check was implemented in first place?

Jeckerson: Memcached - https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L41

Jeckerson: image

File can be created in Linux.

Jeckerson: Redis - https://redis.io/topics/data-types-intro#redis-keys

Jurigag: Seems like most of them are fine. Will check a windows.

Jurigag: Well actually windows doesn't allow it.

Jurigag: I decided to revert allowed characters change, i will just str_replace them to `-` and and use `_` for method separator.

Jurigag: I will actuallty remove caching stuff from here, results are pretty bad, reflection api is fast enough to handle this on PHP 7.

Jurigag:
I moved binding logic from Di to Autowire component, though i will keep methods in Di - I feel it's more convenient to use them from di.

BeMySlaveDarlin:
Can you provide some config file based examples?
4e:
```php
[
    'container' => [
        DispatcherInterface::class => Dispatcher::class,
    ]
]
```

Jurigag:
> Can you provide some config file based examples?
> 4e:
> 
> ```
> [
>     'container' => [
>         DispatcherInterface::class => Dispatcher::class,
>     ]
> ]
> ```

Good idea, implemented it. Now you can just pass array with your binds to `Phalcon\Di\Autowire::__construct`. It accepts exactly they construction like you provided like array looking like this:

```
[
    DispatcherInterface::class => Dispatcher::class,
]
```

Or:

```php
[
    DispatcherInterface::class => new BindDefinition(DispatcherInterface::class, Dispatcher::class, true),
]
```

Jurigag:
Added unit tests for autowire component, later will add integration tests and unit tests for Di.

Jurigag:
@niden i heard you want to replace current DI with new one based on PSR-11, if it was supposed to be done for 5.0 i think this could still be added for 4.1.0 and something from here could be reused in 5.0, since PSR-11 is just container, it doesn't tell anyhow how autowiring should work.

Jeckerson:
Closing, all further work with Di will be done in `v5`.

PR #15116, opened 2020-07-25
Fix the issue that Controller::initialize() is skipped in some cases

zhaoyanliang8:
see https://github.com/phalcon/cphalcon/blob/634e7233a86780c9509614a8d835b188c8be76e5/phalcon/Dispatcher/AbstractDispatcher.zep#L338
This should also be merged into the 3.4 branch

### DIFF
--- a/phalcon/Dispatcher/AbstractDispatcher.zep
+++ b/phalcon/Dispatcher/AbstractDispatcher.zep
@@ -286,6 +286,19 @@ abstract class AbstractDispatcher extends AbstractInjectionAware implements Disp
             let params = this->params;
 
             /**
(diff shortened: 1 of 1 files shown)


Jeckerson:
Hello, thanks for contribution

There are some action points that need to be in consideration, before current PR can be merged:
* Destination need to be changed to `4.1.x` branch
* Version `3.4` is deprecated and it not supported
* There need to be fixed tests
* There need to be example case (with code) to simulate when method `initialize()` is really skipped. You can make it via tests

zhaoyanliang2:
I am not able to complete this pull request, I have submitted an issue to record this.
https://github.com/phalcon/cphalcon/issues/15117

PR #15158, opened 2020-09-23
Feature: callable relation params

zsilbi:
Hello!

* Type: new feature

- [x] I have created a PR for the [documentation](https://github.com/phalcon/docs) about this change: https://github.com/phalcon/docs/pull/2787

Small description of change:

Modified `Phalcon\Mvc\Model\Relation` to accept callable params for model relations.

Example use case:

```php
/**
 * @var DiInterface $container
 */
$container = $this->getDI();

$this->hasMany(
    'id', Documents::class, 'customerId', [
        'params' => function() use ($container) {
            /**
             * @var Languages $language
             */
            $language = $container->getShared('language');

            return [
                'languageId = :languageId:',
                'bind' => [
                    'languageId' => $language->id
                ]
            ];
        }
    ]
);
```

When `language` is changed in the dependency container, the relations will automatically use the new parameter.

Thanks,
zsilbi


Jeckerson:
But isn't it the same, when just pass and array directly inside 'params'?

```
/**
 * @var DiInterface $container
 */
$container = $this->getDI();

/**
  * @var Languages $language
  */
$language = $container->getShared('language');

$this->hasMany(
    'id', Documents::class, 'customerId', [
        'params' => [
                'languageId = :languageId:',
                'bind' => [
                    'languageId' => $language->id
                ]
            ]
    ]
);
```

zsilbi:
@Jeckerson 

The models are only initialized once and I need a way to keep these parameters up to date.

That function is evaluated every time before the relations are accessed, so if the language changes between requests in the container, the languageId in the params will be updated.

Jeckerson (approved):

Issue #15148, opened 2020-09-11
[BUG]: When updating including a relation, the relation destination is not updated


fagai:
**Describe the bug**

The following updates will not be updated.
```php
$user = User::find(1);
$user->userCard->token = $token;
$user->save(); // not update(4.0), update(3.4)
```

You can reinsert the object changed in this way,
```php
$user = User::find(1);
$userCard = $user->userCard;
$userCard->token = $token;
$user->userCard = $userCard;
$user->save(); // update!
```

It works if you run update directly at the relation destination.
```php
$user = User::find(1);
$user->userCard->token = $token;
$user->userCard->save(); // update!
```

I think this is because it has not been possible to detect that the data of the relation destination has been dirty.
This worked until v3.4.

**Details**
 - Phalcon version: 4.0.6
 - PHP Version: 7.4.9
 - Operating System: debian(docker)
 - Installation type: Compiling from source
 - Zephir version (if any):
 - Server: Apache
 - Other related info (Database, table schema):

**Additional context**

zsilbi:
I couldn't find anything in the docs (https://docs.phalcon.io/4.0/en/db-models-relationships) about this use case.
The fixes I made in #14035 affected this issue because I was not aware of this possible functionality.

I've always used it like your second example for updates: 
```php
$user = User::find(1);
$userCard = $user->userCard;
$userCard->token = $token;
$user->userCard = $userCard;
$user->save(); // update!
```

@phalcon/core-team, what's your opinion?

gponcon:
I have the same problem:

```php
# (Address BelongsTo User)
$address = Address::find(1);
$user = new User();
$user->address = $address;

# 3.4: _preSaveRelatedRecords() triggered, _preSave() triggered
# 4.0: _preSaveRelatedRecords() triggered, _preSave() triggered
$user->save();

$user->address->city = 'Paris';

# 3.4: _preSaveRelatedRecords() triggered, _preSave() triggered (saved)
# 4.0: _preSaveRelatedRecords() NOT triggered, _preSave() triggered (NOT saved)
$user->save();
```

However, reinserting the related record does not work for me either:

```php
# ...
$address = $user->address;
$address->city = 'Paris';
$user->address = $address;

# 3.4: _preSaveRelatedRecords() triggered, _preSave() triggered (saved)
# 4.0: _preSaveRelatedRecords() NOT triggered, _preSave() triggered (NOT saved)
$user->save();
```

[...]

PR #15203, opened 2020-11-04
Fix related records auto-save with belongsTo() relation

gponcon:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/15148

Small description of change: 

Models: with a relation `belongsTo()` - `hasMany()`, the related records need to be saved in `preSaveRelatedRecords()`. This fix update the dirtyState of related records to trigger related records save.

### DIFF
--- a/phalcon/Mvc/Model.zep
+++ b/phalcon/Mvc/Model.zep
@@ -1,4 +1,3 @@
-
 /**
  * This file is part of the Phalcon Framework.
  *
@@ -1006,6 +1005,7 @@ abstract class Model extends AbstractInjectionAware implements EntityInterface, 
                 continue;
             }
 
+            record->setDirtyState(self::DIRTY_STATE_TRANSIENT);
             let dirtyRelated[name] = record;
         }
 
@@ -4727,7 +4727,7 @@ abstract class Model extends AbstractInjectionAware implements EntityInterface, 
     protected function preSaveRelatedRecords(<AdapterInterface> connection, related) -> bool
     {
         var className, manager, type, relation, columns, referencedFields,
-            referencedModel, message, nesting, name, record;
+            message, nesting, name, record;
 
         let nesting = false;
 
@@ -4767,7 +4767,6 @@ abstract class Model extends AbstractInjectionAware implements EntityInterface, 
                     }
 
                     let columns = relation->getFields(),
-                        referencedModel = relation->getReferencedModel(),
                         referencedFields = relation->getReferencedFields();
 
                     if unlikely typeof columns == "array" {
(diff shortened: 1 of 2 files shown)


gponcon:
Note that this fix may not be perfect because when you access a linked record it seems to be saved anyway.

According to my tests:

```php
$invoice = Invoices::findFirst();
$invoice->customer->name = "name";
$invoice->save(); // customer is saved 👍 

$invoice = Invoices::findFirst();
$invoice->save(); // customer is not saved 👍 

$invoice = Invoices::findFirst();
echo $invoice->customer->name;
$invoice->save(); // customer is saved 👎 
```

Maybe there is something to do in `__set()` or somewhere else?

zsilbi:
> $invoice = Invoices::findFirst();
> $invoice->save(); // customer is not saved 👍 

What do you expect to be saved for `customer` in this case?

zsilbi:
> ```php
> $invoice = Invoices::findFirst();
> echo $invoice->customer->name;
> $invoice->save(); // customer is saved 👎 
> ```
> 
> Maybe there is something to do in `__set()` or somewhere else?

We don't have knowledge about the state of the properties in related objects.
If the relatations has been accessed in any ways, we will have to save that record, because it could have been modified.
To eliminate this, we would have to enforce using snapshots (which is not a possibility) and compare to the objects before each action.

`dirtyState` does not reflect on property changes, it only indicates if the record is saved to the database.

gponcon:
> > $invoice = Invoices::findFirst();
> > $invoice->save(); // customer is not saved 👍
> 
> What do you expect to be saved for `customer` in this case?

`customer` is not saved because it hasn't been accessed, which sounds pretty good.

gponcon:
> (...)
> We don't have knowledge about the state of the properties in (related) models.
> If the a relatation has been accessed in any way, we will have to save that record, because it could have been modified since the query.
> To eliminate this, we would have to enforce using snapshots (which is not a possibility) and compare them to their current state before each action.
> 
> `dirtyState` does not reflect on property changes, it only indicates if the record is saved to the database.

Thank you for this clarification. Indeed, I did not know how to deal with `dirtyState` for this resolution.

niden:
@gponcon Thank you so much for the bug fix. Could you please change your target branch to be `5.0.x`? We never send PRs to master unless it is a release

It is looking good

gponcon:
@niden I modified the target branch as requested. Let me know if it is correct, I am at your disposal.

niden:
@gponcon Can you please rebase to the latest 5.0.x branch and add an entry to the changelog? Then we can merge this

niden:
Closing in favor of https://github.com/phalcon/cphalcon/pull/15838. Original work preserved

PR #15207, opened 2020-11-10
Fix max allowed version

remicollet:
To not allow 8.0.0 (and all beta/RC), as "max" is allowed


niden (approved):

niden:
Thank you @remicollet

PR #15242, opened 2020-12-27
php8.0 update

defender-11:


Jeckerson:
Hello! You are trying to edit generated C code from Zephir code.

There is open CR for PHP8 in Zephir - https://github.com/phalcon/zephir/pull/2113

PR #15271, opened 2021-01-27
T15270 crypt exception with gcm

### LINKED ISSUE #15270: Incorrect implementation of encryption in GCM mode
Phalcon Version 4.0.6

According to the documentation [here](https://docs.phalcon.io/4.0/en/crypt), this code will work:
```
$crypt = new \Phalcon\Crypt();
$crypt->setKey('sdkfjhsdflksdhf;lkjdflj987kjsfd');
$crypt->setCipher('aes-256-gcm');
$crypt->useSigning(false);

$text = 'This is the text that you want to encrypt.';
$encrypted = $crypt->encrypt($text);

var_dump($encrypted);
die();
```

However, this throws an ErrorException:
```
openssl_encrypt(): A tag should be provided when using AEAD mode
```

This was also reported in #13869 which is marked as being solved in the 4.0.0 release but I'm on 4.0.6.

Also, looking at the code, the tag necessary to decrypt in GCM mode is never actually stored as part of the encryption string.
https://github.com/phalcon/cphalcon/blob/master/phalcon/Crypt.zep#L307

It is stored on the object, but not as part of the returned string. The code in the documentation can not possibly work if the Crypt object is garbage collected, as it would happen between requests.

niden:
Hello!

* Type: bug fix 
* Link to issue: #15270 

Added exception when no authData is present and gcm/ccm modes are selected

Thanks

### DIFF
--- a/CHANGELOG-4.1.md
+++ b/CHANGELOG-4.1.md
@@ -2,7 +2,7 @@
 
 ## Fixed
 - Fixed `Logger\Log::log()` `log` to recognize all log levels [#15214](https://github.com/phalcon/cphalcon/issues/15214)
-
+- Fixed `Crypt::encrypt()` to throw an exception when `ccm` or `gcm` modes are selected and no `authData` is present [#15270](https://github.com/phalcon/cphalcon/issues/15270)
 
 # [4.1.0](https://github.com/phalcon/cphalcon/releases/tag/v4.1.0) (2020-10-31)
 ## Added
--- a/phalcon/Crypt.zep
+++ b/phalcon/Crypt.zep
@@ -304,7 +304,11 @@ class Crypt implements CryptInterface
          * If the mode is "gcm" or "ccm" and auth data has been passed call it
          * with that data
          */
-        if ("-gcm" === mode || "-ccm" === mode) && !empty this->authData {
+        if ("-gcm" === mode || "-ccm" === mode) {
+            if empty this->authData {
+                throw new Exception("A tag must be provided when using AEAD mode");
+            }
+
             let authData      = this->authData,
                 authTag       = this->authTag,
                 authTagLength = this->authTagLength;
--- a/tests/unit/Crypt/EncryptCest.php
+++ b/tests/unit/Crypt/EncryptCest.php
@@ -113,11 +113,50 @@ class EncryptCest
      */
     public function cryptEncryptGcm(UnitTester $I)
     {
-        $I->wantToTest('Crypt - encrypt()');
+        $I->wantToTest('Crypt - encrypt() gcm/ccm');
 
         $ciphers = [
             'aes-128-gcm',
             'aes-128-ccm',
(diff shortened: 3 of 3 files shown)


zachleigh:
Wouldn't it be better to manage the auth data internally? The sixth param of openssl_encrypt is passed by reference and set by PHP so requiring the user to pass it in and manage that themselves doesn't really makes sense to me. If the tag was included in the returned encryption string and had a known length it could then be parsed when decrypting. Then the entire tag system could be done behind the scenes and the code would work as documented.

niden:
> If the tag was included in the returned encryption string and had a known length it could then be parsed when decrypting. Then the entire tag system could be done behind the scenes and the code would work as documented.

How would you do that? Return an array? I am not versed in cryptography and I have very little knowledge of the openssl method (other than reading the manual)

niden:
https://github.com/phalcon/phalcon/blob/v5.0.x/src/Crypt/Crypt.php#L226

This is the same method but only in PHP (Phalcon v5). If you have any suggestions on the implementation of what you are discussing let me know.

zachleigh:
For a gcm only implementation you could do something like this:
```php
public function encrypt($value)
{
    // these three set in settings or on the object
    $key = 'some random string of bits';
    $cipher = 'aes-256-gcm';
    $tagLength = 16;

    $ivLength = openssl_cipher_iv_length($cipher);
    $iv = openssl_random_pseudo_bytes($ivLength);

    $cipherText = openssl_encrypt(
        $value,
        $cipher,
        $key,
        0,
        $iv,
        $tag,
        '',
        $tagLength
    );

    return base64_encode($iv.$cipherText.$tag);
}

public function decrypt($value)
{
    // these three set in settings or on the object
    $key = 'some random string of bits';
    $cipher = 'aes-256-gcm';
    $tagLength = 16;

    $ivLength = openssl_cipher_iv_length($cipher);

    $encrypted = base64_decode($value);

    $iv = substr($encrypted, 0, $ivLength);
    $cipherText = substr($encrypted, $ivLength, -$tagLength);
    $tag = substr($encrypted, -$tagLength);

    return openssl_decrypt($cipherText, $cipher, $key, 0, $iv, $tag);
}
```

niden:
Thank you for these snippets.

For the `encrypt` in your example, it seems that `$tag` is not defined. Is it supplied? Also the second to last parameter is empty. Shouldn't that be `authData`?

zachleigh:
Yes, `$tag` is passed by reference and is set by PHP internally, like the `$matches` parameter in the [preg_match](https://www.php.net/manual/en/function.preg-match.php) function. 

The second to last parameter is for AAD (additional authentication data). This is optional and can be any string. If it is passed it needs to be the same for both encryption and decryption. In the examples I gave, the AAD would need to be passed by the user.

niden:
Closing in favor of https://github.com/phalcon/cphalcon/issues/15717

PR #15320, opened 2021-03-02
Update Session/Manager to validate session IDs before starting session

davidofferman:
Here's a patch to validate session IDs before opening the PHP session. Before this patch, sessions with invalid characters could be created and used.

This patch ensures that sessions IDs are alphanumeric. When a session ID doesn't match that criteria, it's unset and then the function `session_start()` will generate a new ID for the session.

**Concerns:**
This patch could possibly break user implemented Session Adapters if they implemented the method `create_sid()` from `SessionIdInterface`. This would happen if their code were to generate an ID that is not alphanumeric.

**Other Changes:**
I changed the method `getName()` in `Session\Manager` as it wouldn't get the session name unless you had previously set a session name in the manager.

Thanks

### DIFF
--- a/phalcon/Session/Manager.zep
+++ b/phalcon/Session/Manager.zep
@@ -151,7 +151,7 @@ class Manager extends AbstractInjectionAware implements ManagerInterface
      */
     public function getName() -> string
     {
-        if "" !== this->name {
+        if "" === this->name {
             let this->name = session_name();
         }
 
@@ -305,6 +305,10 @@ class Manager extends AbstractInjectionAware implements ManagerInterface
      */
     public function start() -> bool
     {
+        var name, value;
+
+        let name = this->getName();
+
         /**
          * Check if the session exists
          */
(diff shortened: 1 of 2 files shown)


phalcon/Session/Manager.zep
```diff
@@ -323,6 +327,16 @@ class Manager extends AbstractInjectionAware implements ManagerInterface
             throw new Exception("The session adapter is not valid");
         }
 
+        /**
+         * Verify that the session value is alphanumeric, otherwise we
+         * unset the cookie to allow it to be created by session_start().
+         */
+        if fetch value, _COOKIE[name] {
+            if !ctype_alnum(value) {
```
niden, line 335: @davidofferman Can we change this to `preg_match("/^[a-z0-9]+$/iD", value)` which does the same thing and does not rely on `ctype`? I don't want to introduce a new requirement for Phalcon to run.

niden (changes requested):

niden (approved):

niden:
Thank you @davidofferman

PR #15347, opened 2021-03-27
fix #15252 - compare 000012345 and 12345

### LINKED ISSUE #15252: form validation Confirmation strict comparison
Hi guys,
we have a problem using password confirmation validator, 

https://github.com/phalcon/cphalcon/blob/3241d96e4bb01dc7746b74c9ce586250b8887c46/phalcon/Validation/Validator/Confirmation.zep#L136

When we use the confirmation form validation it doesn't consider the type of the data. which is a problem when comparing numbers .
This returns true for `000012345` = `12345`

it should be either `return strcmp($a, $b) === 0;`
or allow us to override the compare method, which we cannot at the moment because of `final protected`

thanks

Izopi4a:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/15252

Small description of change:

changed how we compare strings in Confirmation validator

Thanks


niden (approved):

niden:
Thank you @Izopi4a

PR #15353, opened 2021-03-30
fixing Flash\Session::has() returning always true

### LINKED ISSUE #15204: Phalcon\Flash\Session::has return true when messages array is empty
Hi,

Using `Phalcon\Flash\Session::has()` return true whereas the messages array is empty.

https://github.com/phalcon/cphalcon/blob/a8035814aa9a8a2438f92ebf5128964e7c48fd3f/phalcon/Flash/Session.zep#L43

```
/**
     * Checks whether there are messages
     */
    public function has(type = null) -> bool
    {
        var messages;

        let messages = this->getSessionMessages(false);

        if typeof type != "string" {
            return true;
        }

        return isset messages[type];
    }
```

Sorry I don't understand why `type != 'string` return true,
but maybe it's possible to add something like :
```
if empty messages {
    return false;
}
```

Regards,

Izopi4a:
Hello!

* Type: bug fix 
* Link to issue: https://github.com/phalcon/cphalcon/issues/15204

Small description of change:

I am basically reverting the changes on Flash\Session::has() from this PR https://github.com/phalcon/cphalcon/pull/13994/files

I don't agree with the way funtion is changed, feels like just a bug

Thanks

### DIFF
--- a/phalcon/Flash/Session.zep
+++ b/phalcon/Flash/Session.zep
@@ -31,26 +31,35 @@ class Session extends AbstractFlash
 
     /**
      * Returns the messages in the session flasher
+     *
+     * @param string|null $type
+     * @param bool $remove
+     *
+     * @return array
      */
-    public function getMessages(type = null, bool remove = true) -> array
+    public function getMessages(var type = null, bool remove = true) -> array
     {
         return this->getSessionMessages(remove, type);
     }
 
     /**
      * Checks whether there are messages
+     *
+     * @param string|null $type
+     *
+     * @return bool
      */
-    public function has(type = null) -> bool
+    public function has(var type = null) -> bool
     {
         var messages;
 
         let messages = this->getSessionMessages(false);
 
-        if typeof type != "string" {
-            return true;
+        if typeof type == "string" {
+            return isset messages[type];
         }
 
-        return isset messages[type];
+        return count(messages) > 0;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


phalcon/Flash/Session.zep
```diff
@@ -39,18 +39,22 @@ class Session extends AbstractFlash
 
     /**
      * Checks whether there are messages
+     * 
+     * @param string|null $type
+     *
+     * @return bool
      */
-    public function has(type = null) -> bool
+    public function has(var type = null) -> bool
```
Jeckerson, line 47:
```suggestion
    public function has(string type = null) -> bool
```

niden (dismissed):

Izopi4a:
i will check to see what is wrong, probably i broke something , changed the MR name to wip

Izopi4a:
ok, so if the declarion in zephir is as followed:

```
public function test(string test = null) {
    var_dump(test);
}

test(); // will output string ""
```

Jeckerson:
@Izopi4a Yep, just simulated in Zephir. For now we can specify `var` and check type if is string.

https://github.com/zephir-lang/zephir/issues/2180

Jeckerson (approved):

niden (approved):

niden:
Fixes https://github.com/phalcon/cphalcon/pull/15353#pullrequestreview-625438556

@Izopi4a Thank you!

PR #15395, opened 2021-04-12
T14764 gettext setlocale

### LINKED ISSUE #14764: [BUG]: Phalcon\Translate\Adapter\Gettext:setLocale doesn't allow multiple locales
In our [docs](https://github.com/phalcon/cphalcon/blob/master/phalcon/Translate/Adapter/Gettext.zep#L207) we mention that `setLocale` in `Phalcon\Translate\Adapter\Gettext` allows multiple arguments just like the native [setLocale](https://www.php.net/manual/en/function.setlocale.php) 

The logic to handle this is [there](https://github.com/phalcon/cphalcon/blob/master/phalcon/Translate/Adapter/Gettext.zep#L212) but our method doesn't allow it and will return a `wrong number of parameters`.

Phalcon: 4.0.2
Php: 7.4

niden:
Hello!

* Type: bug fix 
* Link to issue: Fixes #14764 

Changed the signature of setLocale in `Translate\Adapter\Gettext` to allow multiple locale declarations if necessary. Aligns with `setlocale`

Thanks


docker/8.0/Dockerfile
```diff
@@ -1,50 +1,101 @@
...
+        zip
+
+# PECL Packages
+RUN pecl install psr && \
     pecl install -o -f redis && \
     pecl install igbinary && \
     pecl install msgpack && \
     pecl install apcu && \
     pecl install yaml && \
+    pecl install imagick && \
```
Jeckerson, line 49: There is no PHP8 compatible version for it
https://pecl.php.net/package/imagick

We compile it manually below.

PR #15412, opened 2021-04-19
T15411 logger enhancements

### LINKED ISSUE #15411: [BUG]: Logger not using Item time
The Logger component does not use the `Item` time (entry to be logged). This is technical debt from before and the interface needs to change when creating an item.

- Benchmarks need to be created to check whether `Item` (object) is more expensive than an array instead.
- The time needs to be corrected and the timezone needs to be introduced in the adapter
- The transactional mode of the logger has a loop with the queued items and writes each item in the loop. We need to change that to a processing cycle (for/next loop) and then one write - especially for the `Stream` adapter

niden:
Hello!

* Type: bug fix - enhancement
* Link to issue: Fixes #15411 

Changes to the `Phalcon\Logger`:
    - Renamed `Phalcon\Logger\Item::getName` to `Phalcon\Logger\Item::getLevelName`   
    - Renamed `Phalcon\Logger\Item::getType` to `Phalcon\Logger\Item::getLevel`
    - Changed the logger timestamp to be `DateTimeImmutable` on a per item basis - more accurate timing
    - Changed `Phalcon\Logger` to accept a `DateTimeZone` in the constructor 
    - Changed placeholder `%type%` to `%level%` to align with the variables [#15411](https://github.com/phalcon/cphalcon/issues/15411)

Thanks


Jeckerson:
Phalcon isn't ready for `0.13.2` Zephir atm.

niden:
This is on hold until a new version of Zephir is released to correctly handle `DateTimeImmutable` and `DateTimeZone`.

cc @Jeckerson

Jeckerson:
Error in pgsql tests occur due passing empty string schema, instead of null.

phalcon/Logger.zep
```diff
@@ -80,21 +89,35 @@ class Logger implements LoggerInterface
     protected name = "";
 
     /**
-     * The excluded adapters for this log process
-     *
-     * @var AdapterInterface[]
+     * @var DateTimeZone
```
Jeckerson, line 92:
```suggestion
     * @var DateTimeZone|null
```

phalcon/Logger.zep
```diff
@@ -80,21 +89,35 @@ class Logger implements LoggerInterface
...
+    protected timezone = null;
 
     /**
      * Constructor.
      *
-     * @param string name     The name of the logger
-     * @param array  adapters The collection of adapters to be used for logging (default [])
+     * @param string       name     The name of the logger
+     * @param array        adapters The collection of adapters to be used for logging (default [])
+     * @param DateTimeZone timezone The timezone
```
Jeckerson, line 101:
```suggestion
     * @param DateTimeZone|null timezone The timezone
```

phalcon/Logger/Formatter/AbstractFormatter.zep
```diff
@@ -31,12 +31,12 @@ abstract class AbstractFormatter implements FormatterInterface
      * @param string $message
      * @param array $context
      */
-    public function interpolate(string message, var context = null)
+    public function interpolate(string message, var context = null) -> string
```
Jeckerson, line 34: Could we transform `var context = null` to `array context = []`

tests/unit/Logger/Item/GetMessageCest.php
```diff
@@ -25,7 +26,8 @@ class GetMessageCest
     public function loggerItemGetMessage(UnitTester $I)
     {
         $I->wantToTest('Logger\Item - getMessage()');
-        $time = time();
+        $time = new DateTimeImmutable("now");
+        ;
```
Jeckerson, line 30:
```suggestion
```

Jeckerson (approved):

PR #15423, opened 2021-04-21
T15422 version top level

### LINKED ISSUE #15422: [NFR]: Remove top level classes: Version
- Move `Version` from the top level namespace
- Add it to a new `Support` namespace
- Make it non static

### LINKED ISSUE #15424: [BUG]: Phalcon\Cache - keys need to be strings
For cache operations, keys need to be casted to strings prior to sending them to adapters, to avoid fatal errors.

niden:
Hello!

* Type: enhancement
* Link to issue: Fixes #15422 #15424 

`Phalcon\Version` is not moved to `Phalcon\Support\Version`
    - `_getSpecialVersion` and `_getVersion` have been removed (marked deprecated in v4)
    - The class is no longer static; it has to be instantiated first
    - References to `Phalcon\Debug` and the Volt compiler have been adjusted 
Corrected `Phalcon\Cache` to cast keys as strings before sending them to adapters

Thanks


phalcon/Cache.zep
```diff
@@ -60,6 +60,8 @@ class Cache implements CacheInterface
      */
     public function delete(var key) -> bool
     {
+        let key = (string) key;
```
Jeckerson, line 63: It is possible to change method signature -> `public function delete(string key)`

phalcon/Cache.zep
```diff
@@ -103,6 +105,8 @@ class Cache implements CacheInterface
      */
     public function get(var key, var defaultValue = null) -> var
     {
+        let key = (string) key;
```
Jeckerson, line 108: It is possible to change method signature -> `public function get(string key)`

phalcon/Cache.zep
```diff
@@ -144,6 +148,8 @@ class Cache implements CacheInterface
      */
     public function has(var key) -> bool
     {
+        let key = (string) key;
```
Jeckerson, line 151: `has(string key)`

phalcon/Cache.zep
```diff
@@ -164,6 +170,8 @@ class Cache implements CacheInterface
      */
     public function set(var key, var value, var ttl = null) -> bool
     {
+        let key = (string) key;
```
Jeckerson, line 173: `set(string key ...`

phalcon/Cache.zep
niden: No because PSR uses `mixed` for that.

phalcon/Cache.zep
niden: Same as above

phalcon/Cache.zep
niden: Same

phalcon/Cache.zep
niden: Same

Jeckerson (approved):

PR #15428, opened 2021-04-22
package both php 7 and 8 sources for pecl

remicollet:
Trying to make things simpler for pecl, and allowing to use same package for both PHP 7 and 8

notice: config.w32 blindly written (inspired from zip ext by @cmb69)

Side effect: "phpize; ./configure; make" work in top directory

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -76,11 +76,6 @@ jobs:
       - name: Create Pecl Package (PHP 8)
         id: pecl_create
         run: |
-          cp build/php8/safe/config.w32 config.w32
-          cp build/php8/safe/phalcon.zep.c phalcon.zep.c
-          cp build/php8/safe/config.m4 config.m4
-          cp build/php8/safe/php_phalcon.h php_phalcon.h
-          cp build/php8/safe/phalcon.zep.h phalcon.zep.h
           pecl package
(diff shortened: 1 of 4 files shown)


remicollet:
Notice: the phalcon 5.x master tree seems really not ready

in php_phalcon.h 

- PHP_PHALCON_VERSION still 4.1.0

in paclahge.xml

- version is 5.0.0 (instead of 5.0.0alpha1)
- stability still "stable" (instead of alpha)
- max version 8.0.3 doesn't seems to make sense
- pear min version can be set to 1.10 (required for PHP > 7)

remicollet:
PHP 7 failure (missing` #include <ext/session/php_session.h>`) doesn't seems related to this PR

Jeckerson:
> Notice: the phalcon 5.x master tree seems really not ready

Yes, there were no plans to release 1st alpha to the pecl, nor have Win builds.

> allowing to use same package for both PHP 7 and 8

Probably we can resolve it via Zephir with version checks in `phalcon.c` and unify all into single folder. @remicollet what it your opinion about that?

remicollet:
> Probably we can resolve it via Zephir with version checks in `phalcon.c` and unify all into single folder. @remicollet what it your opinion about that?

I don't have any opinion ;)

Various extensions already use the multiple tree way (zip, event, ...)

Jeckerson:
Already implemented, closing.

PR #15453, opened 2021-05-03
Fix form::bind with whitelist not ignoring bad data values

### LINKED ISSUE #15070: [BUG]: \Phalcon\Forms\Form->bind the whitelist doesn't work for getValue
```php
$form = new \Phalcon\Forms\Form();
$form->add(new \Phalcon\Forms\Element\Text('nameFirst'));
$form->add(new \Phalcon\Forms\Element\Text('nameLast'));        
        
$data = [
    'test1' => 'test1',
    'nameFirst' => 'nameFirst',            
    'nameLast' => 'nameLast',
];

$entity = new \stdClass();
$white_list = ['nameFirst', 'nameLast'];
$form->bind($data, $entity, $white_list);        

// return TRUE - valid
var_dump($form->getValue('nameFirst') == 'nameFirst'); 
// return TRUE - NOT valid (not in whitelist/element list)
var_dump($form->getValue('test1') == 'test1');
```

**Details**
 - Phalcon version: 4.0.6
 - PHP Version: 7.4.6

Izopi4a:
Hello!

* Type: bug fix 
* Link to issue: https://github.com/phalcon/cphalcon/issues/15070

Small description of change:

Thanks

### DIFF
--- a/CHANGELOG-5.0.md
+++ b/CHANGELOG-5.0.md
@@ -35,6 +35,7 @@
 - Corrected `Phalcon\Db\Adapter\Pdo\AbstractPdo::connect` to take into account the `persistent` option for relevant connections [#15241](https://github.com/phalcon/cphalcon/issues/15241)
 - Corrected `Phalcon\Image\Adapter\Gd::processText` to correctly call `imagettfbbox` when a font is specified [#15188](https://github.com/phalcon/cphalcon/issues/15188)
 - Corrected `Phalcon\Cache` to cast keys as strings before sending them to adapters [#15249](https://github.com/phalcon/cphalcon/issues/15249)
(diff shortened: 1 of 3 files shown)


niden (approved):

niden:
Thank you @Izopi4a

Izopi4a:
there is a potential people using the Form::bind mething in a "abusive way" because you can bind values without having such elements. 
If people were doing that it would be wrong, but in case of reports, we could alter the changes in a way that we only filter the bindData when whitelist is provided instead as always

niden:
Thank you @Izopi4a 

We will add a note in the documentation for sure about this  (your last comment)

PR #15563, opened 2021-06-30
Update Form.zep

stijn1989:
The uniqueness validator fails when using an entity in a form. In this form I assign a validator object. One of the rules in the validator is that the "name" field must be unique.

Works great but when I use this form to edit, it has a persisted entity. So the uniqueness validator checks that their is no record of that name and doesn't belong to the primary key field. 

Only problem is that the validator doesn't receive the entity object that is assigned to the form. It receives the entity that is passed to the method "isValid($data, $entity)". 

This one line added to the isValid function, passes the form entity as the entity to validation.

Hello!

* Type: bug fix

Small description of change:

Thanks

### DIFF
--- a/phalcon/Forms/Form.zep
+++ b/phalcon/Forms/Form.zep
@@ -566,6 +566,7 @@ class Form extends Injectable implements Countable, Iterator, AttributesInterfac
             this->bind(data, entity);
         } else {
             if typeof this->entity == "object" {
+                let entity = this->entity;
                 this->bind(data, this->entity);
             }
         }


niden (changes requested):
Three things:

- First thank you for the bug fix. Really appreciated
- You need to create an issue and reference this PR to that issue (the issue we explain what is wrong)
- You need to update the `CHANGELOG`

Other than that this looks good !

stijn1989:
> Three things:
> 
> * First thank you for the bug fix. Really appreciated
> * You need to create an issue and reference this PR to that issue (the issue we explain what is wrong)
> * You need to update the `CHANGELOG`
> 
> Other than that this looks good !

Change CHANGELOG or CHANGELOG-5.0?

niden:
> > Three things:
> > 
> > * First thank you for the bug fix. Really appreciated
> > * You need to create an issue and reference this PR to that issue (the issue we explain what is wrong)
> > * You need to update the `CHANGELOG`
> > 
> > Other than that this looks good !
> 
> Change CHANGELOG or CHANGELOG-5.0?

No actually put a line in the CHANGELOG-5.0 (sorry my typo above) that explains what you did and links to the original issue.

niden:
Also you need to rebase and target the 5.0.x branch. `master` is only for releases.

PR #15752, opened 2021-11-02
T14203 fix form validator get value

### LINKED ISSUE #14203: Form Validation and Entity are not playing well together
If you pass both data and entity parameters to `$form->isValid()`, even though you expect `$validation->getValue($field)` to fallback to data if entity does not have that value (like From does), it doesn't happen. **Validation will not try to read value from any other data source if entity is present**.

For an example if you implemented CSRF check through validation, that wont work. That is, unless you have some sick, twisted desire to have a CSRF getter in your entities.

If you split `dateOfBirth` into three fields, glue them back and set to entity after validation, and you attach i.e. presence validators to those fields, the form will still fail validation if you didn't implement getters for those three fields in your entity. That's shame because it forces you to polute the entity with getters you don't need as we can utilize form's getters for displaying calculated values i.e. `$form->getMonth()`.

https://github.com/phalcon/cphalcon/blob/65183349743e899977adc643f103b887612ffe98/phalcon/validation.zep#L463-L524

Another thing that is giving me troubles is the fact that `$form->isValid($data, $entity)` automatically calls `$this->bind($data, $entity)` but doesn't accept whitelist parameter, forcing you to call `$form->isValid($data)` and then `$form->bind($data, $entity)`, which in turn makes the entity unavailable to validators because `$form->isValid()` passes entity to validation class only if it was provided to the function directly, it does not pass entity from `$this->_entity` as a fallback.

So my proposal is following:
1. Rewrite `Phalcon\Validation::getValue()` method so that it tries to retrieve value from the array as well, before it gives up when it can't find one in the entity. **But it should not write back filtered values if they came from an array while entity is present, as validation is unaware of whitelisting.** 

2. Rewrite `Phalcon\Forms\Form::isValid()` to accept whitelist as a third parameter which will be passed onto `Phalcon\Forms\Form::bind()`

BeMySlaveDarlin:
Hello!

* Type: new feature
* Link to issue: #14203
* **POSSIBLY BREAKS BACKWARD COMPATIBILITY**

Small description of change:

- `Phalcon\Validation::getValue()` seeks through `this->data` if not found in `this->entity`;
- `Phalcon\Forms\Form::isValid()` signature was changed: added 3rd argument `whitelist`;
- Added setter and getter for `whitelist` in `Phalcon\Forms\Form`.

Thanks

### DIFF
--- a/phalcon/Forms/Form.zep
+++ b/phalcon/Forms/Form.zep
@@ -205,8 +205,7 @@ class Form extends Injectable implements Countable, Iterator, AttributesInterfac
             /**
              * Check if the item is in the whitelist
              */
-
-            if !in_array(key, whitelist) {
+            if !empty whitelist && !in_array(key, whitelist) {
                 continue;
             }
 
--- a/phalcon/Forms/Form.zep
+++ b/phalcon/Forms/Form.zep
@@ -76,12 +76,12 @@ class Form extends Injectable implements Countable, Iterator, AttributesInterfac
     /**
      * @var ValidationInterface|null
      */
-    protected validation = null { set, get };
+    protected validation = null { get };
 
     /**
      * @var array
      */
-    protected whitelist = [];
+    protected whitelist = [] { get };
 
     /**
      * Phalcon\Forms\Form constructor
@@ -383,16 +383,6 @@ class Form extends Injectable implements Countable, Iterator, AttributesInterfac
     }
 
     /**
-     * Returns whitelist array
-     *
-     * @return array
-     */
-    public function getWhitelist() -> array
-    {
-        return this->whitelist;
-    }
-
-    /**
      * Returns a label for an element
      */
     public function getLabel(string! name) -> string
@@ -616,8 +606,9 @@ class Form extends Injectable implements Countable, Iterator, AttributesInterfac
      *
      * @param array data
      * @param object entity
+     * @param array whitelist
      */
-    public function isValid(var data = null, var entity = null, var whitelist = null) -> bool
+    public function isValid(var data = null, var entity = null, array whitelist = []) -> bool
     {
         var messages, element, validators, name, filters, validator, validation,
             elementMessage;
@@ -627,7 +618,7 @@ class Form extends Injectable implements Countable, Iterator, AttributesInterfac
             return true;
         }
 
-        if whitelist === null && !empty this->whitelist {
+        if empty whitelist {
             let whitelist = this->whitelist;
         }
 
@@ -837,9 +828,21 @@ class Form extends Injectable implements Countable, Iterator, AttributesInterfac
     }
 
     /**
+     * Sets the default validation
+     *
+     * @param ValidationInterface validation
+     */
+    public function setValidation(<ValidationInterface> validation) -> <Form>
+    {
+        let this->validation = validation;
+
+        return this;
+    }
+
+    /**
      * Sets the default whitelist
      *
-     * @param object entity
+     * @param array whitelist
      */
     public function setWhitelist(array whitelist) -> <Form>
     {
--- a/phalcon/Forms/Form.zep
+++ b/phalcon/Forms/Form.zep
@@ -79,6 +79,11 @@ class Form extends Injectable implements Countable, Iterator, AttributesInterfac
     protected validation = null { set, get };
 
     /**
+     * @var array
+     */
+    protected whitelist = [];
+
+    /**
      * Phalcon\Forms\Form constructor
      */
     public function __construct(var entity = null, array userOptions = [])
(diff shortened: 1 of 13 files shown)


phalcon/Validation.zep
```diff
@@ -238,57 +238,74 @@ class Validation extends Injectable implements ValidationInterface
...
+
+        if method_exists(entity, method) {
+            return entity->{method}();
+        } elseif method_exists(entity, "readAttribute") {
+            return entity->readAttribute(field);
+        } elseif isset entity->{field} {
+            return entity->{field};
+        } else {
+            return null;
+        }
```
Jeckerson, line 255:
```suggestion
        return null;
```

BeMySlaveDarlin:
Rebased due to fix conflicts

phalcon/Validation.zep
```diff
@@ -238,57 +238,80 @@ class Validation extends Injectable implements ValidationInterface
         return this->validators;
     }
 
+    private function getValueByEntity(entity, string field) -> var | null
```
Jeckerson, line 241: Please add PHP docs with types.

phalcon/Validation.zep
```diff
@@ -324,8 +347,8 @@ class Validation extends Injectable implements ValidationInterface
                 /**
                  * Set filtered value in entity
                  */
-                if typeof entity == "object" {
-                    let method = "set" . camelizedField;
+                if typeof entity == "object" && isRawFetched == false {
```
Jeckerson, line 350:
```suggestion
                if typeof entity == "object" && isRawFetched === false {
```

phalcon/Forms/Form.zep
```diff
@@ -598,7 +617,7 @@ class Form extends Injectable implements Countable, Iterator, AttributesInterfac
      * @param array data
      * @param object entity
      */
-    public function isValid(var data = null, var entity = null) -> bool
+    public function isValid(var data = null, var entity = null, var whitelist = null) -> bool
```
Jeckerson, line 620: Why not make `whitelist` always `array` with default value `[]`?

Jeckerson: Also, please update `@param` in docblock.

phalcon/Validation.zep
```diff
@@ -238,57 +238,80 @@ class Validation extends Injectable implements ValidationInterface
...
+        var value, values;
+
+        // Check if there is a calculated value
+        let values = this->values;
+
+        if fetch value, values[field] {
+            return value;
+        }
+
+        if typeof data == "array" {
```
Jeckerson, line 273:
```suggestion
        if typeof data === "array" {
```

phalcon/Validation.zep
```diff
@@ -238,57 +238,80 @@ class Validation extends Injectable implements ValidationInterface
...
+            if isset data[field] {
+                return data[field];
+            }
+        }
+
+        if typeof data == "object" {
```
Jeckerson, line 279:
```suggestion
        if typeof data === "object" {
```

Jeckerson (changes requested):

phalcon/Forms/Form.zep
```diff
@@ -78,6 +78,11 @@ class Form extends Injectable implements Countable, Iterator, AttributesInterfac
      */
     protected validation = null { set, get };
```
niden, line 79: Remove the `set` here, implement it in the method and return `this` so that it can be chained.

phalcon/Forms/Form.zep
```diff
@@ -78,6 +78,11 @@ class Form extends Injectable implements Countable, Iterator, AttributesInterfac
...
 
+    /**
+     * @var array
+     */
+    protected whitelist = [];
```
niden, line 84: add `{ get }` at the end and remove `getWhitelist()` below

phalcon/Forms/Form.zep
niden: `whitelist` needs to be an array here. One less check in the `if`.

phalcon/Forms/Form.zep
```diff
@@ -608,6 +627,10 @@ class Form extends Injectable implements Countable, Iterator, AttributesInterfac
             return true;
         }
 
+        if whitelist === null && !empty this->whitelist {
+            let whitelist = this->whitelist;
+        }
+
```
niden, line 633: And if the `whitelist` is an array, then the `if` conditions are redundant.

phalcon/Validation.zep
niden: Also please add `var entity`

niden (changes requested):

niden:
Other than the above changes, this is looking good. If you can please add a test or two.

BeMySlaveDarlin:
Fixed all comments. Rebased to 5.0.x resolving conflicts

BeMySlaveDarlin:
Rebased due to fix conflicts
Draft until all tests passes

niden (approved):

niden:
Thank you @BeMySlaveDarlin

PR #15764, opened 2021-11-04
T15361 find toarray

### LINKED ISSUE #15361: [BUG]: Mode::find() not casting numeric and float ->toArray()
Phalcon 4.1.0 / PHP 7.4.16 / PgSql 12 / Ubuntu 20.04

When I use find and call toArray the numeric and float fields are not casting, bool,string and int are fine,
also on findFirst is working. 

To work i have to do a foreach and create a new array, and each row calling toArray();

I did try with and without modelsMetadata.

I have try both configurations

```
phalcon.db.force_casting
phalcon.orm.cast_on_hydrate
```

```
 Model::setup([
          'forceCasting' => true,
          'castOnHydrate' => true
      ]);

```

Does not cast!

```
$objects = Model:find();
$object->toArray() // not casting numeric and float
```

Works if  you loop or call by a key.
`$object[0]->toArray();
` 

string,bool and int are ok.

niden:
Hello!

* Type: bug fix
* Link to issue: #15361 

Small description of change:

Fixed `Phalcon\Db\Adapter\Pdo\Mysql` to not use `PDO::ATTR_EMULATE_PREPARES` and `PDO::ATTR_STRINGIFY_FETCHES` by default. This allows numbers to be returned with resultsets instead of strings for numeric fields

### DIFF
--- a/phalcon/Db/Adapter/Pdo/Mysql.zep
+++ b/phalcon/Db/Adapter/Pdo/Mysql.zep
@@ -68,7 +68,7 @@ class Mysql extends PdoAdapter
     {
         /**
          * Returning numbers as numbers and not strings. If the user already
-         * set this option in the descriptor["options"], we do not have to set`
+         * set this option in the descriptor["options"], we do not have to set
          * anything
          */
         if (!isset(descriptor["options"][\PDO::ATTR_EMULATE_PREPARES])) {
--- a/phalcon/Db/Adapter/Pdo/Mysql.zep
+++ b/phalcon/Db/Adapter/Pdo/Mysql.zep
@@ -50,6 +50,38 @@ class Mysql extends PdoAdapter
     protected type = "mysql";
 
     /**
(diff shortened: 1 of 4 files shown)


phalcon/Db/Adapter/Pdo/Mysql.zep
```diff
@@ -49,6 +49,38 @@ class Mysql extends PdoAdapter
...
+     *     'options' => [],
+     *     'dsn' => null,
+     *     'charset' => 'utf8mb4'
+     * ]
+     */
+    public function __construct(array! descriptor)
+    {
+        /**
+         * Returning numbers as numbers and not strings. If the user already
+         * set this option in the descriptor["options"], we do not have to set`
```
Jeckerson, line 71:
```suggestion
         * set this option in the descriptor["options"], we do not have to set
```

Jeckerson (dismissed):

PR #15772, opened 2021-11-07
T15515 Fix allow empty validation

### LINKED ISSUE #15491: [BUG]: allowEmpty for form validation does not work correctly when an entity is bound
Using Phalcon 4.1.2 on PHP 7.4.18
Another big thing for me, sadly. Took me a while to figure it out.

Having a simple form:

```
$form = new Form();
$form->add(
    (new Text('name'))->addValidators([
        new Validation\Validator\Alnum([
            'message' => 'Invalid value'
            'allowEmpty' => true,
        ]),
    ])
);
$company = new Company();
if ($form->isValid($_POST, $company)) {
    // do something
}
```

No matter if `$_POST` contains a value for `name` or not, all is fine.
**But** I can't use `allowEmpty=true` in my projects, because this is also true for string value `0`, which is wrong IMHO. To not skip any validations, I have to use `allowEmpty=[null, '']` instead. When modifying the above code to:

```
$form = new Form();
$form->add(
    (new Text('name'))->addValidators([
        new Validation\Validator\Alnum([
            'message' => 'Invalid value'
            'allowEmpty' => [null,''],
        ]),
    ])
);
$company = new Companies();
if ($form->isValid($_POST, $company)) {
    // is not reached for all cases
}
```

the form validation for `Alnum` gives following results now:

- `$_POST['name'] = 'Acme';` = success
- `$_POST['name'] = '';` = success
- `$_POST['name'] = null;` = **fails**
- having `name` absent from `$_POST` = **fails**

When the form is **not** bound to the model, the validation does not fail for the previously mentioned cases. When `var_dump()`ing the validation object, I can see, that the property of the model is not `null` but `&null`. There seems to be any internal references. Maybe that's the reason why comparing `allowEmpty` values (see [here](https://github.com/phalcon/cphalcon/blob/4.1.2-release/phalcon/Validation.zep#L523)) fails.

I expect, that `allowEmpty = [null, '']` works for bound entities on forms too, even when the related property in entity is `null`.

### LINKED ISSUE #15515: [BUG]: Some Custom Validators don't handle the allowEmpty
I'm trying to use the built-in Date validator in 4.1.2, and it seems to be ignoring the "allowEmpty" option.

Something analogous to this still generates an error when `payment_2_date` is NULL;
```php
$Validator->add(
	[
		'payment_1_date',
		'payment_2_date'
	],
	new Date([
		'format'=>[
			'payment_1_date'=>'Y-m-d',
			'payment_2_date'=>'Y-m-d H:i:s'
		]
		'message'=>[
			'payment_1_date'=>'Payment 1 date is not valid',
			'payment_2_date'=>'Payment 2 date is not valid'
		]
		'allowEmpty'=>[
			'payment_1_date'=>FALSE,
			'payment_2_date'=>TRUE
		]
	])
);
```

Looking at the [Zephir source for the Date validator](https://github.com/phalcon/cphalcon/blob/master/phalcon/Validation/Validator/Date.zep), it has `allowEmpty()` as an option in the comments for `__construct()`, but doesn't act on it anywhere.  I also checked the `Alnum` and `Alpha` validators - same deal.

I know comments in Zephir source code aren't considered "public documentation", but there's been a mistake somewhere - either in copy-pasting the comments, or omitting this necessary functionality.

BeMySlaveDarlin:
Hello!

* Type: bug fix
* Link to issue: #15515

Small description of change:
Fixed usage of `allowEmpty` option in common validation.

Thanks

### DIFF
--- a/phalcon/Validation/AbstractValidator.zep
+++ b/phalcon/Validation/AbstractValidator.zep
@@ -211,6 +211,24 @@ abstract class AbstractValidator implements ValidatorInterface
     }
 
     /**
+     * Checks if field can be empty.
+     *
+     * @return bool
+     */
+    protected function allowEmpty(var field, var value) -> bool
+    {
+        var allowEmpty;
+
+        let allowEmpty = this->getOption("allowEmpty", false);
+
+        if typeof allowEmpty == "array" {
+            let allowEmpty = isset allowEmpty[field] ? allowEmpty[field] : false;
+        }
+
+        return allowEmpty && empty value;
+    }
+
+    /**
      * Create a default message by factory
      *
      * @return Message
--- a/phalcon/Validation/AbstractValidator.zep
+++ b/phalcon/Validation/AbstractValidator.zep
@@ -213,6 +213,9 @@ abstract class AbstractValidator implements ValidatorInterface
     /**
      * Checks if field can be empty.
      *
+     * @param mixed field
+     * @param mixed value
+     *
      * @return bool
      */
     protected function allowEmpty(var field, var value) -> bool
@@ -221,7 +224,7 @@ abstract class AbstractValidator implements ValidatorInterface
 
         let allowEmpty = this->getOption("allowEmpty", false);
 
-        if typeof allowEmpty == "array" {
+        if typeof allowEmpty === "array" {
             let allowEmpty = isset allowEmpty[field] ? allowEmpty[field] : false;
         }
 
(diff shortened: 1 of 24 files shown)


phalcon/Validation/AbstractValidator.zep
```diff
@@ -210,6 +210,24 @@ abstract class AbstractValidator implements ValidatorInterface
         return label;
     }
 
+    /**
+     * Checks if field can be empty.
+     *
+     * @return bool
+     */
+    protected function allowEmpty(var field, var value) -> bool
```
Jeckerson, line 218: Missing types in docblock.

phalcon/Validation/AbstractValidator.zep
```diff
@@ -210,6 +210,24 @@ abstract class AbstractValidator implements ValidatorInterface
...
+    {
+        var allowEmpty;
+
+        let allowEmpty = this->getOption("allowEmpty", false);
+
+        if typeof allowEmpty == "array" {
```
Jeckerson, line 224:
```suggestion
        if typeof allowEmpty === "array" {
```

niden (approved):

niden:
Thank you @BeMySlaveDarlin

PR #15781, opened 2021-11-09
T15413 Fixed raw value bug in lowupdate

### LINKED ISSUE #15413: [BUG]: Dynamic update model issue
**Describe the bug**
Same as [#12773](https://github.com/phalcon/cphalcon/issues/12773),  when field type is `Decimal` or `Float`。

If old value in database is 1.0, then use RawValue to update this field, will reproduce this bug. Because of following logic
```
case Column::TYPE_DECIMAL:
case Column::TYPE_FLOAT:
    let changed = floatval(snapshotValue) !== floatval(value);
    break;
```

Because:
```
floatval(new \Phalcon\Db\RawValue("f+1")) === float(1)
```

So to reproduce the issue, should be 

1. snapshotValue is 1, 
2. use  RawValue to update this field
3. the field type is DECIMAL or FLOAT

BeMySlaveDarlin:
Hello!

* Type: bug fix
* Closes #15413

Small description of change:

Thanks

### DIFF
--- a/phalcon/Mvc/Model.zep
+++ b/phalcon/Mvc/Model.zep
@@ -3900,6 +3900,16 @@ abstract class Model extends AbstractInjectionAware implements EntityInterface, 
                                         );
                                     }
 
+                                    /**
+                                     * Get actual values before comparison
+                                     */
+                                    if snapshotValue instanceof RawValue {
+                                        let snapshotValue = snapshotValue->getValue();
+                                    }
+                                    if value instanceof RawValue {
+                                        let value = value->getValue();
+                                    }
+
                                     switch dataType {
 
                                         case Column::TYPE_BOOLEAN:
--- a/phalcon/Mvc/Model.zep
+++ b/phalcon/Mvc/Model.zep
@@ -3903,10 +3903,10 @@ abstract class Model extends AbstractInjectionAware implements EntityInterface, 
                                     /**
                                      * Get actual values before comparison
                                      */
-                                    if snapshotValue instanceof RawValue {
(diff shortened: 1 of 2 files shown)


phalcon/Mvc/Model.zep
```diff
@@ -3900,6 +3900,16 @@ abstract class Model extends AbstractInjectionAware implements EntityInterface,
                                         );
                                     }
 
+                                    /**
+                                     * Get actual values before comparison
+                                     */
+                                    if is_object(snapshotValue) && snapshotValue instanceof RawValue {
```
Jeckerson, line 3906:
```suggestion
                                    if snapshotValue instanceof RawValue {
```

phalcon/Mvc/Model.zep
```diff
@@ -3900,6 +3900,16 @@ abstract class Model extends AbstractInjectionAware implements EntityInterface,
...
+                                        let snapshotValue = snapshotValue->getValue();
+                                    }
+                                    if is_object(value) && value instanceof RawValue {
```
Jeckerson, line 3909:
```suggestion
                                    if value instanceof RawValue {
```

phalcon/Mvc/Model.zep
BeMySlaveDarlin: `instanceof` expects an object instance, but `snapshotValue` can be other type

phalcon/Mvc/Model.zep
BeMySlaveDarlin: `instanceof` expects an object instance, but `value` can be other type

niden (approved):

PR #15843, opened 2021-12-28
T15842 volt compiler

### LINKED ISSUE #15842: [BUG]: Volt Compiler throws RuntimeException - container related
When registering a custom function for Volt, there is an error thrown:

Trying to call method "has" on an non object

The issue appears when there is no container injected in Volt and as such line 1881:

`if true === this->container->has("tag") {`

throws the error. It should be changed to:

`if typeof this->container == "object" && this->container->has("tag") {`

niden:
Hello!

* Type: bug fix 
* Link to issue: #15842 

Corrected Volt Compiler `functionCall` to first check for the container object before checking for the `tag` service.

Thanks

### DIFF
--- a/phalcon/Mvc/View/Engine/Volt/Compiler.zep
+++ b/phalcon/Mvc/View/Engine/Volt/Compiler.zep
@@ -233,7 +233,7 @@ class Compiler implements InjectionAwareInterface
                  * Services registered in the dependency injector container are
                  * available always
                  */
-                if typeof this->container == "object" && this->container->has(variable) {
+                if this->container !== null && this->container->has(variable) {
                     let exprCode .= "$this->" . variable;
                 } else {
                     let exprCode .= "$" . variable;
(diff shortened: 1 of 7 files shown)


phalcon/Mvc/View/Engine/Volt/Compiler.zep
```diff
@@ -1878,7 +1878,7 @@ class Compiler implements InjectionAwareInterface
             /**
              * These are for the TagFactory
              */
-            if true === this->container->has("tag") {
+            if typeof this->container == "object" && true === this->container->has("tag") {
```
Jeckerson, line 1881:
```suggestion
            if this->container !== null && true === this->container->has("tag") {
```

niden: Resolved

Jeckerson (approved):

PR #15983, opened 2022-06-08
T15978 html string 3

### LINKED ISSUE #15978: [BUG]: Filter::StringVal with <<>>
`FILTER_SANITIZE_STRING` is to be removed from PHP 8.1. The `StringVal` filter relies on `filter_var()` to sanitize strings. 

As per the documentation as well as the relevant RFC that removes the constant, `htmlspecialchars()` is a better alternative to the functionality.

The current implementation - which no longer uses `filter_var` is based on `preg_match` is:

```php
public function __invoke($input)
{
    $input = str_replace(chr(0), "", (string) $input);
    $input = preg_replace("/<[^>]*>?/", "", $input);

    return str_replace(["'", "\""], ["&#39;", "&#34;"], $input);
}
```

When using the following string:

```php
$input = '<< >> ' ' ' ' " " " " < >  - -  ... '
```

the expected result is:

```php
' &#39; &#39; &#39; &#39; &#34; &#34; &#34; &#34;   - -  ... '
```
but instead we get:

```php
'> &#39; &#39; &#39; &#39; &#34; &#34; &#34; &#34;   - -  ... '
```

Options: 
1. correct the `preg_match` to pick up the `>`
2. Change the `StringVal` filter to use `htmlspecialchars()` and also introduce `StringValLegacy` as a new filter which will only work for PHP 7.4 and 8.0 and will utilize `filter_var()`. This way we have something that people can use as they have been, and offer a path to upgrade the filter within the next major version.

niden:
Hello!

* Type: bug fix 
* Link to issue: #15978 

Corrected the `StringVal` filter to now use `htmlspecialchars()`. Added `StringValLegacy` filter with the same functionality as before for PHP < 8.1

Thanks


phalcon/Filter/Sanitize/StringValLegacy.zep
```diff
@@ -0,0 +1,37 @@
...
+     * @param string $input The text to sanitize
+     *
+     * @return string|false
+     */
+    public function __invoke(mixed input)
+    {
+        var version;
+
+        let version = phpversion();
+        if version_compare(version, "8.1", "<") {
```
Jeckerson, line 31: I suggest to use `PHP_VERSION` constant, which might be slightly faster comparing to calling of `phpversion()`.

Jeckerson (approved):

Issue #16040, opened 2022-08-04
[BUG]: AbstractSerializer isSuccess never getting reset to true


LlamaDuckGoose:
**Describe the bug**
Since AbstractSerializer isSuccess is initiated with the class and is set as true and is only is updated to false, if it fails to unserialize one bit of data the rest will also return isSuccess as false but getData will return valid data.

**To Reproduce**
This example isn't the best but it should give you a way to recreate it:

```php
$this->_di->setShared('cache', function () {
  $redisCache =  new \Phalcon\Cache\Adapter\Redis(
      new \Phalcon\Storage\SerializerFactory()),
      $redisConfig
  );

  return new \Phalcon\Cache\Cache($redisCache);
});

$cache = \Phalcon\Di\Di::getDefault()->getShared('cache');

for ($x = 0; $x <= 5; $x++) {
    var_dump($x);
    if ($x < 3 || $x > 3) {
        if ($cache->has('testing_1')) {
            var_dump($cache->get('testing_1'));
            $cache->set('testing_1', ['fake' => 'data']);
        } else {
            var_dump('no cache');
            $cache->set('testing_1', ['fake' => 'data']);
        }
    } else {
        $cache->set('testing_1', false);
        var_dump($cache->get('testing_1'));
    }
}

var_dump('end');
die();
```
You can see, if the first loop, no cache so we set it an an array, 2-3 loop, it returns the cache, 4th loop it set it as false which on line phalcon\Storage\Serializer\Php.zep::70 checks if the result === false which causes it to set isSuccess as false. and then because of that the rest are returning null cause it's failing on phalcon\Storage\Adapter\AbstractAdapter.zep::328 since isSuccess is now false.
![image](https://user-images.githubusercontent.com/55490546/182761351-701fff9f-1b7b-4f2e-a85e-fe3e16e0d705.png)

**Details**
 - Phalcon version: 5.0.0RC3
 - PHP Version: 7.4.30
 - Operating System: Ubuntu 20.04.4
 - Installation type: Any
 - Zephir version (if any): NA
 - Server: Apache

niden:
Resolved in https://github.com/phalcon/cphalcon/pull/16041/

PR #16041, opened 2022-08-04
[BUG]: AbstractSerializer isSuccess never getting reset to true

ALameLlama:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/16040

Small description of change:
Since AbstractSerializer isSuccess is initiated with the class and is set as true and is only is updated to false, if it fails to unserialize one bit of data the rest will also return isSuccess as false but getData will return valid data.

Thanks


niden (approved):

niden:
Thank you @LlamaDuckGoose

Issue #16042, opened 2022-08-05
[BUG]: Paginator Model Adapter Incorrect total items when grouped


LlamaDuckGoose:
**Describe the bug**
This is currently always returning 1 for total items when grouped which also breaks the last page/total page count as well.

**To Reproduce**
Create a model paginator without passing a group in the parameters and then one with group.

```php
    $paginator = new \Phalcon\Paginator\Adapter\Model([
        'model' => Model::class,
        'parameters' => [
            'type IN ({type:array})',
            'bind' => [
                'type' => [
                    Model::TYPE_1,
                    Model::TYPE_2,
                ],
            ],
            'group' => 'type'
        ]
    ]);
```

Since this is grouped by type and 2 types are provided, you'd expect it to return total items as 2 but this will return 1

**Details**
 - Phalcon version: 5.0.0RC3
 - PHP Version: 7.4.30
 - Operating System: Ubuntu 20.04.4
 - Installation type: Any
 - Zephir version (if any): NA
 - Server: Apache

niden:
Resolved in https://github.com/phalcon/cphalcon/pull/16043/

PR #16043, opened 2022-08-05
[BUG]: Paginator Model Adapter Incorrect total items when grouped

ALameLlama:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/16042

Small description of change:
when the parameters include a group, it will return a simple result instead of an int.

This is currently getting type cast to int which will return 1 for simple result.
`call_user_func([modelClass, "count"], parameters);`

Now it will check if it's an object and if it is get the count from that.

Thanks


niden (approved):

niden:
Thank you @LlamaDuckGoose

PR #16179, opened 2022-10-24
adjust default for form::label attributes

### LINKED ISSUE #16180: [BUG]: Form::label parameters
`element::label` is updated from `array attributes = null` to `array attributes = []` 

Using this method with the default secondary argument will now cause an 

"Fatal error: Uncaught TypeError" to be thrown

1RV34:
Hello!

* Type: bug fix / code quality

Small description of change:
since element::label is updated from `array attributes = null` to `array attributes = []` using this method with the default secondary argument will now cause an "Fatal error: Uncaught TypeError" to be thrown

Thanks

### DIFF
--- a/phalcon/Forms/Form.zep
+++ b/phalcon/Forms/Form.zep
@@ -754,7 +754,7 @@ class Form extends Injectable implements Countable, Iterator, AttributesInterfac
     /**
      * Generate the label of an element added to the form including HTML
      */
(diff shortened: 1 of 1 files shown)


niden (approved):

niden:
Thank you @1RV34

1RV34:
@niden one note, this issue seems to have been introduced in 4.0, it would be nice to backport a patch for 4.0 as well

PR #16197, opened 2022-11-03
Fix no recursive array value with multiple sanitizers in Filter

### LINKED ISSUE #16199: [BUG]: Filter:sanitize array - noRecursive
Hey, I was going after the new version and found that there seems to be an bug when we call sanitize with an array value, array sanitizers and noRecursive = true. The output is always the same as the initial value.

I think this change should do it as it goes to a [previous way](https://github.com/phalcon/cphalcon/pull/15734/files#diff-c589afb489a0f55dca8b29abf141bcf4d31783462ab3bd4e1780a4a86c8d5f7aL173-L179) to handle this.

Credit: @lubatti

lubatti:
Hello!

* Type: bug fix
* Link to issue: -

Small description of change:
Hey, I was going after the new version and found that there seems to be an bug when we call `sanitize` with an array `value`, array `sanitizers` and `noRecursive = true`. **The output is always the same as the initial value**.
I think this change should do it as it goes to a [previous way](https://github.com/phalcon/cphalcon/pull/15734/files#diff-c589afb489a0f55dca8b29abf141bcf4d31783462ab3bd4e1780a4a86c8d5f7aL173-L179) to handle this.

Thanks

### DIFF
--- a/phalcon/Filter/Filter.zep
+++ b/phalcon/Filter/Filter.zep
@@ -286,18 +286,20 @@ class Filter implements FilterInterface
              * has been defined it is a straight up; otherwise recursion is
              * required
              */
-            let value = this->processValueIsArray(
-                value,
-                sanitizerName,
-                sanitizerParams,
-                noRecursive
-            );
-
-            let value = this->processValueIsNotArray(
-                value,
-                sanitizerName,
-                sanitizerParams
-            );
+             if typeof value === "array" {
+                let value = this->processValueIsArray(
+                    value,
+                    sanitizerName,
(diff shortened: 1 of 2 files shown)


niden:
@lubatti Thank you for this change but it seems that it double encodes values. See the tests.

lubatti:
> @lubatti Thank you for this change but it seems that it double encodes values. See the tests.

Fixed, and added new test

niden (approved):

niden:
Thank you @lubatti

PR #16279, opened 2023-02-07
Update Route.zep

lieshanhe:
update the description of cli route. previously, it was Mvc route usage.

It's not a bug but just something regarding documentation.

### DIFF
--- a/phalcon/Cli/Router/Route.zep
+++ b/phalcon/Cli/Router/Route.zep
@@ -547,12 +547,9 @@ class Route implements RouteInterface
      * Sets the route's name
      *
      *```php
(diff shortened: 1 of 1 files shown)


phalcon/Cli/Router/Route.zep
```diff
@@ -547,12 +547,9 @@ class Route implements RouteInterface
      * Sets the route's name
      *
      *```php
-     * $router->add(
-     *     "/about",
-     *     [
-     *         "controller" => "about",
-     *     ]
+     *
      * )->setName("about");
```
Jeckerson, line 551: This line should be also removed.

PR #16347, opened 2023-05-30
T16339 stream exception

### LINKED ISSUE #16339: [BUG]: Random exceptions when retrieving data using Stream + Json
**Describe the bug**
The Json serialiser uses `if unlikely` when determining if json_decode gives an error or not. This is faster, but can give unexpected results. https://igoro.com/archive/fast-and-slow-if-statements-branch-prediction-in-modern-processors/.

This is no problem for caching scenario's, but the Stream storage adapter may as well be used for permanent storage. 

**To Reproduce**
This is not easy to reproduce with just a simple script, you need to run a lot of writing / reading of Streams over some time. Then exceptions will appear.

**Expected behavior**
Not have random errors when reading stream data with the Json adapter.
Of course this is easily fixed by simply using `if` instead of `if unlikely`, but this probably more a design philosophy question.

**Details**
 - Phalcon version: 4.2 (But still present in 5.x)
 - PHP Version: 7.4
 - Operating System: Linux
 - Installation type: Cloudlinux
 - Server: Apache

niden:
Hello!

* Type: bug fix 
* Link to issue: #16339 

Removed `unlikely` from `if` statements from the Stream storage adapter and Json serializer

Thanks

### DIFF
--- a/CHANGELOG-5.0.md
+++ b/CHANGELOG-5.0.md
@@ -5,6 +5,7 @@
 ### Fixed
  
 - Fixed `Encryption\Crypt::checkCipherHashIsAvailable` to allow proper setting of the hash [#16314](https://github.com/phalcon/cphalcon/issues/16314) 
+- Removed `unlikely` from `if` statements from the Stream storage adapter and Json serializer [#16339](https://github.com/phalcon/cphalcon/issues/16339)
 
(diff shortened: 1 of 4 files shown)


niden:
PR to the correct branch now

PR #16377, opened 2023-07-17
added Weak Storage and Cache, for objects only using WeakReference

### LINKED ISSUE #16372: [NFR]: Add a WeakCache using WeakReference
The idea is to add a key => value Cache service that uses a WeakReference Object has value.

WeakReference was implemented in PHP 7.4, it uses creates an objects that has a reference to an object but does not prevent it from being destroyed by the Garbage Collector.

So we can have a key => value cache that will return an object if it exists anywhere in memory and has not been destroyed by the GB.

This will help solve issues with the ORM in the future.

For more information about WeakReference

https://www.php.net/manual/en/class.weakreference.php

rudiservo:
Hello!

* Type: new feature
* Link to issue: #16372 

Small description of change:

Added a object only in memory cache/storage system using WeakReference.

Thanks


phalcon/Storage/Adapter/Weak.zep
```diff
@@ -0,0 +1,235 @@
...
+        let this->fetching = null;
+        /**
+         * value could be null, object could be destroid while fetching
+         */
+        if null === value {
+            this->delete(key);
+        }
+        return value;
+    }
+
```
niden, line 167: I don't think the `doGet` is required here. The `get()` in the abstract class checks if the key exists then if not returns the default value and then it performs the `doGet()`

So I was wondering if the above (get) can be adjusted so that the code is used from the abstract class and when the actual `get` happens, that goes in the `doGet()` method (including the `delete() part that is in line 162

rudiservo: Maybe it's not required, Weak does not have an Adapter, so this->adapter in the AbstractAdapter doGet() wont work.
I checked this against Storage\Adapter\Memory that currently is like this.

```
    protected function doGet(string key)
    {
        return this->data[this->getPrefixedKey(key)];
    }

```

I think it's like this mainly for compatibility reasons.
maybe I can scrap it.

Has for the get(), my reasoning is this is a bit optimized, I could scrape it and use the AbstractAdapter::get(), make the get() into the doGet() and set the serializer to null, just trying to avoid a jump to has() and another jump to getUnserializedData() just to return content unaltered.
Trying to save 2 or 3 jumps, in my use case is going to use Weak every time ResultSet::current() is called.

I could do something more optimized in the model later on.

What do you think?

niden: If we are to override `get` and have it run from the Weak, then no need for the `doGet`. 

I would go with that really since the abstract `get` will never be called

rudiservo: OK, if you are ok with this, I will remove the doGet().
Thanks!

niden:
@rudiservo Don't forget to add a line to the CHANGELOG :)

niden (approved):

niden:
Thank you @rudiservo

Issue #16391, opened 2023-08-04
[NFR]: Small refactoring of Model


rudiservo:
Model::postSaveRelatedRecords has a lot of many-to-many complex logic mixed with normal to-many , making a clear separation of both will be easier to debug and make less checks during execution.
Model messages can be wrapped into a function that will clear up some code and add functionality.
Model::doLowUpdate could benefit a better seperation of DynamicUpdate from regular update all fields.

doLowUpdate probably does not need to add this portion in the dynamic update, if the field isn't set, no change has happened.

```
                        let fields[]    = field,
                            values[]    = null,
                            bindTypes[] = bindSkip;
                    }
```
Better Error reporting in models by adding model name and/or referenceModel to exception

Better Message reporting in models by adding model name and/or referenceModel to Message::MetaData

niden:
Thank you @rudiservo

PR #16392, opened 2023-08-06
Refactored Model, added Loader::isRegistered, fixed Metadata behaviour

rudiservo:
Hello!

* Link to issue: [#16391]

Small description of change:

Refactored the Model::doLowUpdate for better seperation of dynamicUpdate from normal full update.
Added better error reporting adding the class on which it happened.
Added Model::appendMessagesFrom for code quality and added the model name to the message MetaData
Added Model and relatedName to Message MetaData in several lines where there was no reference to model and/or relation where the non-exception failure occurred.

Thanks

### DIFF
--- a/phalcon/Mvc/Model/MetaData.zep
+++ b/phalcon/Mvc/Model/MetaData.zep
@@ -484,10 +484,10 @@ abstract class MetaData implements InjectionAwareInterface, MetaDataInterface
         let columnMap = this->getReverseColumnMap(model);
 
         if typeof columnMap == "array" {
-            return isset columnMap[attribute];
+            return isset(columnMap[attribute]);
         }
 
-        return isset this->readMetaData(model)[self::MODELS_DATA_TYPES][attribute];
+        return isset(this->readMetaData(model)[self::MODELS_DATA_TYPES][attribute]);
     }
 
     /**
@@ -525,21 +525,16 @@ abstract class MetaData implements InjectionAwareInterface, MetaDataInterface
      */
     final public function readColumnMap(<ModelInterface> model) -> array | null
     {
-        var keyName, data;
+        var keyName;
 
         if !globals_get("orm.column_renaming") {
             return null;
         }
-
-        let keyName = get_class_lower(model);
-
-        if !fetch data, this->columnMap[keyName] {
-            this->initialize(model, null, null, null);
-
-            let data = this->columnMap[keyName];
+        let keyName = this->getColumnMapUniqueKey(model);
+        if likely keyName !== null {
+            return this->columnMap[keyName];
         }
-
-        return data;
+        return null;
     }
 
     /**
@@ -554,25 +549,18 @@ abstract class MetaData implements InjectionAwareInterface, MetaDataInterface
      * );
      *```
      */
-    final public function readColumnMapIndex(<ModelInterface> model, int index)
+    final public function readColumnMapIndex(<ModelInterface> model, int index) -> array | null
     {
-        var keyName, columnMapModel, map;
+        var keyName;
 
         if !globals_get("orm.column_renaming") {
             return null;
         }
-
-        let keyName = get_class_lower(model);
-
-        if !fetch columnMapModel, this->columnMap[keyName] {
-            this->initialize(model, null, null, null);
-
-            let columnMapModel = this->columnMap[keyName];
+        let keyName = this->getColumnMapUniqueKey(model);
+        if likely keyName !== null {
+            return this->columnMap[keyName][index];
         }
-
-        fetch map, columnMapModel[index];
-
-        return map;
+        return null;
     }
 
     /**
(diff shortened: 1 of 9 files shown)


niden:
@rudiservo Can you split that into much smaller PRs please? Say one for the loader, one for the messages changed/exceptions, one for metadata etc.

There are a lot of areas that I do not understand what the goal of the changes are, for instance the keys for metadata are no longer held in lowercase.

rudiservo:
@niden sure, I did tried to make seperate pull requests, github just put it into one dont know why.
If you don't mind giving me a hint on how to make it, and if there is an easy way to just split code from a commit?

Metadata changes are on their own commit.

Metadata keys are in lowercase, using get_class_lower(), what they are is no longer concat with the schema and source in case of the actual MetaData, ColumnMap was always lowercase but it used two functions, get_class() and strtolower().

Schema and source are pulled from the manager, and the manager associates them with the get_class_lower() only, makes no sense to make a metadata key with all that when the information is only associated with the full namespace class itself.

In the Model Messages the changes are for better error reporting, many times I do not know where the error is and I have to do a game of var_dump() die() or figuring out where in xdebug trace it failed.

The postSaveRelatedRecords is to seperate the logic through from normal, it makes code analysis simpler has well has less checks, maybe slight performance improvement.
The doLowUpdate is to seperate dynamicUpdate logic from normal full update, this helps a lot also for code analysis and has less checks, it either is or is not, maybe slight performance improvement.
Both are not behaviour changes, just code reorganization.

rudiservo:
And metadata::initialize was overblown with code that was not supposed to be there, now it's cleaner, probably very very very very very small performance improvement.

niden (approved):

phalcon/Mvc/Model/MetaData.zep
```diff
@@ -525,21 +525,13 @@ abstract class MetaData implements InjectionAwareInterface, MetaDataInterface
...
-
-        if !fetch data, this->columnMap[keyName] {
-            this->initialize(model, null, null, null);
-
-            let data = this->columnMap[keyName];
-        }
-
-        return data;
+        let keyName = this->getColumnMapUniqueKey(model);
+        return this->columnMap[keyName];
```
Jeckerson, line 534: This might generate notice error. Use ternary condition:
```
isset(this->columnMap[keyName]) ? this->columnMap[keyName] : null;
```

phalcon/Mvc/Model/MetaData.zep
```diff
@@ -556,23 +548,13 @@ abstract class MetaData implements InjectionAwareInterface, MetaDataInterface
...
-            this->initialize(model, null, null, null);
-
-            let columnMapModel = this->columnMap[keyName];
-        }
-
-        fetch map, columnMapModel[index];
-
-        return map;
+        let keyName = this->getColumnMapUniqueKey(model);
+        return this->columnMap[keyName][index];
```
Jeckerson, line 557: This might generate notice error. Use ternary condition:
```
isset(this->columnMap[keyName]) ? this->columnMap[keyName] : null;
```

phalcon/Mvc/Model/MetaData.zep
```diff
@@ -848,47 +794,52 @@ abstract class MetaData implements InjectionAwareInterface, MetaDataInterface
...
+     */
+    final protected function initializeColumnMap(<ModelInterface> model, key)
+    {
+        var strategy, data, modelColumnMap, container;
+        string prefixKey;
         /**
          * Check for a column map, store in columnMap in order and reversed order
          */
-        if !globals_get("orm.column_renaming") {
+         if !globals_get("orm.column_renaming") {
```
Jeckerson, line 809:
```suggestion
        if !globals_get("orm.column_renaming") {
```

phalcon/Mvc/Model/MetaData.zep
rudiservo: actually we only need to check if the keyName is not null.

niden: What if the key does not exist? That would throw a warning no?

rudiservo: it already throws errors in other places 
The model does this
```
        let dataTypes           = metaData->getDataTypes(this),
            bindDataTypes       = metaData->getBindTypes(this),
            nonPrimary          = metaData->getNonPrimaryKeyAttributes(this),
            automaticAttributes = metaData->getAutomaticUpdateAttributes(this);
```

and all of them throws exceptions in the metadata.
```
        if unlikely typeof data != "array" {
            throw new Exception("The meta-data is invalid or is corrupt");
        }
```

do you want me to send warnings to php?

niden: Throwing exceptions is fine (for instance if `data` is not an array as shown above). I would prefer an extra check to happen if the key of the array does not exist.

Jeckerson (approved):

Issue #16222, opened 2022-11-22
Saving related entity in Phalcon 5


slechtic:
### Discussed in https://github.com/phalcon/cphalcon/discussions/16203

<div type='discussions-op-text'>

<sup>Originally posted by **slechtic** November 11, 2022</sup>
Hi,

I'm trying to migrate from Phalcon v3 to v5 but I have a problem with updating related entity.

I have an entity Document and this entity has related entity DocumentHistory.

On Document entity there is hasMany relationship defined:

`$this->hasMany('id', DocumentHistory::class, 'idDocument', ['alias' => 'documentHistories']);`

DocumentHistory entity has belongsTo relationship:

`$this->belongsTo('idDocument', Document::class, 'id',['alias' => 'document']);`

when I create new DocumentHistory entity and assign existing Document entity, change some values on this entity (Document) and save DocumentHistory entity, only DocumentHistory is saved (created). Document entity is not updated but foreign key is set on DocumentHistory entity correctly. In case when both entities exist it works correctly.

```
$documentHistory = new DocumentHistory();

$document = Document::findFirst($id);
$document->setNumber($number);

$documentHistory->setDocument($document);
$documentHistory->save();
```

On DocumentHistory there is a setter:

```
public function setDocument(Document $document) {
    $this->document = $document;
}
```

This works for Phalcon v3. This will not be supported in v5 or it is a bug? Thank you.

My environent:

PHP 8.1.
Phalcon 5.1.0.
Postgresql 11</div>

slechtic:
Hi guys, did you have any chance to check this problem? We need to know whether it is a bug or expected behavior so that we can proceed with Phalcon 5 upgrade.

Jeckerson:
Did you tried to migrate to v4 first?

slechtic:
> Did you tried to migrate to v4 first?

Yes I did. But there was a problem with segmentation fault and  suggestion was migrate to v5.  https://github.com/phalcon/cphalcon/discussions/16028

slechtic:
Here you can test it https://github.com/slechtic/phalcon5_relationships

markofo:
Hi @Jeckerson 
could you please let us know whether this is expected behavior, bug, or we are using it wrong?

In the previous post from @slechtic, there is a docker image with the example of the issue where you can easily test it.

Is there someone from Phalcon team who can help us even as a paid consultation on this matter?

Thank you in advance!

Marek

slechtic:
[...]

PR #16402, opened 2023-08-15
Fix #[16222], Added Model::setRelated()

rudiservo:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/16222

Small description of change:
Added Model::setRelated() for those who want to make a setter for relations or avoid magic __set(), changed the magical __set() to use setRelated()

Thanks

### DIFF
--- a/CHANGELOG-5.0.md
+++ b/CHANGELOG-5.0.md
@@ -1,6 +1,9 @@
 # Changelog
 
-## [5.4.0](https://github.com/phalcon/cphalcon/releases/tag/v5.3.1) (xxxx-xx-xx)
+## [5.4.0](https://github.com/phalcon/cphalcon/releases/tag/v5.4.0) (xxxx-xx-xx)
+
+### Added
+- Added `Phalcon\Mvc\Model::setRelated()` to allow setting related models and automaticly de added to the dirtyRelated list [#16222] (https://github.com/phalcon/cphalcon/issues/16222)
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


niden:
Addressed in https://github.com/phalcon/cphalcon/pull/16970

Thank you @rudiservo

rudiservo:
I don't think it's the same thing @niden

SetRelated can be called from anypoint the developer wants to bypass __set and have the relation set, the performance implications are still significant if people want to not use __set for the logic of setting a related item to be in the dirty list.
I would reconsider this.

niden:
> I don't think it's the same thing @niden
> 
> SetRelated can be called from anypoint the developer wants to bypass __set and have the relation set, the performance implications are still significant if people want to not use __set for the logic of setting a related item to be in the dirty list. I would reconsider this.

@rudiservo How would that work then? I get a record then I can add setRelated (a collection of sorts) and then what? save/update/delete?

Do you have an example that I can see what you have in mind? The first thing that I understood when reading this was the usage of the model with related data as a value object but I don't think that is what you need.

rudiservo:
the example is has follows

```
$customer->setRelated("invoices", $invoices);
$customer->save();
```

it works normally, the developer experience remains the same, it just gives the option to not use magical __set
Also is some ground work for future logic for appending relations that I set forth a few years ago.

So in the goal is
```
$customer->GetInvoices()->append($invoice);
```

So the logic in the relation would call `$invoice->setRelated("customer", $customer)` because there is a backwards reference to the parent relation.


we also get a measurable performance boost for skipping __set specially in heavy workloads.
 I think also the logic is better and probably faster.

PR #16450, opened 2023-10-02
Fix #16411, added ORM SessionCache AKA FirstLevelCache

### LINKED ISSUE #16411: [NFR]: ORM: Model Thread Cache, AKA First Level Cache or L1 cache
**Is your feature request related to a problem? Please describe.**
The ORM does not have a single source of truth when it fetches a new model.

That means that fetching the same model data from the database actually gives you two different instance objects of the same data.

This leads to false data, data losses, incorrect saves and longer saves.

**Describe the solution you'd like**
A thread cache that has all the models that are in memory, if that model with that Primary key is fetched the Thread cache will return the already instance model in memory.

If the model does not exists any longer in memory then a new instance will be made with fresh data.

**Describe alternatives you've considered**
There are none, this is by default in most ORMs, in java Hibernate it's not even an option, it is always on.

**Additional context**
https://www.linkedin.com/advice/3/what-benefits-drawbacks-using-orm-caching
https://howtodoinjava.com/hibernate/understanding-hibernate-first-level-cache-with-example/

rudiservo:
Hello!

* Type: bug fix | new feature | code quality
* Link to issue: #16411 

Small description of change:

Thanks

### DIFF
--- a/config.json
+++ b/config.json
@@ -153,6 +153,10 @@
       "type": "bool",
       "default": true
     },
+    "orm.session_cache": {
+      "type": "bool",
+      "default": false
+    },
     "warning.enable": {
       "type": "bool",
       "default": true
--- a/config.json
+++ b/config.json
@@ -165,6 +165,10 @@
       "type": "bool",
       "default": true
     },
+    "orm.session_cache": {
+      "type": "bool",
+      "default": false
+    },
     "warning.enable": {
       "type": "bool",
       "default": true
--- a/phalcon/Mvc/Model.zep
+++ b/phalcon/Mvc/Model.zep
@@ -164,6 +164,11 @@ abstract class Model extends AbstractInjectionAware implements EntityInterface, 
     protected uniqueTypes = [];
 
     /**
+     * @var string
+     */
+     protected modelUUID;
+
+    /**
      * Phalcon\Mvc\Model constructor
      */
     final public function __construct(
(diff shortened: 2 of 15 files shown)


rudiservo:
SqLite tests fail, the tests are fixed in one of the PR I previously made, I will rebase this once they are approved.
feel free to check and comment.

niden:
@rudiservo I had some time (finally) to check this one. 

I see the benefit and am not opposed to it but there are some concerns I have in particular with the __destruct part and orphan references. 

We can certainly add checks for objects that are null (modelsManager etc.) and that is fine but I am worried about the internal dependencies of objects and the references that they hold to other objects. Therefore the `__destruct` is the weak point here (and we don't have tests for that). If for whatever reason, the DI container is destructed before the model, then we will be accessing a dead reference which will cause a segfault. Static variables have always been fragile so that is something to consider. 

Finally, do we really need a reference to the modelsManager? Can't we use the cache to store all that and only keep the key?

rudiservo:
Hey, the __destruct will be called by the php GC, meaning that the worst case scenario is a GC bug and there is no longer a single source of truth, that is what we have now with the current system, this is only to remove keys that no longer has any stable reference.

So the philosophy of a first level cache system is a single source of truth, meaning that anywhere on the current thread the Model called is always the same.

I think we could pass a cache, in fact, if we are using something like Swoole or Kraken we should have a new cache for each thread, because every thread is independent on its working data.

I'll check it out.

rudiservo:
still checking this btw.

rudiservo:
OK so I get what you mean by models manager, going to remove.

To get the caching situation cleared
SessionCache and cache are two different things in the resultset and in ORM theory

Thread cache/session cache/First Level cache, is only for that request and it's use in PHP with the WeakCache that I added before, so WeakCache will work with the current thread and only know object references via WeakReference, I left it with the Abstract adapter so anyone can make a different cache if they need to for their use case.

The other one is considered the second level cache, slower and longer  lived, that is for redis, memcached or other.

I dont like using __destruct, I don't know what to use to clear the keys in the WeakCache so it wont grow indefinitely, a problem that only affects long lived requests or background tasks.

The only solution I have if a make a Check() function in the weakCache to go over all keys find nulls and deference them so a developer call it if he runs things longer then a quick request, and the rest just let the cache clear out by itself.

I will make these changes and update the PR.

PR #16468, opened 2023-11-23
T16467 column 0 to array

### LINKED ISSUE #16467: [BUG]: `Column '0' is not part of the column map`. After iterating for the second time in a Resultset
We have a problem in our project since the upgrade from phalcon `3.4.5` to the latest version `5.4.0`.
When using the `\Phalcon\Mvc\Model\Query\Builde`r in any of our models we have the same issue after iterating the second time over the `ResultSet` we get the error `Column '0' is not part of the column map` when doing `toArray()` in that result set.

**To Reproduce**

**Code example of bug:**

```php
$result = $models_manager
    ->createBuilder()
    ->addFrom(Products::class, 'p')
    ->limit(10)
    ->getQuery()
    ->execute();

foreach ($result as $value) {}
foreach ($result as $value) {}

$result->toArray();  // Error here - `Column '0' is not part of the column map` 
```

We can also reproduce the error like this.

```php
$result = $models_manager
    ->createBuilder()
    ->addFrom(Products::class, 'p')
    ->limit(10)
    ->getQuery()
    ->execute();

$result->rewind();
$result->next();
dump($result);
$result->rewind();
dump($result); exit;
```

We notice that in the second dump, the property `row` has more properties and with the wrong indexs.

**Screenshots**
**First dump:**
image

**Second dump:**
image

**Details**
 - Phalcon version: 5.4.0
 - PHP Version: 8.1.25
 - Operating System: Docker image - php:8.1-fpm
 - Installation type: installing via package manager
 - Server: Nginx 
 - Database: mysql:8

niden:
Hello!

* Type: bug fix
* Link to issue: #16467 

Fixed `Phalcon\Mvc\Model\ResultsetSimple::toArray` to ignore numeric indexes in case results come as not `fetch_assoc`

Thanks

### DIFF
--- a/CHANGELOG-5.0.md
+++ b/CHANGELOG-5.0.md
@@ -9,6 +9,7 @@
 ### Fixed
 
 - Fixed `Phalcon\Filter\Validation\Validator\Numericality` to return false when input has spaces [#16461](https://github.com/phalcon/cphalcon/issues/16461)
+- Fixed `Phalcon\Mvc\Model\ResultsetSimple::toArray` to ignore numeric indexes in case results come as not `fetch_assoc` [#16467](https://github.com/phalcon/cphalcon/issues/16467)
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


Jeckerson (approved):

apoca:
I'm sorry I was a little confused... Where was this fix? in which version?

niden:
@apoca It is in v5. It will be available with the next release which should come out in a few days.

Issue #16637, opened 2024-08-18
[BUG]: Email validation does not allow UTF8SMTP


noone-silent:
**Describe the bug**
With the UTF8SMTP Extension, it is possible to allow any utf-8 character in the local part. Also domain names can have umlauts, but needs different parsing. RFC: https://www.rfc-editor.org/rfc/rfc5336#section-3.2

**To Reproduce**
Use any email, like `täst@example.com`

```php
            $validation = new Validation();
            $validation->add('email', new Validation\Validator\Email());
            var_dump($validation->validate(['email' => 'täst@example.com']));
```

**Expected behavior**
Add a flag/option to allow UTF8SMTP email addresses.

When this filter is set, pass `FILTER_FLAG_EMAIL_UNICODE` to the `filter_var` function.

**Details**
 - Phalcon version: v5.8.0
 - PHP Version: 8.2.20
 - Operating System: debian
 - Installation type: Compiling from source
 - Zephir version (if any): -
 - Server: Nginx
 - Other related info (Database, table schema): MariaDB

niden:
Resolved in https://github.com/phalcon/cphalcon/pull/16638

Thank you @noone-silent

PR #16638, opened 2024-08-19
[#16637] - refactor: allow utf8 in email local part.

noone-silent:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/16637

Small description of change: Allow UTF8 in local part of the email, if `allowUTF8` is `true`

Thanks

### DIFF
--- a/CHANGELOG-5.0.md
+++ b/CHANGELOG-5.0.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [5.8.x](https://github.com/phalcon/cphalcon/releases/tag/v5.8.x) (xxxx-xx-xx)
+
+### Changed
+
+- Changed `Phalcon\Filter\Validation\Validator\Email` to allow UTF8 in local part. [#16637](https://github.com/phalcon/cphalcon/issues/16637)
+
 ## [5.8.0](https://github.com/phalcon/cphalcon/releases/tag/v5.8.0) (2024-07-09)
 
 ### Changed
--- a/phalcon/Filter/Validation/Validator/Email.zep
+++ b/phalcon/Filter/Validation/Validator/Email.zep
@@ -46,6 +46,16 @@ use Phalcon\Filter\Validation\AbstractValidator;
  *         ]
  *     )
  * );
+ *
+ * $validator->add(
+ *     "täst@example.com",
+ *     new EmailValidator(
+ *         [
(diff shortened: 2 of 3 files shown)


niden (approved):

niden:
Thank you @noone-silent

s-ohnishi:
It seems that you are focusing on internationalization of the local part, but don't you need to consider internationalizing the domain name?

noone-silent:
> It seems that you are focusing on internationalization of the local part, but don't you need to consider internationalizing the domain name?
> 
> By the way, RFC5336 has been abolished and the revised version is RFC 6531.

The domain part with umlauts or non ascii chars is normally not valid and needs to be transformed to ascii through punycode. There is a function in the [INTL extension](https://www.php.net/manual/en/function.idn-to-ascii.php). Non ASCII chars in the local part is totally valid and doesn't need any transformation.

s-ohnishi:
Non-ASCII characters in the local part are still uncommon (which is why they're disabled by default even with this update, perhaps?), so when considering validation of email addresses that contain non-ASCII characters, it seems unfair to not take the domain part into consideration.
Even though it's non-standard, are you validating it by adding a special flag, and then converting the part to the right of the @ character to punycode before validating it?
I think it's only natural that there would be a request to allow non-ASCII characters in the entire email address, rather than just in the local part.
I would simply like to pass it to the validator (even if a flag is specified).

noone-silent:
> Non-ASCII characters in the local part are still uncommon (which is why they're disabled by default even with this update, perhaps?), so when considering validation of email addresses that contain non-ASCII characters, it seems unfair to not take the domain part into consideration. Even though it's non-standard, are you validating it by adding a special flag, and then converting the part to the right of the @ character to punycode before validating it? I think it's only natural that there would be a request to allow non-ASCII characters in the entire email address, rather than just in the local part. I would simply like to pass it to the validator (even if a flag is specified).

We could implement this, but that would need the intl extension and I don't know if this is a requirement for Phalcon or if it is usable in Zephir.

s-ohnishi:
I don't know if this is a breaking change or a requirement of some kind.
But I can't imagine that progressive users who want to allow non-ASCII characters in email addresses don't know about IDNs, and would want to use them, right?

PR #16639, opened 2024-08-28
Update Request.zep

lieshanhe:
enhance getScheme() function.

Hello!

* Link to issue:

Small description of change:

Thanks

### DIFF
--- a/phalcon/Http/Request.zep
+++ b/phalcon/Http/Request.zep
@@ -819,6 +819,11 @@ class Request extends AbstractInjectionAware implements RequestInterface, Reques
             return "https";
         }
 
+        let forwarded_proto = this->getServer("HTTP_X_FORWARDED_PROTO");
+        if forwarded_proto && forwarded_proto == "https" {
+            return "https";
+        }
+
         return "http";
     }
 


raicabogdan:
Unfortunately this is not that simple, it also needs a way to be able to accept reading the header only from specific trusted proxies, you can't just take the forwarded_proto header and use it as a last resort like that without verifying you trust that header. 

Dealing with proxies needs to be careful and I would rather leave this into the control of the proxy backend, or in this case the origin server to read the header and update it for us. But this can be taken into v6 I believe, it is worth having a way to detect if this is a proxy and work with it.

elcreator:
I think this should not be implemented this way as local environment setup for testing the load balancer can also be plain HTTP and we should not confuse developers who will receive SSL error instead of response. 
X-Forwarded-Proto is about scheme browser use to connect to the load balancer (if it exists) https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto .

niden:
@lieshanhe thank you for this.

This doesn't actually compile unless a couple of changes are made. 

Regardless, this introduces an issue where a attacker can send a `X-Forwarded-Proto: https` over plain HTTP and the `isSecure()` will return true. We cannot have that.

There is already work in the `getClientAddress()` and that is the right way forward. 

I have noted this in the plan for the HTTP layer rewrite #17054

PR #16694, opened 2025-02-06
Resolving Issue With Named Parameters and Dispatcher

Fenikkusu:
Hello!

* Type: bug fix
* Link to issue: N/A

Small description of change:

Since PHP 8.0, `call_user_func_array` now uses arrays with non-numeric keys as named parameters. If you have an action that does not use/define properties, this causes the error `Unknown named parameter`. While this can be avoided by adding the parameters, this change fixes it so this is no longer an issue. It also adds in the ability to hook into an event for the calling of the handler, though that was originally just confirmation of the bug. Just making the change to call `array_values` is enough otherwise.

Thanks

### DIFF
--- a/phalcon/Dispatcher/AbstractDispatcher.zep
+++ b/phalcon/Dispatcher/AbstractDispatcher.zep
@@ -19,6 +19,7 @@ use Phalcon\Events\ManagerInterface;
 use Phalcon\Filter\FilterInterface;
 use Phalcon\Mvc\Model\Binder;
 use Phalcon\Mvc\Model\BinderInterface;
+use Phalcon\Support\Collection;
 
 /**
  * This is the base class for Phalcon\Mvc\Dispatcher and Phalcon\Cli\Dispatcher.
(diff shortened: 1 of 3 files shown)


Fenikkusu:
> @Fenikkusu I will look at this a bit later on. Can you please rebase to target the 5.0.0 branch? We don't merge to master other than releases.
> 
> Thanks

I was wondering about that. Since it was 5.8.0, I didn't see a compatible branch, hence I went with master. I've updated it to 5.0.x.

Fenikkusu:
...And Now I've got to do a full rebase...fudge...

niden:
@Fenikkusu don't worry about the PHP 8.4 build failing (from PECL). That is what we are working on now.

niden:
I mean leave the header files as they are. If your PR only fails on PHP 8.4 build then it is good enough.

Fenikkusu:
> I mean leave the header files as they are. If your PR only fails on PHP 8.4 build then it is good enough.

I've reverted them out. I get anal about making sure all the testing is passing.

phalcon/Dispatcher/AbstractDispatcher.zep
```diff
@@ -154,10 +155,52 @@ abstract class AbstractDispatcher extends AbstractInjectionAware implements Disp
...
-            [handler, actionMethod],
-            params
+        var result, observer, altHandler, altAction, altParams;
+
+        let altHandler = handler;
+        let altAction = actionMethod;
+        let altParams = params;
+
+        if this->eventsManager !== null && this->eventsManager instanceof ManagerInterface {
+            let observer = <Registry> this->getDi()->get(
```
niden, line 165: Use a `Collection` object here. Pretty much the same functionality. I have not decided yet if we will keep the registry.

niden (changes requested):

Fenikkusu:
- Cleaned & Squashed Commits
- Added Testing For No Params, With Named Params, and Action
- Updated Change Log

Side Thought: I normally use Route Converters to Convert IDs to Models. One downside of doing it this way is that you cannot throw proper errors and have them redirected as you can with traditional dispatcher events. In theory, someone could use the new events to do this instead of using Converters.

niden (approved):

Issue #16702, opened 2025-02-26
[BUG]: Messages object not present in $form->isValid() with validation events


raicabogdan:
**Describe the bug**
When using `$form->isValid()` to validate form and you have a validation class with `beforeValidation()` method, you get error:
`Trying to call method count on a non-object`

**To Reproduce**
As per documentation https://docs.phalcon.io/5.8/filter-validation/#events
```php
class LoginValidation extends Validation
{
    public function initialize()
    {
        // ...
    }

    public function beforeValidation($data, $entity, $messages)
    {
        if ($this->request->getHttpHost() !== 'admin.mydomain.com') {
            $messages->appendMessage(
                new Message(
                    'Only users can log on in the admin domain'
                )
            );

            return false;
        }

        return true;
    }
}
```

**Details**
 - Phalcon version: 5.8
 - PHP Version: php 8.3

**Additional context**
The issue starts from `$form->isValid()` when `validation->validate(data, entity);` is being called
https://github.com/phalcon/cphalcon/blob/83677f1df8dde17946eee54ddbc900c8839c2b53/phalcon/Forms/Form.zep#L719-L723
and expects back a `Phalcon\Messages` object.

In the `Phalcon\Filter\Validation` class, in the `validate()` method, the `messages` variable is getting passed along to event `beforeValidation` and once the `status` returned from `validate()` is false (as per documentation), the `messages` do not get returned back.
https://github.com/phalcon/cphalcon/blob/83677f1df8dde17946eee54ddbc900c8839c2b53/phalcon/Filter/Validation.zep#L501-L521

Not entirely sure but, is it not possible to do something along the lines

```
// Phalcon\Forms\Form
        /**
        * Perform the validation
        */
        let result = validation->validate(data, entity);
        if result === false {
            let messages = validation->getMessages();
        } else {
            let messages = result;
        }
        
        if messages->count() {
            // Add validation messages to relevant elements
            for elementMessage in iterator(messages) {
                this->get(elementMessage->getField())->appendMessage(elementMessage);
            }

            messages->rewind();

            let validationStatus = false;
        }
```

and also

```
// Phalcon\Filter\Validation
         /** 
          * Implicitly creates a Phalcon\Messages\Messages object 
          */ 
         let this->messages = new Messages(); 
  
         if entity !== null { 
             this->setEntity(entity); 
         } 
  
         /** 
          * Validation classes can implement the 'beforeValidation' callback 
          */ 
[...]

PR #16704, opened 2025-02-27
Fix beforeValidation() event in class validation

raicabogdan:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/16702

Small description of change:
I'm not satisfied with this fix, I personally would not like to have the argument $messages passed to the event method, it is redundant and you can just use `$this->appendMessage()`, but it is what it is and its non-breaking for previous code.

**Please review before doing any approval.**

Thanks

### DIFF
--- a/CHANGELOG-5.0.md
+++ b/CHANGELOG-5.0.md
@@ -11,6 +11,7 @@
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


raicabogdan:
Actually we can simply `Form` class even more and just do

```
        validation->validate(data, entity);
        let messages = validation->getMessages();
```
I don't think anything will break.

niden (approved):

Issue #16705, opened 2025-02-28
Performance Issue: Redundant `has` Call in `get` Method Slowing Down Redis


kunalray1993:
https://github.com/phalcon/cphalcon/blob/83677f1df8dde17946eee54ddbc900c8839c2b53/phalcon/Storage/Adapter/AbstractAdapter.zep#L166

### Description
After upgrading Phalcon from `4.0.6` to `5.8.0` and PHP from `7.2` to `8.3`, we noticed a performance issue related to the `get` method when using Redis as a cache backend. The method first calls `has` (which internally executes an `EXISTS` command) before calling `get`, effectively doubling the number of Redis operations. Under high traffic, this redundancy negatively impacts Redis performance.

### Expected Behavior
The `get` method should retrieve the value directly without first checking for existence, unless absolutely necessary. The default value handling should be done in a more optimized manner.

### Current Behavior
1. The `get` method first calls `has` -> Redis executes `EXISTS key`.
2. Then, `get` is called -> Redis executes `GET key`.
3. This results in two commands for every single retrieval operation, reducing efficiency under load.

### Steps to Reproduce
1. Upgrade Phalcon from `4.0.6` to `5.8.0`.
2. Use Redis as a cache backend.
3. Call `$cache->get('some_key', 'default_value')` in a high-traffic scenario.
4. Monitor Redis and observe the `EXISTS` call happening before `GET`.

### Proposed Solution
Modify the `get` method to avoid calling `has`. Instead, rely on the `GET` result directly:
- If `GET` returns `null` (or false), use the default value.
- This would eliminate the extra `EXISTS` call and improve Redis performance.

### Environment
- Phalcon Version: `5.8.0`
- PHP Version: `8.3`
- Redis Version: `7.0.15`
- Operating System: `Ubuntu 22.04`

![Image](https://github.com/user-attachments/assets/388a8f10-968f-471c-8c51-31beb1029627)
![Image](https://github.com/user-attachments/assets/f5c8c226-ec18-47db-ba80-32ae7c6068ad)

Would it be possible to optimize the `get` method to avoid this redundancy? Looking forward to your feedback.

francoisgrogor:
Upvote this. This needs to be addressed please.

niden:
Resolved in https://github.com/phalcon/cphalcon/pull/16709

PR #16709, opened 2025-03-05
[#16705] - refactor: remove has() check in doGet().

noone-silent:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/16705

Small description of change: Remove the `has()` check in `doGet()`

Thanks


phalcon/Storage/Adapter/AbstractAdapter.zep
```diff
@@ -163,12 +163,6 @@ abstract class AbstractAdapter implements AdapterInterface, EventsAwareInterface
 
         this->fire(this->eventType . ":beforeGet", key);
 
-        if (true !== this->has(key)) {
-            this->fire(this->eventType . ":afterGet", key);
-
-            return defaultValue;
-        }
-
         let content  = this->doGet(key);
```
niden, line 166: Add a check here after the `this->doGet(key)` for `false`. Then return the `defaultValue` if it is. Redis for isntance returns `false` if the operation is not successful. Then all we have to do is adjust each adapter's `doGet` to return `false` on failure if they do not already do so.

phalcon/Storage/Adapter/AbstractAdapter.zep
```diff
@@ -376,7 +370,7 @@ abstract class AbstractAdapter implements AdapterInterface, EventsAwareInterface
             let content = this->serializer->getData();
         }
 
-        return content;
+        return is_null(content) ? defaultValue : content;
```
niden, line 373: This is not needed based on the above.

phalcon/Storage/Adapter/Apcu.zep
```diff
@@ -244,7 +244,11 @@ class Apcu extends AbstractAdapter
      */
     protected function doGet(string key)
     {
-        return this->phpApcuFetch(this->getPrefixedKey(key));
+        var result;
```
niden, line 247: This returns `false` so leave it as is. See comment above

niden (changes requested):

kunal-lottostar (approved):

niden:
@noone-silent Can you rebase please. The changelog entry needs to go to 5.9.1

This looks good.

noone-silent:
> @noone-silent Can you rebase please. The changelog entry needs to go to 5.9.1
> 
> This looks good.

done

niden (approved):

niden:
Thank you @noone-silent

Sn0wCrack:
I've created a bug report in relation to this one, but this is causing issues with adapters that do not do existence checks within themselves or return false naturally.

One issue here is the cached value is actually a PHP `false` then it will always return the default value, which is null. In a loose checked world this is probably fine, but in a strict checked application this may cause problems.

@niden I think this may need to be reverted unless a quick fix can be implemented.

Issue #16749, opened 2025-04-10
[BUG]: Router no longer matches with query strings


raicabogdan:
**Describe the bug**
When you use query strings in URLs, the router no longer matches the specific route without the query string.

**To Reproduce**

```
router->handle(
    '/admin/private/businesses/list/my/123?varname=1234'
);

There was 1 failure:
1) HandleCest: Test Mvc\Router - handle() - with placeholders
 Test  tests/integration/Mvc/Router/HandleCest.php:mvcRouterHandleWithPlaceholders
 Step  Assert equals "admin",""
 Fail  Failed asserting that two strings are equal.
- Expected | + Actual
@@ @@
-'admin'
+''

Scenario Steps:

 1. $I->assertEquals("admin","") at tests/integration/Mvc/Router/HandleCest.php:131


FAILURES!
Tests: 8, Assertions: 27, Failures: 1.
```

Add a query string to a url and try accessing the url

**Expected behavior**
I expect the route to be matched correctly.

**Screenshots**

**Details**
 - Phalcon version: 5.9.2
 - PHP Version: PHP 8.4.5
 - Operating System: Docker php84
 - Installation type: Compiling from source
 - Zephir version (if any): 0.18.0
 - Server: php built-in server with .htrouter

**Additional context**
I believe bug was introduced from issue https://github.com/phalcon/cphalcon/issues/16741

niden:
Resolved in https://github.com/phalcon/cphalcon/pull/16751

Thank you @raicabogdan

Issue #16741, opened 2025-04-04
[BUG]: Router fails to match routes starting with "/number:number/" due to internal parse_url() behavior


maximecaze:
#### Describe the bug
After migrating from Phalcon 3.4.5 (PHP 7.3) to Phalcon 5.9 (PHP 8.4), certain route patterns no longer match correctly if URLs start with `/number:number/` segments (e.g., `/1:1/...` or `/150:200/...`). This is caused by the internal use of PHP's native `parse_url()` function in Phalcon's Router component, which returns `false` for these specific URL patterns.

Such URLs were previously correctly routed in Phalcon 3.x (which did not rely on `parse_url()` internally).

#### To Reproduce
Steps to reproduce clearly and quickly:

```php
<?php
use Phalcon\Di\FactoryDefault;
use Phalcon\Mvc\Router;

$di = new FactoryDefault();
$router = new Router(false);
$router->setDI($di);

// Simple catch-all route
$router->add('/{param:.+}', [
    'controller' => 'index',
    'action'     => 'test',
]);

// Explicitly set request method (for CLI testing)
$_SERVER["REQUEST_METHOD"] = "GET";

// Problematic URLs
$testUrls = [
    '/1:1/test',
    '/a:1/test',
    '/1:a/test',
    '/a:a/test',
];

foreach ($testUrls as $uri) {
    $router->handle($uri);
    echo "Testing URI: $uri\n";
    echo $router->getMatchedRoute()
        ? "✅ MATCHED\n"
        : "❌ NOT MATCHED\n";
}
```

Result in Phalcon 5.9:
```
Testing URI: /1:1/test
❌ NOT MATCHED
Testing URI: /a:1/test
❌ NOT MATCHED
Testing URI: /1:a/test
✅ MATCHED
Testing URI: /a:a/test
✅ MATCHED
```

#### Expected behavior
All provided URIs should match correctly. In Phalcon 3.x, these routes matched properly, regardless of whether the segments contained numeric patterns before and after the ":" character.

#### Root cause clearly identified
Phalcon 5.x internally calls PHP's native `parse_url()` here:

https://github.com/phalcon/cphalcon/blob/v5.9.0/phalcon/Mvc/Router.zep#L360

```zephir
let uri = parse_url(uri, PHP_URL_PATH);
```

PHP's native `parse_url()` returns `false` on URLs starting with "/number:number/", because it interprets such strings as URLs that specify a scheme and port number invalidly:

[...]

PR #16751, opened 2025-04-10
Correctly match route when using query string

raicabogdan:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/16749

Small description of change:
The recent fix for issue https://github.com/phalcon/cphalcon/issues/16741 has introduced this current bug. Missed one place to extract the correct URI for route matching when the `route->handle()` receives the `uri`.

Added a new reusable method `extractRealUri` to avoid replicating the same code in all 3 needed places.

Thanks

### DIFF
--- a/CHANGELOG-5.0.md
+++ b/CHANGELOG-5.0.md
@@ -13,6 +13,7 @@
 - Fixed `Phalcon\Mvc\Router` to correctly handle numeric URI parts as it was in v3 [#16741](https://github.com/phalcon/cphalcon/issues/16741)
 - Fixed `Phalcon\Mvc\Model\Binder` to use ReflectionParameter::getType() instead of deprecated method, PHP 8.0 or higher issue. [#16742](https://github.com/phalcon/cphalcon/issues/16742)
 - Fixed `Phalcon\Mvc\Model\Query` to check if cache entry exists. [#16747](https://github.com/phalcon/cphalcon/issues/16747)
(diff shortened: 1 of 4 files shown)


niden (approved):

niden:
Thank you @raicabogdan

noone-silent:
Shouldn't we use `parse_url` for that?

`return parse_url('/admin/private/businesses/list/my/123?query=string', PHP_URL_PATH)`;

raicabogdan:
It looks cleaner, but I think it would be marginally slower than plain explode().

noone-silent:
> It looks cleaner, but I think it would be marginally slower than plain explode(). The parse_url() will parse and validate before retrieving the thing you need, eg the path.

Actually it is not about cleaner. But this is a function with the whole purpose to parse URLs. I am by far no hacker, but I remember there have been a few attacks based on hosts etc. explode doesn't check if the provided URL is valid at all.

raicabogdan:
> > It looks cleaner, but I think it would be marginally slower than plain explode(). The parse_url() will parse and validate before retrieving the thing you need, eg the path.
> 
> Actually it is not about cleaner. But this is a function with the whole purpose to parse URLs. I am by far no hacker, but I remember there have been a few attacks based on hosts etc. explode doesn't check if the provided URL is valid at all.

It doesn't need to be valid, the router handles the path uri, if its not valid, it just won't match, but when it does match, that is when this is faster.

PR #16776, opened 2025-06-22
fix(config): handle null from yaml_parse_file in Yaml adapter

ucando:
### Description

This PR fixes a potential bug in `Phalcon\Config\Adapter\Yaml` where `yaml_parse_file()` may return `null` when parsing an empty YAML file. Without proper handling, this could result in a `null` value being passed to the parent constructor, leading to unexpected errors.

### Changes

- ✅ Added a `null` check in the constructor of `Yaml` adapter and fallback to an empty array.
- ✅ Added a unit test (`configAdapterYamlConstructHandlesEmptyYaml`) to ensure that empty YAML files return an empty configuration array without throwing an exception.
- ✅ Added `tests/_data/assets/config/empty.yml` as a test fixture file.


niden (changes requested):

niden:
And it seems you need to run `composer cs-fix` to fix PHPCS issues.

Thanks

niden (approved):

PR #16784, opened 2025-07-05
Correctly return IP address when using trusted header

raicabogdan:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/16777

Small description of change:

When using `$this->request->getClientAddress(true);` to extract IP address from `X-Forwarded-For` header it needs to select the first left side IP that is not private or reserved IP address.

Added support for users to add their own `trustedProxies` via method, accepts array of IPs or range of IPs.
```
$this->request->setTrustedProxies(['1.1.1.1']);
```
This will trigger an `Exception("The forwarded proxy IP addresses are not trusted.")` if users mess up their proxy setup.

Thanks


niden (approved):

niden:
Thank you @raicabogdan !!

PR #16793, opened 2025-07-24
Removed method getPatchPut() and passed post|patch|put to new parsing method

raicabogdan:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/16792

Small description of change:

I removed private method getPatchPut() and created a new one for parsing the post data / raw body data and passed post|patch|put through the new parsing method.

This ensure that for all requests methods some form of data is returned, be it json, form-data or x-www-form-urlencoded

Technically users can just use `$request->getPost()` to return their data now, and the `getPut()` and `getPatch()` can be used to have a more semantic understanding through which type of request method the request is being sent.

Thanks

### DIFF
--- a/phalcon/Http/Request.zep
+++ b/phalcon/Http/Request.zep
@@ -68,12 +68,7 @@ class Request extends AbstractInjectionAware implements RequestInterface, Reques
     /**
      * @var array|null
      */
-    private patchCache = null;
-
-    /**
-     * @var array|null
-     */
-    private putCache = null;
+    private postCache = null;
 
     /**
      * @var string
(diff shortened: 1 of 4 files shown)


phalcon/Http/Request.zep
```diff
@@ -763,6 +765,8 @@ class Request extends AbstractInjectionAware implements RequestInterface, Reques
         bool notAllowEmpty = false,
         bool noRecursive = false
     ) -> var {
+        let _POST = this->getPostData(_POST);
```
niden, line 768: I am not sure we should be manipulating the $_POST array here. Shouldn't we just use the getPostData(_POST) in the getHelper? Same comment for below.

niden: What I mean is that the request should be processing data that it got and that data has to be immutable. If the $_POST changes it should change in the next request not the current one mid way so to speak.

niden (changes requested):

niden (approved):

niden:
Thank you @raicabogdan

Issue #16800, opened 2025-08-12
[NFR]: Ability to assign data to entity using the standalone Validation class


raicabogdan:
**Is your feature request related to a problem? Please describe.**
Looking at the documentation, it allows you to assign a specific entity to the validation process, however, no data is being assign to the entity object, rather it is only used for retrieval of existing entity data if fields are matching.
While this is fine, usually you don't use the entity for that purpose when using a form or a validation, but rather want to retrieve the validation data into the entity. Similar to how the Form class handles the data binding.

This does work if the entity doesn't have any field properties, though it just fails back to retrieving the value from the validation data. So you can argue that this is failing as well since its not using the entity.

There are also some inconsistencies within validation when user tries to use the entity within events for example to manipulate certain data.

**Describe the solution you'd like**
Need to ensure the binding of the data and entity happens similar to how the Form class handles the data binding.

**Additional context**
A draft solution incoming soon

niden:
Resolved in https://github.com/phalcon/cphalcon/pull/16801

Thank you @raicabogdan

PR #16801, opened 2025-08-12
Entity binding of validation data when using the standalone Validation class.

raicabogdan:
Hello!

* Type: bug fix | new feature
* Link to issue: https://github.com/phalcon/cphalcon/issues/16800

Small description of change:
This is an attempt fix for the validation entity binding of data

Thanks

### DIFF
--- a/CHANGELOG-5.0.md
+++ b/CHANGELOG-5.0.md
@@ -3,6 +3,8 @@
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


raicabogdan:
For some reason, suspect it is how C uses argument pointers, if I don't add a new variable `inputData` and use `let data = this->data` on line 592, the entity is being assigned to the data array. I know it sounds crazy, but that's what happens and I can only think that this is a C compiler problem.

niden (approved):

niden:
Thank you @raicabogdan

PR #16821, opened 2025-11-22
Add ignoreCheckUploadedFile option to File validator (Fixes #16710)

### LINKED ISSUE #16710: Implement option for ignoring `is_uploaded_file` check in File validator
The [AbstractFile](https://github.com/phalcon/cphalcon/blob/v5.8.0/phalcon/Filter/Validation/Validator/File/AbstractFile.zep) class has the method `checkIsUploadedFile`. However if I want to mock/fake an uploaded file for in example unit testing, I have implement/override all these classes, since they all extend from it:
- Phalcon\Filter\Validation\Validator\File\Resolution\Min
- Phalcon\Filter\Validation\Validator\File\Resolution\Max
- Phalcon\Filter\Validation\Validator\File\Resolution\Equal
- Phalcon\Filter\Validation\Validator\File\Size\Min
- Phalcon\Filter\Validation\Validator\File\Size\Max
- Phalcon\Filter\Validation\Validator\File\Size\Equal
- Phalcon\Filter\Validation\Validator\File\MimeType
- Phalcon\Filter\Validation\Validator\File

https://github.com/phalcon/cphalcon/blob/83677f1df8dde17946eee54ddbc900c8839c2b53/phalcon/Filter/Validation/Validator/File/AbstractFile.zep#L381-L392

Would be much easier if an option would be implemented like:
```php
    protected function checkIsUploadedFile(string name) -> bool
    {
         if ($this->getOption('ignoreCheckUploadedFile')) {
            return true;
        }

        return is_uploaded_file(name);
    }

```

As an addition, it would also be nice if this validator accepted the Phalcon native `$this->request->getUploadedFiles()` instead of the PHP's `$_FILES`.

t-sakaguchi77:
Hello!

* Type: new feature | bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/16710

- [x] I have read and understood the Contributing Guidelines
- [ ] I have created a PR for the documentation about this change

---

### Small description of change:

This PR implements the `ignoreCheckUploadedFile` option in the `File` validator, allowing
developers to bypass the internal `is_uploaded_file()` check. This option is particularly
useful when mocking uploaded files in unit tests.

Additionally, when `ignoreCheckUploadedFile` is set to `true`, empty upload arrays
(e.g. `[]`) are now treated as valid rather than causing a validation failure.

This update improves testability while maintaining backwards compatibility.  
A unit test has been added to verify this behavior.

Note: I did not update the CHANGELOG because Phalcon maintainers usually handle it on merge. 
If it is preferred that contributors update it, I’m happy to add it.

Thanks

### DIFF
--- a/phalcon/Filter/Validation/Validator/File/AbstractFile.zep
+++ b/phalcon/Filter/Validation/Validator/File/AbstractFile.zep
@@ -1,4 +1,3 @@
-
 /**
  * This file is part of the Phalcon Framework.
  *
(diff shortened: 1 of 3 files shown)


tests/unit/Filter/Validation/Validator/FileCest.php
```diff
@@ -0,0 +1,68 @@
+<?php
```
niden, line 1: Can we change this to a PHPUnit test? Apologies, we have changed the testing framework after your PR and it did take a long time to review this :/

phalcon/Filter/Validation/Validator/File/AbstractFile.zep
```diff
@@ -105,18 +109,44 @@ abstract class AbstractFile extends AbstractValidator
         <Validation> validation,
         string field
     ) -> bool {
-        var label, replacePairs, value;
+        var value, label, replacePairs;
```
niden, line 112: This broadens the scope and has the side effect of not checking the empty/UPLOAD* stuff. It needs to be removed.

phalcon/Filter/Validation/Validator/File/AbstractFile.zep
```diff
@@ -87,9 +86,14 @@ abstract class AbstractFile extends AbstractValidator
      */
     public function checkUpload(<Validation> validation, string field) -> bool
     {
-        return this->checkUploadMaxSize(validation, field) &&
-            this->checkUploadIsEmpty(validation, field) &&
-            this->checkUploadIsValid(validation, field);
+        if this->getOption("ignoreCheckUploadedFile") {
```
niden, line 89: I don't think this is necessary. The check in the protected method is enough.

niden (changes requested):

Issue #16836, opened 2026-02-17
[BUG]: Refactor $request->getClientAddress(true)


raicabogdan:
**Describe the bug**
Issue described in detail here: https://github.com/phalcon/cphalcon/issues/16777

**Additional context**
Incoming fix soon

niden:
Resolved in https://github.com/phalcon/cphalcon/pull/16837

Thank you @raicabogdan

PR #16837, opened 2026-02-18
Refactor Request::getClientAddress() when using trustForwardedHeader

raicabogdan:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/16836

Small description of change:

Two changes introduced, to support fallback to `REMOTE_ADDR` when invalid proxy IP was used, and introduced a new  setter `setTrustedProxyHeader` to retrieve the IP address from a user provided header eg. `HTTP_CLIENT_IP`. The user provided trusted header takes priority before checking `X-Forwarded-For` header.

The way it works is ...
1. Using trusted proxies list.
```
$request
    ->setTrustedProxies($trustedProxies)
    ->getClientAddress(true);
```
You can also not use a trusted proxies list, though I recommend using a trusted list for security reasons. It will parse the forwarded header, but may not be as reliable as with a trusted list.

2. Using user provided trusted header,  the header should only ever contain 1 IP address
```
$request
    ->setTrustedProxyHeader('HTTP_CLIENT_IP')
    ->getClientAddress(true);
```
If the `trustedProxyHeader` is empty, it will fallback to x-forwarded-for, in which case I would again recommend setting a proper trusted proxy list.

Using the proxy method should be used in careful controlled environment.

Requesting a review before merging please.

Thanks

### DIFF
--- a/CHANGELOG-5.0.md
+++ b/CHANGELOG-5.0.md
@@ -1,4 +1,17 @@
 # Changelog
+## 5.11.0 (2026-XX-XX)
+
+### Changed
+
+### Added
+
+### Fixed
+
+- Fixed `Phalcon\Http\Request` method `getClientAddress()` when using `trustForwardedHeader`.[#16836](https://github.com/phalcon/cphalcon/issues/16836)
+
+### Removed
+
+# Changelog
 ## 5.10.0 (2025-12-25)
 
 ### Changed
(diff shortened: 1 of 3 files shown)


niden (approved):

niden:
Thank you @raicabogdan

Issue #16852, opened 2026-03-09
[BUG]: Logger performance and level changes


niden:
For the Logger class the levels are being recorded in upper case. For phalcon/phalcon using PSR we record them lower case.

- The levels need to change to record lower case to align with phalcon/phalcon
- There is a known performance hit for the Stream adapter which opens the stream every time it needs to write a message. It is best to open it once, write everything and then close it.

niden:
Resolved in https://github.com/phalcon/cphalcon/pull/16853

PR #16853, opened 2026-03-09
T16852 logger adjustments

niden:
Hello!

* Type: bug fix 
* Link to issue: #16852 

For the `Phalcon\Logger`

The levels need to change to record lower case to align with phalcon/phalcon
There is a known performance hit for the Stream adapter which opens the stream every time it needs to write a message. It is best to open it once, write everything and then close it.

Thanks

### DIFF
--- a/phalcon/Logger/Adapter/Stream.zep
+++ b/phalcon/Logger/Adapter/Stream.zep
@@ -93,9 +93,12 @@ class Stream extends AbstractAdapter
      */
     public function close() -> bool
     {
+        var handler;
+
         if (this->handler !== null) {
-            let this->handler = null;
-            return this->phpFclose(this->handler);
+            let handler = this->handler,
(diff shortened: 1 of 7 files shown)


phalcon/Logger/Adapter/Stream.zep
```diff
@@ -85,7 +93,17 @@ class Stream extends AbstractAdapter
...
+
+        let result = true;
+
+        if (is_resource(this->handler)) {
+            let result = this->phpFclose(this->handler);
+        }
+
+        let this->handler = null;
+
+        return result;
```
Jeckerson:
```suggestion
        if (this->handler !== null) {
            let this->handler = null;
            return this->phpFclose(this->handler);
        }

        return true;
```

Jeckerson (changes requested):

Issue #16955, opened 2026-04-30
[BUG]: Segmentation Fault (Core Dump) in ORM findFirst during high-iteration loops


dz3n:
**Describe the bug**
Running a high-iteration loop using the ORM (`findFirst`) results in a hard Segmentation Fault (Core Dump) rather than a standard PHP userland memory exhaustion error.

> Provide minimal script to reproduce the issue

```php
for ($i = 1; $i <= 5000000; $i++) {
    Business::findFirst($i);
}
```

**Result**
```
#34  0x61b367ba910c [/usr/local/bin/php(execute_ex+0x785c) [0x61b367ba910c]]
#35  0x61b367ba00d3 [/usr/local/bin/php(zend_execute+0x143) [0x61b367ba00d3]]
#36  0x61b367c08d80 [/usr/local/bin/php(zend_execute_script+0x50) [0x61b367c08d80]]
#37  0x61b367a9fab5 [/usr/local/bin/php(php_execute_script_ex+0x165) [0x61b367a9fab5]]
#38  0x61b367c0ae0b [/usr/local/bin/php(+0x80ae0b) [0x61b367c0ae0b]]
#39  0x61b367653b29 [/usr/local/bin/php(+0x253b29) [0x61b367653b29]]
#40  0x7b7826fef24a [/lib/x86_64-linux-gnu/libc.so.6(+0x2724a) [0x7b7826fef24a]]
#41  0x7b7826fef305 [/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x85) [0x7b7826fef305]]
#42  0x61b367654da1 [/usr/local/bin/php(_start+0x21) [0x61b367654da1]]
```

**Expected behavior**
The script should eventually run out of memory and throw a standard PHP Fatal error: Allowed memory size exhausted. It should not trigger a Segmentation Fault at the C level.

**Details**
 - Phalcon version: 5.12.0
 - PHP Version: 8.5

niden:
I have investigated this and it seems that it is no longer an issue. I threw AI on it to perform runs and proof of the problem. There is a combination that generates a segfault but it is upstream on php itself

PHP 8.5's new `fatal_error_backtraces` feature (default on) infinitely recurses when OOM strikes with no allocator slack. 

From the core dump:

```
zend_error_noreturn ("Out of memory")
  -> `zend_error_zstr_at` -> `zend_fetch_debug_backtrace`
    -> `zend_hash_real_init_mixed` -> OOM again
[...]

PR #16988, opened 2026-05-05
T16976 findfirst segfault memory cache limit

Lucifer07:
Hello!

* Type: bug fix 
* Link to issue: https://github.com/phalcon/cphalcon/issues/16955

- [x] I have read and understood the Contributing Guidelines (https://github.com/phalcon/cphalcon/blob/master/CONTRIBUTING.md)
- [x] I have created a PR for the documentation (https://github.com/phalcon/docs) about this change

Small description of change:

Running findFirst in a high-iteration loop (e.g. 5M iterations) causes a segmentation fault instead of a standard PHP Allowed memory size exhausted error. Two fixes applied:
1. Phalcon\Mvc\Model\Query::parse() — Added FIFO eviction on internalPhqlCache: when cache exceeds 1024 entries, the oldest 512 are dropped via array_slice, preventing unbounded memory growth from dynamic PHQL conditions.
2. Phalcon\Db\Result\PdoResult::__destruct() — Added destructor that calls closeCursor() on the PDOStatement and nullifies resource references, preventing accumulation of unreleased prepared statements in tight loops. Wrapped in try/catch for safe PHP shutdown order.

Thanks


niden:
@Lucifer07 I apologize I did not post in our documentation early enough regarding AI development so you did not know.

I would kindly ask you to redo this PR using instructions from here: 

https://docs.phalcon.io/5.12/ai-development/

In short, every commit assisted by an AI has to have it in the commit message. 

Apologies again.

phalcon/Db/Result/PdoResult.zep
```diff
@@ -97,6 +97,24 @@ class PdoResult implements ResultInterface
...
+     * Prevents unbounded accumulation of open statements in tight loops
+     * that would otherwise lead to a segmentation fault.
+     */
+    public function __destruct()
+    {
+        if typeof this->pdoStatement == "object" {
+            try {
+                this->pdoStatement->closeCursor();
+            } catch \Exception {
+            }
```
niden, line 111: There has to be some code in the `catch` branch to alert the developer that an exception has happened. The way this is now, it will not only work in this case (fixing this bug) but it will also suppress any other exception that might happen unrelated to this fix.

phalcon/Mvc/Model/Query.zep
```diff
@@ -648,9 +648,20 @@ class Query implements QueryInterface, InjectionAwareInterface
...
          */
         if typeof uniqueId == "int" {
+            if count(self::internalPhqlCache) > 1024 {
+                let self::internalPhqlCache = array_slice(
+                    self::internalPhqlCache,
+                    -512,
+                    null,
+                    true
+                );
+            }
```
niden, line 706: Please add the 1024 and 512 as constants in this file with a docblock of what they represent

tests/database/Db/Result/Pdo/DestructTest.php
```diff
@@ -0,0 +1,145 @@
...
+namespace Phalcon\Tests\Database\Db\Result\Pdo;
+
+use PDO;
+use PDOStatement;
+use Phalcon\Db\Result\PdoResult;
+use Phalcon\Tests\AbstractDatabaseTestCase;
+use Phalcon\Tests\Support\Traits\DiTrait;
+
+/**
+ * @issue  https://github.com/phalcon/cphalcon/issues/16955
```
niden, line 23: This one goes in the test i.e. method

tests/database/Db/Result/Pdo/DestructTest.php
```diff
@@ -0,0 +1,145 @@
...
+
+    public function tearDown(): void
+    {
+        $this->tearDownDatabase();
+    }
+
+    /**
+     * Tests that PdoResult::__destruct calls closeCursor and nullifies resources.
+     *
+     * @author Phalcon Team <team@phalcon.io>
```
niden, line 43: This one is your name - we didn't do this work.

niden: Same with the rest of the tests in this file

phalcon/Mvc/Model/Query.zep
niden: I would change that to be 75% instead of 50%. It does reduce the internal cache but not halving it which could in theory have a bigger impact.

niden (changes requested):

niden:
Closing this one - we could not replicate it in the source issue https://github.com/phalcon/cphalcon/issues/16955

Thank you @Lucifer07

PR #16990, opened 2026-05-05
Fix query builder order by complex expressions

QUAD69:
Hello!

* Type: bug fix
* Link to issue: #17077 

Small description of change:

This PR fixes `Phalcon\Mvc\Model\Query\Builder::orderBy()` when the array syntax is used with complex PHQL expressions.

Previously, any array item containing a space was treated as a simple `column direction` pair. This worked for values like `inv_title DESC`, but corrupted complex expressions such as `CASE WHEN inv_status_flag = 1 THEN 0 ELSE 1 END ASC`.

The builder now checks whether the last part of the order item is `ASC` or `DESC`. Simple column ordering keeps the existing `autoescape()` behavior, while complex expressions are preserved as-is.

A regression test was added for complex `ORDER BY` expressions.


niden:
@QUAD69 Thank you for the pull request but for now it is going to be rejected.

We announced this morning rules about AI assisted development, and I understand you might have not seen it yet. This is the only reason for the rejection.

I will close this PR, and please create an issue describing what you are trying to fix, and then issue a new PR after you familiarize yourself with the AI development rules.

In short: every commit has to have `Authored-by: xxx" when using AI.

Thanks

niden:
Thanks @QUAD69

Issue #17141, opened 2026-06-12
[NFR]: Add PHQL support for common table expressions


QUAD69:
## Summary

Add PHQL support for common table expressions in SELECT queries.

## Motivation

PHQL currently cannot parse queries that use `WITH` or `WITH RECURSIVE`. Applications that need common table expressions, including recursive SELECT queries, have to fall back to raw SQL and bypass PHQL query parsing/generation.

## Proposed approach

Support common table expressions for SELECT PHQL queries:

- `WITH name AS (SELECT ...) SELECT ...`
- `WITH name (column1, column2) AS (SELECT ...) SELECT ...`
- multiple CTE definitions
- `WITH RECURSIVE`
- `UNION` and `UNION ALL` inside CTE SELECT statements

CTE names should be treated as virtual query sources and should not require model hydration.

## Example

```sql
WITH recent AS (
    SELECT inv_id, inv_total
    FROM Invoices
)
SELECT recent.inv_id, recent.inv_total
FROM recent
```
```sql
WITH RECURSIVE tree AS (
    SELECT id, parent_id FROM Categories WHERE parent_id IS NULL
    UNION ALL
    SELECT c.id, c.parent_id FROM Categories c, tree t WHERE c.parent_id = t.id
)
SELECT tree.id FROM tree
```

niden:
Closing this in favor of #14608

PR #17142, opened 2026-06-12
Add PHQL common table expression support

QUAD69:
Hello!

* Type: new feature
* Link to issue: #17141 

Small description of change:

Adds SELECT-only PHQL support for common table expressions, including `WITH`, `WITH RECURSIVE`, and `UNION` / `UNION ALL` in CTE SELECT statements. CTE sources are handled as virtual query sources and do not require model hydration.

Also adds Query Builder support for non-recursive CTE declarations through `with()` and `addWith()`.

Thanks!


phalcon/Mvc/Model/Query/Builder.zep
```diff
@@ -1592,6 +1700,117 @@ class Builder implements BuilderInterface, InjectionAwareInterface
...
+
+        if typeof query == "string" {
+            return query;
+        }
+
+        if typeof query == "object" {
+            if unlikely !method_exists(query, "getPhql") {
+                throw new Exception(
+                    "A common table expression query must be a PHQL string or a builder"
+                );
```
niden, line 1765: This needs to be either a dedicated exception or reusing one of the existing ones.

phalcon/Mvc/Model/Query/Builder.zep
```diff
@@ -1592,6 +1700,117 @@ class Builder implements BuilderInterface, InjectionAwareInterface
...
+            }
+
+            let phql = query->getPhql();
+
+            if method_exists(query, "getBindParams") {
+                let bindParams = query->getBindParams();
+                if typeof bindParams == "array" && count(bindParams) {
+                    let this->bindParams = this->bindParams + bindParams;
```
niden: There is going to be a collision here with the existing APL parameters. They start with APL0/APL1. The collision happens when you have a limit clause

phalcon/Db/Dialect.zep
```diff
@@ -674,6 +683,13 @@ abstract class Dialect implements DialectInterface
             );
         }
 
+        /**
+         * Resolve UNION clauses
+         */
+        if fetch unions, definition["union"] && unions {
+            let sql .= " " . this->getSqlExpressionUnion(unions, escapeChar, bindCounts);
```
niden: If we have a `UNION` + `ORDER BY` this will create an incorrect SQL. Add a test to confirm.

phalcon/Mvc/Model/Query.zep
```diff
@@ -3346,6 +3410,18 @@ class Query implements QueryInterface, InjectionAwareInterface
          * Check for select * (all)
          */
         if columnType == PHQL_T_STARALL {
+            let sqlAliasesVirtual = this->sqlAliasesVirtual;
+            if count(this->models) == 0 && count(sqlAliasesVirtual) {
```
niden: You are ignoring the CTE columns here

tests/unit/Mvc/Model/Query/Phql/Select/WithTest.php
```diff
@@ -18,6 +18,328 @@
...
+            unset($actual['id']);
+
+            $this->assertSame($name, $actual['with'][0]['name']);
+            $this->assertSame($name, $actual['select']['columns'][0]['column']['domain']);
+            $this->assertSame($name, $actual['select']['tables']['qualifiedName']['name']);
+        }
+    }
+
+    /**
+     * @author Phalcon Team <team@phalcon.io>
```
niden: This has to be your information not the team

tests/unit/Mvc/Model/Query/Phql/Select/WithTest.php
```diff
@@ -18,6 +18,328 @@
...
+     * @since  2026-06-12
+     */
+    public function testMvcModelQueryPhqlSelectWithRecursiveCommonTableExpression(): void
```
niden, line 253: This tests the happy path only. I would like to see error paths such as CTEs defined twice, UNION, ORDER BY etc. to make this more thorough.

niden (changes requested):

niden:
@QUAD69 With the recent enhancements to the Db/Mvc layers (using AI), I investigated the use of WITH and CTEs and after a lot of pain and suffering I opted not to implement them. There are fundamental issues here primarily with how the existing code is structured. In order to change it to something that CTEs can be just a small change will take a very long time

For this PR I thought after the first run of reviews that it looked OK but I did ask AI to review this and unfortunately something that scares me came up - we are breaking backwards compatibility. I checked the suggestions myself and it is accurate.

Honestly I don't think we are adding a lot of value to the framework with this change, while at the same time increasing the risk of breaking things. We can get away with the additions to the interface (they are additive) but some stuff might break while we do this. The biggest problem is the lexer which mixes union and recursitve. 

Sorry but this cannot be added - at least not without serious refactoring which cannot happen in one minor version - this needs a major one. We can certainly consider it for v7, the next cphalcon release but not today. 

What I am going to do is move the NFR to the voting list issue in order to see if the community really wants this and then pick it from there. 

Thank you so much for this - really appreciated.

PR #17261, opened 2026-07-04
CP-17254 regex for validatorfiles allowedtypes

niden:
Hello!

* Type: new feature
* Link to issue: #17254 

Added an opt-in `allowWildcards` option to `Phalcon\Filter\Validation\Validator\File\MimeType` (and passed through from `Phalcon\Filter\Validation\Validator\File` via the same key) that, when `true`, matches each configured type against the detected MIME type as an anchored regular expression (e.g. `image/.*`, `video/.*`) in addition to an exact string match; it defaults to `false`, preserving the existing exact-match behavior. 

Thanks


phalcon/Filter/Validation/Validator/File/MimeType.zep
```diff
@@ -76,7 +76,8 @@ class MimeType extends AbstractFile
      */
     public function validate(<Validation> validation, var field) -> bool
     {
-        var fieldTypes, mime, replacePairs, tmp, types, value;
+        var allowWildcards, fieldTypes, matched, mime, replacePairs, tmp,
```
Jeckerson:
```suggestion
        var allowWildcards, fieldTypes, matched = false, mime, replacePairs, tmp,
```

And remove `let matched = false;` below.

Issue #17546, opened 2026-09-01
[BUG]: Resultset close upper transaction


ryinner:
Questions? Discussions: https://phalcon.io/discussions or Discord: https://phalcon.io/discord

**Describe the bug**

If i use delete or update methods, they will begin new transaction and commit it, but if i have upper transaction it break pipeline

**To Reproduce**

```php
$dbConnection = Di::getShared()->get('db');

$dbConnection->begin();

MyExampleModel::find(['id BETWEEN 3 AND 5'])->delete();
// Transaction is already commited

$dbConnection->commit();
```

niden:
Resolved in https://github.com/phalcon/cphalcon/pull/17593

Thank you @ryinner

PR #17547, opened 2026-09-01
5.0.x

ryinner:
Hello!

* Type: bug fix
* Link to issue: https://github.com/phalcon/cphalcon/issues/17546

Small description of change:

Add connection->isUnderTransaction for Resultset::delete and Resultset::update


niden:
@ryinner Thank you for this

We will need some changes. For starters this does not compile because the variable you use has not been declared in the `var` line. You need to declare that there first, and then the `let ....` works and it compiles.

You did not cover the `rollback()` branch here though, as such we are not fixing the issue.

This however is a much bigger fix than this change. There are other spots that this happens and the correct fix would be the `AbstractPdo` to do this.

Also, there are no tests in this PR or a changelog entry. 

Let me know how you want to proceed. I can help with the AbstractPdo.

ryinner:
Thank you, I’m not sure I have enough knowledge of Zephir to properly refactor AbstractPdo. 

After studying AbstractPdo and Db\Dialect, I see that the solution has already been implemented based on Savepoints in the transaction. I can assume that transactionLevel is being triggered incorrectly somewhere. 

I’ll try to find the place where the main transaction might be closing instead of the savepoint.

If I manage to find this place, I’ll rework the PR.

niden:
@ryinner I can help with the AbstractPdo. I will post some guidance shortly. 

Just a tip. If you want you can develop this fix/change in phalcon which is php based and your changes are immediate i.e. you do not need to recompile the extension. Then converting that code to zephir is really easy.

ryinner:
Sorry to bother you @niden,
Apparently, this error has specific conditions for its occurrence. Now I can't reproduce it, I've done the same db pipeline, but now everything works out as it should.

I think I'll close this PR for now and if I catch it again, I'll come back to this PR again

